Issue #21, opened 2014-11-22
LoadError running rake assets:precompile:primary


pmackay:
This is to build a vagrant box, happens in development, test and staging environments:

```
LoadError: cannot load such file -- open_food_network/searcher
/home/vagrant/.gem/ruby/1.9.1/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
/home/vagrant/.gem/ruby/1.9.1/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:251:in `block in require'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:236:in `load_dependency'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:251:in `require'
/home/vagrant/apps/openfoodnetwork/current/config/initializers/spree.rb:11:in `<top (required)>'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:245:in `load'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:245:in `block in load'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:236:in `load_dependency'
/home/vagrant/.gem/ruby/1.9.1/gems/activesupport-3.2.19/lib/active_support/dependencies.rb:245:in `load'
/home/vagrant/.gem/ruby/1.9.1/gems/railties-3.2.19/lib/rails/engine.rb:593:in `block (2 levels) in <class:Engine>'
/home/vagrant/.gem/ruby/1.9.1/gems/railties-3.2.19/lib/rails/engine.rb:592:in `each'
/home/vagrant/.gem/ruby/1.9.1/gems/railties-3.2.19/lib/rails/engine.rb:592:in `block in <class:Engine>'
/home/vagrant/.gem/ruby/1.9.1/gems/railties-3.2.19/lib/rails/initializable.rb:30:in `instance_exec'
/home/vagrant/.gem/ruby/1.9.1/gems/railties-3.2.19/lib/rails/initializable.rb:30:in `run'
[...]

PR #27, opened 2015-03-29
Remove spree.rb

pmackay:
Updates to match https://github.com/openfoodfoundation/openfoodnetwork/issues/443 and also resolves the issue that caused #21.


rafaqz:
Hey perfect we don't need spree.rb at all any more with the ENV variables being in master.

PR #41, opened 2015-10-15
Unprivileged web user

RohanM:
Create an unprivileged user, and run unicorn from that user.


roles/common/tasks/main.yml
```diff
@@ -33,8 +33,22 @@
     #- fontconfig # needed for phantomjs
     #- nfs-common # make virtualbox faster
 
-- name: copy up .bashrc
-  template: src=bashrc.j2 dest=/{{ home_path }}/.bashrc
+- name: copy up .bashrc (admin user)
```
mkllnk: Maybe we can just add rbenv setup to the .bashrcs instead of replacing the whole file. The file might have other important customisations.

roles/dbserver/tasks/main.yml
```diff
@@ -11,17 +11,26 @@
...
+  sudo_user: "{{ unicorn_user }}"
 
 - name: copy fix_pg_encoding PG script
-  copy: src=fix_pg_encoding.txt dest={{ home_path }}
+  copy: src=fix_pg_encoding.sql dest={{ unicorn_home_path }}
+  sudo: yes
+  sudo_user: "{{ unicorn_user }}"
   tags: pg-en
 
 - name: fix postgres db encoding problem
```
mkllnk: Maybe there is a better way to call the script passed in one command instead of uploading, executing and deleting.

roles/unicorn_user/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+--- # Setup an unprivileged user - needs to be run as root.
+- name: create unprivileged user
+  user: name="{{ unicorn_user }}" password="{{ unicorn_user_pass }}" append=yes state=present shell=/bin/bash
```
mkllnk: Better to setup the unicorn user without password so that a login is not possible unless you modify .ssh/authorized_keys.

PR #44, opened 2015-11-19
Canada updates

mkllnk:
All fixes required to deploy the Canada instance. All commits are independent and address one particular problem.


pmackay:
@mkllnk could add a bit more info on what the problem is, for reference? Questions:

- is moving to ruby v2 valid for any install of OFN now? 
- what has changed about the seeds config?

mkllnk:
Hi Paul, good to see you online.

> is moving to ruby v2 valid for any install of OFN now? 

Ruby 2 is used by our master branch and will go into the next release (1.3). Do you think we should  choose the ruby version depending on the Gemfile?

> what has changed about the seeds config?

Not much. It's just that we don't have any code references to suburbs any more. We are using Google Maps to find places near by. And since the seeding would fail with an empty suburbs file, I thought I can remove the suburb seeding completely. Did I overlook something?

The unicorn restart issue was introduced when we separated the ansible admin user (e.g. ubuntu) from the unprivileged unicorn user (e.g. openfoodnetwork).

Any more questions?

MikeiLL:
In case it's at all useful, I ended up using an i18n file for the US instance where suburbs contains exactly one element (as opposed to the thousands that exist in the US) and the map pulls up perfectly.

pmackay:
@mkllnk thanks for clarifications :)

Issue #39, opened 2015-09-27
You don’t have to write every role from scratch.


ypid:
First of all, nice that there is an Ansible Playbook and a Vagrantfile to setup openfoodfoundation.

The problem is that it is not easy that way to integrate the roles into my Ansible setup. I already use other roles to manage various aspects of my servers so the current approach of ofn_deployment kind of conflicts with this. One way to solve this would be to move the currently "playbook centric" approach to a more "role centric" approach. And then publish the openfoodfoundation related roles individually.

Any suggestions for such an advanced user?

Second related hint: You don’t have to write common roles like Database setup or Webserver setup yourself. Have a look at the [DebOps project](http://debops.org/) for a set of high-quality roles for all the basics like this.

pmackay:
Hi @ypid, I'm aware that in this first version of ansible scripts for OFN, there is a lot that is coded here that could be managed by other roles. Its a refactoring task that would be great to do, but currently not so high on the priority list.

ypid:
Understandable. Maybe someday :smile:

pmackay:
@ypid just wondering, do you have any thoughts on best practice of developing a role at the same time as a playbook using the role?

ypid:
I would go with one git repository per role, ideally all interdependently tested using CI and then create a repository for the playbooks which can include the individual roles using git submodules. I kind of do similar things for my personal role/playbook stack: https://github.com/ypid/ypid-ansible-common
Alternatively, you can also use Ansible Galaxy and use it to handle role dependencies of the playbooks. For reference, here is also the main playbooks repo of the DebOps project: https://github.com/debops/debops-playbooks

pmackay:
Closing but this is something we are gradually evolving to. Not sure if it will be reached though :)

PR #47, opened 2016-03-26
Infrastructure config

pmackay:
The main purpose of this branch is to refactor the project so as much of infrastructure config is committed, with these general goals: 

* remove the need for `vars.yml`
* moving towards supporting #39 better
* have a single project that can easily manage any OFN server, *assuming people have the ability to access the server*. So it might be that someone in the Aus team could help manage upgrades. If this is undesirable, at the least it creates more consistency in how to manage servers even if each regional team does it.

There are some other tidyup commits here too but the main inventory one is 7b96174. 

Related to this I'd propose changing the repo name to `infrastructure` because overall it does provisioning and deployment, ultimately its about managing OFN infrastructure.

Broad areas still to be looked at:

* How to manage secrets. Do we use ansible vault? Keep them local and add to `gitignore`? Use a separate private credentials repo? 
* Separating out the roles into new role projects that can be included independently if possible, e.g. create a `ansible-role-openfoodnetwork` project.

@mkllnk @RohanM @oeoeaio @lin-d-hop would appreciate your input on this.


mkllnk:
I really like the new inventories with the server configs. That gives newcomers pretty good examples of how to do it. And it's so much more organised. :+1:

> How to manage secrets

Which secrets do we have?

- ssh key: I would prefer to copy the developer's public key to the target machine first. That is the authorize step.
- ssl certificates: It would be great to integrate letsencrypt into the scripts. The server could obtain certificates for itself then.
- aws, newrelic, bugsnag... credentials: Can we run that as optional steps after deployment? And can these steps ask for these credentials on the command line? If yes, I would do that and then store the credentials only on the server.

pmackay:
We could store people's public keys in git as well. OKFN do it and then have scripts for adding/removing keys based on who should have access where. Worth thinking about? 

SSL certs - yes would be great to do more here, could be a separate feature? Do you want to create a new issue if we dont have one about it? 

>If yes, I would do that and then store the credentials only on the server.

Do you mean in some kind of `.env` file or similar? Only problem with asking is that we dont want to create scripts that have input prompts all the time. It could work if it prompted the first time but didnt ask again if they were present maybe?

mkllnk:
> We could store people's public keys in git as well.

Interesting. Seems convenient.

>> If yes, I would do that and then store the credentials only on the server.

> Do you mean in some kind of .env file or similar?

No. I thought of a task *install bugsnag*. That task needs the credentials. But once installed, you never need to touch it unless you need to change the credentials. Then you have to enter the new ones. They are stored on the server in the bugsnag.rb initializer file. If that is an optional step, it would not be run during the normal provision/deploy.

Well, I'm thinking of the small scale we have at the moment. If we want to provision ten or more servers with the same bugsnag keys, we should have them on the local machine. Unless we can tell the script: "Take the credentials from server1 and copy them to servers 2..n."

Ansible Vault sounds good as well. We would need to share the password then. It would be perfect if the vault files could be encrypted using the public keys of the people that are allowed to access the servers. That is probably a tool that doesn't exist yet.

Vagrantfile
```diff
@@ -20,26 +19,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     # VM network config.
     config.vm.network "forwarded_port", guest: 80, host: 8080
-    #config.vm.network "forwarded_port", guest: 443, host: 44343
-    config.vm.synced_folder ".", "/vagrant", :create => true
-    config.ssh.forward_agent = true
+    config.vm.network "private_network", ip: "192.168.50.4"
```
mkllnk: What is this for? Why that specific IP address?

pmackay: Because a specific IP address allows the same thing to be added to an Ansible hosts file and then ansible commands can be run on the vagrant host directly, without having to run through the vagrant provisioning process.

inventory/host_vars/_example.com/secrets.example.yml
```diff
@@ -0,0 +1,15 @@
...
+
+db_password:
+
+admin_password:
+
+
+# Secret Token
+# This is required to verify session. It must be at least 30 characters and random. This
+# must be changed for a production server.
+# Generate one with a bash command like: openssl rand -hex 128
```
mkllnk: Can we automate this?

pmackay: Potentially but could that be recorded as a future improvement?

mkllnk: Sure.

PR #49, opened 2016-05-01
lowercase L was mistypes for i in i10n_repo variable

MikeiLL:
easy fixes are nice.


pmackay:
Looks like you did these ages ago! I will try to review soon.

MikeiLL:
Hi Paul. My apologies. Most of the commits here are from months ago. I probably should have done a merge as opposed to a rebase. Only the [latest commit](https://github.com/openfoodfoundation/ofn_deployment/pull/49/commits/8d663ecffb0bf586553970e35f2935babab969bb) is relevant, or very relevant at any rate.

MikeiLL:
Avoid:

	server {
		listen 80;
		server_name  www.127.0.0.1;
		rewrite ^/(.*) http://127.0.0.1/$1 permanent;
	}

deploy.yml
```diff
@@ -7,7 +7,7 @@
 
   vars_files:
     - vars/default_vars.yml
-    - vars.yml
+    - host_vars/staging.vars.yml
```
pmackay: #47 will introduce host and group vars, so hopefully this is covered.

roles/app/templates/post-receive.j2
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-APP_PATH="$HOME/apps/openfoodnetwork"
+APP_PATH="$HOME/apps/ofn_america"
```
pmackay: Why change this? Its not really intended to be changed. Does it need to be a variable?

roles/language/tasks/main.yml
```diff
@@ -1,9 +1,13 @@
 ---
-- name: install language package(s)
-  sudo: yes
-  apt: pkg={{ item }} state=latest force=yes
-  with_items: language_packages
+#- name: install language package(s)
+#  sudo: yes
+#  apt: pkg={{ item }} state=latest force=yes
+#  with_items: language_packages
```
pmackay: Why comment this out?

roles/language/tasks/main.yml
```diff
@@ -1,9 +1,13 @@
...
 
+- name: generate locales
+  shell: "locale-gen {{ language }}"
+  sudo: yes
+
```
pmackay: What is this doing? Can you explain a bit?

roles/webserver/templates/nginx_unicorn.j2
```diff
@@ -1,5 +1,5 @@
 upstream unicorn {
-  server unix:{{ unicorn_sock }} fail_timeout=0;
+  server unix:{{ unicorn_sock }} fail_timeout=1;
```
pmackay: What effect does this have?

roles/webserver/templates/nginx_unicorn.j2
```diff
@@ -18,7 +18,7 @@ server {
 
   include {{ app }}_ssl;
 
-  server_name www.{{ domain }};
+  server_name {{sub_domain}} {{ domain }};
```
pmackay: Really the domain variable should include the "www". So I wonder if rather than adding a `sub_domain` variable, we should rename `domain` to `hostname`, because hostname implies the whole domain?

roles/app/templates/post-receive.j2
MikeiLL: No. It certainly shouldn't be changed universally. Didn't mean to push this change, and have since realized that it makes sense to just leave the app name as universal. At least until there's a reason not to.

roles/language/tasks/main.yml
MikeiLL: I was getting `invalid locale settings` error on provisioning (which I luckily [documented](http://www.mzoo.org/beyond-t1-micro-for-ofn_deployment/), since it's been since November that I've looked at any of this). I think this [SO](http://askubuntu.com/questions/162391/how-do-i-fix-my-locale-issue) posting may have lead me in this direction.
Local-gen man page explains:

> Compiled locale files take about 50MB of disk space, and most users only need few locales. In order to save disk space, compiled locale files are not distributed in the locales package, but selected locales are automatically generated when this package is installed by running the locale-gen program.

MikeiLL: I also had to run in the terminal:

`export LANGUAGE="en_US.UTF-8"
echo 'LANGUAGE="en_US.UTF-8"' | sudo tee --append /etc/default/locale
echo 'LC_ALL="en_US.UTF-8"' | sudo tee --append /etc/default/locale`

roles/language/tasks/main.yml
MikeiLL: Probably didn't mean to push this through. I [documented](http://www.mzoo.org/deploying-openfoodnetwork-to-staging-on-aws-ec2/) having done it when deploying to AWS, but not very clearly. So as with most of these changes, probably best to ignore and not spend any more time on it than I've already wasted in my haste in ignorance. : )

deploy.yml
MikeiLL: Very cool!

roles/webserver/templates/nginx_unicorn.j2
MikeiLL: I didn't document doing it and it seems like it should either be zero or a number of seconds.

roles/webserver/templates/nginx_unicorn.j2
MikeiLL: Yes. I thought I had reverted this change. I had seen the server_name `www.127.0.0.1` and thought that would be a problem, but subsequently realized it's only a name - and the second `server_name` ended up being a duplicate of the first (which broke provisioning - or at least broke something along the way) when done this way. I think this is just a line of code I forgot to revert back.

pmackay: Do you think `domain` would be more clearly defined as `hostname`?

MikeiLL: Yes. Domain is fine, but `hostname` is better. As you said, it's the actual term for that part of the url. It's a little more obtuse in terms of the general population, but the `www.` preceding it makes it quite clear what is being referred to.

pmackay: Actually I've noticed the `domain` and `protocol` vars come from the `mortik.nginx-rails` role. So cannot change them just like that. I think `mortik.nginx-rails` assumes a bare domain so am wondering about changing the default nginx config.

pmackay:
Closing without merging as I think several of the useful changes here have been covered by other bits of work, and other changes were not intended for merging. Please say @MikeiLL if you think any parts of this should still be included.

PR #55, opened 2016-07-19
no interactive prompt during language pack install

orangeman:
running the playbooks in latest ubuntu container
during installing ubuntu german language packs
ansible prompts interactive input for locale

This is also happens during manual install
Might be due to ubuntu version difference?

### DIFF
--- a/roles/language/tasks/main.yml
+++ b/roles/language/tasks/main.yml
@@ -11,5 +11,5 @@
   become: yes
 
 - name: reconfigure locales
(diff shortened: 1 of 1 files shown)


pmackay:
What ubuntu version were you using?

orangeman:
ubuntu version 14.04 trusty

pmackay:
I've re-submitted this in https://github.com/openfoodfoundation/ofn-install/pull/68 as I found it too and this change was out of date. Sorry not to have merged earlier!

PR #57, opened 2016-07-19
dockerfy

orangeman:
run playbooks inside docker container

not sure if this is of much use to anyone else.

After struggling to run the playbooks on a debian host
this is what I could get running:
`apt-get install docker`
`docker build .`
`docker run -p 80:80 -p 443:443 <image>`

It needs to run on the server for letsencrypt to work :/

Another way might be to docker run a base image
with sshd pre-installed and also expose port 22
Then run ansible remotely as usual from home

..or have ansible build and deploy docker image(s)...


pmackay:
Keen to see what more can be done with Docker. What's your reason for using it though, how does it help you?

>It needs to run on the server for letsencrypt to work :/

Does the letsencrypt stuff need to be more configurable at all, if not working on localhost?

orangeman:
> how does it help you?
The playbooks did not run on a debian machine.
That's why I used a docker ubuntu base image.

orangeman:
> Does the letsencrypt stuff need to be more configurable at all?

letsencrypt does some challenge response stuff for proof of ownership of the domain. This obviously needs to run on the life host not the local host. Is there a better way to do this?

How do you refresh letsencrypt once the certificates expire?

orangeman:
> be more configurable at all?

hmm... thinking about how to (issue or) renew letsencrypt certificates separately from the ofn provisioning. It would be great to have one nginx reverse proxy handling https/ssl for several domains with different websites or ofn instances running on the same machine / ip address. Also postgres could be shared between instances. Maybe dockerizing the database, the rails stuff and nginx into different containers would provide more flexibility compared to one monolythic all-inclusive playbook...

just thinking out loud. What do you think?

elf-pavlik:
@almereyda wrote docker compose based setup which uses letsencrypt
* https://community.openfoodnetwork.org/t/early-draft-of-a-docker-compose-deployment/764
* https://github.com/allmende/ofn-docker

pmackay:
@elf-pavlik do you have a sense of how these should work together, given it seems completely separate from ofn-install?

elf-pavlik:
@pmackay I will do some more work with @almereyda on using docker to deploy demo/stating for .mx and prepare for live (production) deployment. We should have more insights to share in upcoming weeks.

@orangeman I suggested Jon to have a chat/call with you about deploying OFN and you plans for .de instance https://twitter.com/almereyda

enricostano:
Not sure if this is still of any interest, but if we're going to contribute to OFN it's really probable that we'll use LXC containers to develop OFN, as we do for all the other projects.

If I correctly understand the use case here I strongly recommend LXC containers instead Docker in this case.

This is one of our development environment setup, just to get the idea:
https://github.com/coopdevs/timeoverflow/wiki/Build-development-environment-with-LXC-for-GNU-Linux

pmackay:
@enricostano we've not merged in any docker functionality into ofn-install yet. A key need is probably to maintain DRY where we can, so can you see how LXC support could be integrated into this without radically changing the rest of it? Perhaps create a new issue to discuss?

enricostano:
@pmackay yep, sure. I'm playing a bit with it for now. I'll open an issue and maybe we can study a `development.yml` playbook. The idea is to keep development, staging and production environments as similar as possible.

orangeman:
@enricostano any progress using LXC?

orangeman:
> We should have more insights to share in upcoming weeks.

@elf-pavlik @almereyda any new insights? A chat/call would be great. Sorry for not replying earlier!

@pmackay My main motivation for docker was to make it "easier" to deploy OFN. I'm not a sysadmin. only a (rails) developer. It took me three days to get OFN running the first time. After reconfiguring the vps I gave up the second time. It's the last bits that's hard: the background job runner and memecached..

@maikel Maybe docker is not the right tool? As a non-sysadmin it just looked appealing to "simply" do a _docker-compose up_ and have a new OFN instance running (in a multi host environment). I did try pure ansible first but my vps runs debian and the ansible scripts require ubuntu. After struggling a few days trying to install the ubuntu dependencies I switched to a docker ubuntu image which worked (97%)

What do you think of plain LXC?

enricostano:
@orangeman We finnaly got time to open a PR to use LXC for development environments https://github.com/openfoodfoundation/ofn-install/pull/79

Any input is welcome!

enricostano:
I think we can close this PR and have a look later if we need. Please feel free to ping me if anyone still think this is relevant.

Thanks!

PR #61, opened 2016-12-30
Create stveep_2.pub

stveep:


pmackay:
@stveep do you need both keys installed on a box? Just wondering why have two?

stveep:
No not needed actually, sorry!

PR #62, opened 2017-01-20
Resolve ansible-lint issues and other cleanup changes

pmackay:
@mkllnk would appreciate it if you could review this.


roles/deploy/handlers/main.yml
```diff
@@ -42,3 +44,10 @@
     state: started
   become: yes
   become_user: "{{ unicorn_user }}"
+
+- name: restart nginx
```
mkllnk: Is this re-starting nginx or only starting it when it's not running yet?

pmackay: Well ansible just uses state: started so could it be either? Would you prefer if these tasks were called start xxx?

pmackay:
>Should we mention how to use ansible-lint in the README? Can we add that to Travis?

I've added some notes on it, please check. 

This is already run in Travis.

README.md
```diff
@@ -29,6 +29,20 @@ You will need Ansible on your machine to run the playbooks.
...
+## Code quality
+
+Install [ansible-lint](https://github.com/willthames/ansible-lint) by running:
+```
+pip install ansible-lint
+```
+
+Run the checks using:
+```
+ansible-lint site.yml
```
mkllnk: I think `--exclude=community` is missing. Otherwise you get a lot of errors.

mkllnk: Added it myself. :-)

PR #63, opened 2017-01-20
Update rbenv and nginx roles

pmackay:
Improve some other config.
Update OFN version.

Please also check this over :)


inventory/group_vars/_example.yml
```diff
@@ -10,3 +10,9 @@ language_packages:
   - language-pack-en-base
 l10n_repo: https://github.com/openfoodfoundation/l10n_au.git
 timezone:
+
+developer_email: <dev email>
```
mkllnk: Some comments about the purpose would be nice. And how to keep this private if someone doesn't want to publish their email address.

inventory/group_vars/all.yml
```diff
@@ -15,21 +15,27 @@ ssh_pubkey: "{{ lookup('file', '~/.ssh/{{ ssh_pubkey_name }}') }}"
 protocol: https
 
 git_repo: https://github.com/openfoodfoundation/openfoodnetwork.git
-git_version:  v1.7.0
+git_version:  v1.8.6
```
mkllnk: Do we want to keep updating this? Or do we want to use a "latest" branch with the last release? On the other hand, it's good to document here the last tested version. People may need to update stuff if they want a newer OFN version. And you can always update the OFN after the initial provisioning. So I guess it's good to keep the version number here.

pmackay: OK so you seem to have argued yourself into continuing to update it :)

inventory/group_vars/_example.yml
pmackay: Please check new commit. I'd like to handle the privacy issue by introducing the use of Ansible Vault, if thats ok. We need to have various properties stored secretly and Vault is probably the simplest way to do that?

pmackay:
@mkllnk please confirm if this is ok to merge. I've updated with other 2 commits.

PR #64, opened 2017-01-24
Be able to restore the database using db2fog from last backup

pmackay:
I've added this as a way to trigger a new install being synced directly from an existing backup.

Also @mkllnk do you think https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/0_depricated_db2fog.rb should be deleted?

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -28,8 +28,6 @@ security_autoupdate_enabled: false
 #----------------------------------------------------------------------
 # Rails variables
 ruby: 2.1.5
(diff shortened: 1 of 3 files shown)


mkllnk:
> do you think https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/0_depricated_db2fog.rb should be deleted?

I just created `db2fog.rb` initializer files on our servers. I don't know which other servers need to do that. Maybe that's something for the release notes? We don't have upgrade scripts for out Git push deployment. One point for Ansible deployment.

PR #66, opened 2017-01-28
Improve security config

pmackay:
Makes auto update the default.
Updates community security role.
Removes module from common role that is installed by security role.
Sets email to use for updates.


inventory/group_vars/uk.yml
```diff
@@ -13,6 +13,7 @@ timezone: London
 
 
 developer_email: dev.ofnuk@gmail.com
+security_autoupdate_mail_to: "{{ developer_email }}"
```
mkllnk: Could this line go into all.yml?

PR #69, opened 2017-02-02
add config vars

pmackay:
Some new config variables to be setup in application.yml.

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -16,3 +16,9 @@ CURRENCY: {{ currency }}
 CHECKOUT_ZONE: {{ checkout_zone }}
 
 GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
+
+S3_ACCESS_KEY: {{ s3_access_key }}
+S3_SECRET: {{ s3_secret }}
+
+STRIPE_CLIENT_ID: {{ stripe_client_id }}
+STRIPE_INSTANCE_SECRET_KEY: {{ stripe_instance_secret_key }}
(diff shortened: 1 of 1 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -14,3 +14,11 @@ LOCALE: {{ locale }}
 CURRENCY: {{ currency }}
 # Spree zone.
 CHECKOUT_ZONE: {{ checkout_zone }}
+
+GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
+
+S3_ACCESS_KEY: {{ s3_access_key }}
```
mkllnk: Is this for images or for backups? They could be different.

pmackay: You mean different S3 accounts could be used for images and backups? Why would that be likely? Wouldnt it just be different buckets?

mkllnk: I'm not super familiar with the AWS access management, but it could be that developers have access to the images (public data), but not to the backups (highly sensitive data). And I don't know how people are organising there AWS keys and buckets. If I don't make sense, leave it. You could add a comments to the application.yml.

pmackay:
@enricostano yes I think `application.yml.example` should be updated to capture these better. Keeping this in sync with ofn directly is a bit of a challenge :)

enricostano:
@pmackay so you confirm that those variables are mandatory now?

pmackay:
Sort of:

Maps key needed for map to work.
S3 key needed if using S3 for image storage.
Stripe keys needed if using Stripe (feature not in place yet but will be)

enricostano:
I see that there are some `if` there. Can we make this optional? I'm thinking about something like:
https://github.com/openfoodfoundation/ofn-install/blob/e42f86d66312cef21fac1b2f806c22ad3bd53efe/roles/app/tasks/main.yml#L53

So in `application.yml.example` we can set those variables to `none` by default.

What do you think?

mkllnk:
The Google Maps key is mandatory, I think. You can't deactivate the maps at the moment. S3 and Stripe are optional.

PR #70, opened 2017-02-25
Ubuntu 16.04 test

pmackay:
Changes to support building on Ubuntu 16.04.

### DIFF
--- a/roles/cert/tasks/main.yml
+++ b/roles/cert/tasks/main.yml
@@ -1,0 +1,4 @@
+---
+
+- name: Create certificates using certbot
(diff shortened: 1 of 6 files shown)


roles/cert/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+- name: Create certificates using certbot
+  command: "/opt/certbot/certbot-auto certonly --standalone --agree-tos --email {{ developer_email }} --noninteractive -d {{ domain }}"
```
mkllnk: This needs a condition to run only once. Just saw that in Travis.

PR #73, opened 2017-06-12
Running tmp:cache:clear for i18n-js before assets:precompile

jeronimo:
This is related to Task https://github.com/openfoodfoundation/openfoodnetwork/issues/1562 
for PR https://github.com/openfoodfoundation/openfoodnetwork/pull/1564 because of https://github.com/fnando/i18n-js/issues/306


mkllnk:
I was wondering if we need a playbook to change current post-receive hooks, but I would actually not use it. Our hook is a bit older and contains some other comments. I don't want to overwrite the whole file and a patch may fail. So I would recommend everyone to update their post-receive hooks manually. What do you think?

Anyway, we need something in the release notes when we merge PR openfoodfoundation/openfoodnetwork#1564.

PR #77, opened 2017-08-29
Separate default_user in a new playbook

danypr92:
### Problem
The first time we execute `playbook/provisioning.yml` it does the following things with the `default_user` role:

- connect via SSH with `root` user
- create the default user `ofn-admin`
- copy the ssh key

After that we execute the role `geerlingguy.security` which changes the `/etc/ssh/sshd_config` file to deny access with `root` user.

At this point we cannot run the `playbook/provisioning.yml` playbook again since the `default_user` role would fail trying to access the host with the `root` user.

### Solution
Until now we're just commenting out the `default_user` role but it's not a comfortable solution.

With this PR we propose to call separately the `default_user` role from a new playbook.
The first time we provision a server, we must execute `playbooks/default_user.yml` playbook first of all to create the default user `ofn-admin` and then execute the `playbooks/provision.yml`. All other times you only need to run `playbooks/provisioning.yml` because the user has already been created.


enricostano:
@danypr92 could you please add the new playbook to https://github.com/openfoodfoundation/ofn-install#playbooks? Thanks!

danypr92:
Done

enricostano:
@danypr92 cool, thanks! Now we just need all green tests again and we'll be ready to go.

Issue #80, opened 2017-09-22
Role disappeared: conorsch.create-droplet


danypr92:
https://github.com/openfoodfoundation/ofn-install/blob/bd9fbf9057de4754ab69911dcf050cb0d8c49b5d/playbooks/create_server.yml#L27

Hi,
we noticed that this role disappeared from both Ansible Galaxy and GitHub. 

We opened an issue in other repo of the role owner that uses this disappeared role. 

https://github.com/conorsch/ansible-digitalocean-vpn/issues/4

enricostano:
cc/ @Matt-Yorkley

pmackay:
@danypr92 are you actively using that role? I think I added it to support quickly spinning up Digital Ocean droplets, maybe thats not that much used?

enricostano:
@pmackay we don't use it in Spain/Catalunya instance, I guess only UK uses it but maybe not even anymore.

I vote for removing the specific DO role and its dependencies.

enricostano:
Thanks @danypr92 !

PR #81, opened 2017-09-26
Remove community role conorsch.create-droplet

danypr92:
Fix issue https://github.com/openfoodfoundation/ofn-install/issues/80

The role `conorsch.create-droplet` [disappeared](https://github.com/conorsch/ansible-digitalocean-vpn/issues/4) but it is used in `create_server.yml` playbook. 

We propose to remove `create_server.yml` playbook.

@pmackay @enricostano


orangeman:
if it disappeared then +1 remove it

PR #87, opened 2017-10-25
Compile bugsnag's initializer template

sauloperez:
## What?

Although in https://github.com/openfoodfoundation/ofn-install/blob/master/roles/deploy/tasks/main.yml#L60 we have a step that symlinks the bugsnag config/initializer from the `config_path` to the app's config folder, there was no step at provisioning time that copied the template to `config_path`.

I do it this way because this is the pattern I see in `ofn-install` but it'd be much simpler to have the `config/initializer/bugsnag.rb` in OFN's repo and read the api key from an env var like any other regular Rails app. No need for any extra steps at provisioning and deploy time then.

### DIFF
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -42,6 +42,17 @@
     - { src: "seed.sh.j2", dest: "{{ config_path }}/seed.sh" }
   tags: app_templates
 
+- name: copy the bugsnag's initializer file into the shared folder
+  template:
+    src: "bugsnag.rb.j2"
+    dest: "{{ config_path }}/bugsnag.rb"
+    owner: "{{ unicorn_user }}"
+    mode: 0775
+  tags:
+    - app_templates
+    - bugsnag
+  when: bugsnag_key != "none"
+
 - name: copy the db2fog file into the shared folder
   template:
     src: db2fog.rb.j2
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -42,6 +42,8 @@
     - { src: "seed.sh.j2", dest: "{{ config_path }}/seed.sh" }
   tags: app_templates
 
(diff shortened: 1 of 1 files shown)


sauloperez:
Tested and working in our staging.katuma.org

sauloperez:
Totally agree @enricostano

enricostano:
```
TASK [app : copy the bugsnag's initializer file into the shared folder] ***************************************************************
Sunday 29 October 2017  12:37:58 +0100 (0:00:22.373)       0:08:15.898 ******** 
fatal: [openfoodnetwork.net]: FAILED! => {"failed": true, "msg": "The conditional check 'bugsnag_key != \"none\"' failed. The error was: error while evaluating conditional (bugsnag_key != \"none\"): 'bugsnag_key' is undefined\n\nThe error appears to have been in '/home/enrico/dev/ofn/ofn-install/roles/app/tasks/main.yml': line 47, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# template will be copied\n- name: copy the bugsnag's initializer file into the shared folder\n  ^ here\n"}
```
🤔

sauloperez:
It never happened to me because I defined the var. We must check for the existence of the var first, I guess. Does pulling the latest changes from `katuma_secrets` fix it as well?

enricostano:
This patch fixes it:
```diff
diff --git a/roles/app/tasks/main.yml b/roles/app/tasks/main.yml
index 4ed903f..5745094 100644
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -53,7 +53,7 @@
   tags:
     - app_templates
     - bugsnag
-  when: bugsnag_key != "none"
+  when: bugsnag_key is defined
 
 - name: copy the db2fog file into the shared folder
   template:
```

PR #92, opened 2017-10-31
Fix bugsnag configuration

enricostano:
Fixes issue reported here https://github.com/openfoodfoundation/ofn-install/pull/87#issuecomment-340255922

### DIFF
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -53,7 +53,7 @@
   tags:
     - app_templates
     - bugsnag
-  when: bugsnag_key != "none"
+  when: bugsnag_key is defined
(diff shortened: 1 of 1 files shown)


enricostano:
cc/ @sauloperez

sauloperez:
This should be done as well for `s3_backups_bucket` (see https://github.com/openfoodfoundation/ofn-install/blob/master/roles/deploy/tasks/main.yml#L79). https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/group_vars/es.yml#L22 is why it worked

enricostano:
@sauloperez please open a separate issue 😬

Issue #82, opened 2017-10-05
Deploy fails to restart unicorn


sauloperez:
```shell
RUNNING HANDLER [deploy : restart unicorn step 2] ******************************************************************************************
Thursday 05 October 2017  12:38:14 +0200 (0:00:02.295)       0:04:52.566 ******
fatal: [staging.katuma.org]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service unicorn_openfoodnetwork: Failed to restart unicorn_openfoodnetwork.service: The name org.freedesktop.PolicyKit1 was not provided by any .service files\nSee system logs and 'systemctl status unicorn_openfoodnetwork.service' for details.\n"}
```

enricostano:
This is still broken 😭

elf-pavlik:
Same here, anyone already has strategy for how to deal with it?

enricostano:
@elf-pavlik yep, just log in as `ofn-admin` and run:
`sudo systemctl restart unicorn_openfoodnetwork.service`

agorapy:
It is not enought to have the following (in `roles/deploy/handlers/main.yml`):

```
- name: restart unicorn step 2
  shell: sudo /etc/init.d/unicorn_openfoodnetwork restart
  become: yes
  become_user: "{{ user }}"
  when: table_exists.stderr.find('does not exist') == -1
  # If unicorn isn't actually started yet we probably need this:
  notify: start unicorn

- name: start unicorn
  shell: sudo /etc/init.d/unicorn_openfoodnetwork start
  become: yes
  become_user: "{{ user }}"
```

This calls the `roles/webserver/templates/unicorn_init.j2` script which seems to work nice.

enricostano:
@agorapy I don't think so, `restart` also `start` a service if not started already. The problem is with permissions. But thanks for the help!

elf-pavlik:
I do some research on this issue, sharing some notes here:

@MikeiLL posted this question here but haven't received helpful answers: https://serverfault.com/questions/733084/running-service-with-systemd-as-default-user-via-ansible-in-ec2-requests-passwo

Based on this gist, sudoers file for `openfoodnetwork` will not play well with ansible service module https://gist.github.com/nanobeep/3b3d614a709086ff832a

We could consider running unicorn as systemd user service: https://www.brendanlong.com/systemd-user-services-are-amazing.html

[...]

PR #103, opened 2017-12-22
Unicorn user sudoers

enricostano:
Fixes https://github.com/openfoodfoundation/ofn-install/issues/82

### WAT
We use `openfoodnetwork` user to run unicorn process, but it can't manage the process itself. This breaks the unicorn restart after the deploy and is really annoying. The current workaround is to log in with `ssh-admin` user and run the `systemctl` commands.

With this PR we allow `openfoodnetwork` to manage `unicorn_openfoodnetwork.service`.


enricostano:
cc/ @lin-d-hop @danypr92

roles/unicorn_user/tasks/main.yml
```diff
@@ -33,3 +33,10 @@
     owner: "{{ unicorn_user }}"
     group: "{{ unicorn_user }}"
     mode: 0644
+
+- name: Create sudoers configuration for unicorn user
+  become: true
+  template:
+    src: sudoers.j2
+    dest: "/etc/sudoers.d/90-unicorn"
```
sauloperez: I'd add a different number other than the one in the existing sudoers file to make a bit more obvious what the actual order is. According to the man page, having numbers repeated in filenames doesn't cause any problem though.

sauloperez:
Tested and working :+1:

PR #106, opened 2018-01-16
Delayed_job systemd unit

enricostano:


roles/webserver/templates/delayed_job.service.j2
```diff
@@ -0,0 +1,17 @@
...
+
+[Service]
+Type=simple
+User=openfoodnetwork
+WorkingDirectory={{ current_path }}
+Environment=RAILS_ENV={{ rails_env }}
+SyslogIdentifier=openfoodnetwork-delayed_job
+PIDFile={{ delayed_job_pid }}
+
+ExecStart=script/delayed_job
```
sauloperez: start? stop?

sauloperez:
What's left here @enricostano . Something I can help you with?

roles/webserver/templates/delayed_job.service.j2
```diff
@@ -0,0 +1,17 @@
+# Open Food Network delayed_job systemd service
+
+[Unit]
+Description=Open Food Network delayed_job
+
+[Service]
+Type=simple
```
enricostano: Not sure if `simple` or `forking`

roles/webserver/templates/delayed_job.service.j2
sauloperez: This line should contain `script/delayed_job start`. We're also missing an `ExecStop`

elf-pavlik: https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files#unit-specific-section-directives
> *ExecStop=*: This indicates the command needed to stop the service. If this is not given, the process will be killed immediately when the service is stopped.

Do we need to run `script/delayed_job start` or can just let it get killed? I don't really know how it uses PIDs etc.
Possibly relevant: http://www.rubydoc.info/gems/delayed_job/4.1.4#Running_Jobs

roles/webserver/tasks/main.yml
```diff
@@ -12,13 +12,20 @@
...
-- name: update unicorn
+- name: Enable unicorn unit
   service:
     name: "unicorn_{{ app }}"
     enabled: yes
   become: yes
-  tags: unicorn
+  tags: init
+
+- name: Enable delayed_job unit
```
elf-pavlik: It probably needs to become root just like in #144 or run as user service like in #143 
Depends on how we resolve #82

roles/webserver/templates/delayed_job.service.j2
enricostano: We need `start` in `ExecStart` for sure, and I would add `ExecStop` just to be consistent with DelayedJob API.

roles/webserver/tasks/main.yml
enricostano: Probably we'll need to become `root`

Matt-Yorkley:
It would be great to get this systemd unit working. We had more issues today.

roles/webserver/tasks/main.yml
Matt-Yorkley: With the current script for restarting it we run it as the unicorn user and it works fine...

roles/webserver/templates/delayed_job.service.j2
Matt-Yorkley: We can consider using this under `[Service]` as well: 
```
Restart=always
RestartSec=30
```

sauloperez:
WAT? what was it @Matt-Yorkley ??

Matt-Yorkley:
I've pushed an extra commit for this at https://github.com/coopdevs/ofn-install/pull/4 to finish it off and make `monit` obsolete.

sauloperez:
and how? I'm a bit curious. Was it the recovering option after 60s of being dead?

Matt-Yorkley:
The unit has the settings `Restart=on-failure` and `RestartSec=60` which tell systemd to monitor the service and restart it when something goes wrong, so now systemd is natively doing the old job of the third-party monit service. 

I wasn't sure about the timer of 60 seconds, it's recommended to not set it too low, and the script we run as part of this service takes a little while to execute because it goes via ruby and rbenv, so I chose 60 seconds. Monit was previously checking every 120 seconds.

sauloperez:
Should we close this one in favor of the one you opened?

Matt-Yorkley:
Closing as now covered by #192.

Issue #105, opened 2018-01-12
Create certificates using certbot task fails


enricostano:
### WHAT
I'm not sure since when, provisioning a server that has already been provisioned before will fail running the task [Create certificates using certbot](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/cert/tasks/main.yml#L3) due to missing permissions.

### TODO
 - [ ] Check if this happen also to brand new servers
 - [ ] Check if this task is correct / needed / idempotent enough
 - [ ] Check if something changed in `certbot`

### LOGS
```
TASK [cert : Create certificates using certbot] ***************************************************************************************
task path: /home/enrico/dev/ofn/ofn-install/roles/cert/tasks/main.yml:3
Friday 12 January 2018  12:03:06 +0100 (0:00:01.967)       0:05:06.602 ******** 
Using module file /usr/lib/python2.7/site-packages/ansible/modules/commands/command.py
<212.47.234.234> ESTABLISH SSH CONNECTION FOR USER: ofn-admin
<212.47.234.234> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=ofn-admin -o ConnectTimeout=10 -o ControlPath=/home/enrico/.ansible/cp/%h-%r 212.47.234.234 '/bin/sh -c '"'"'echo ~ && sleep 0'"'"''
<212.47.234.234> (0, '/home/ofn-admin\n', '')
<212.47.234.234> ESTABLISH SSH CONNECTION FOR USER: ofn-admin
[...]

PR #117, opened 2018-02-08
update geerlingguy.certbot to 3.0.0 and also use it to generate cert …

elf-pavlik:
…with standalone

Before merging it we may want to make sure #105 gets resolved so we can properly test the whole playbook. This version fails in exactly the same way as the old one and works if `--webroot -w` + `python -m SimpleHTTPServer 80` used instead of `--standalone`, just as the old version do.

---
This change is [Reviewable](https://reviewable.io/reviews/openfoodfoundation/ofn-install/117)

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -48,12 +48,16 @@
       db_user_roles: CREATEDB
 
     - role: geerlingguy.certbot
-      become: yes
+      certbot_install_from_source: yes
       certbot_auto_renew_user: "{{ unicorn_user }}"
-      when: protocol == "https"
-      tags: cert
-
-    - role: cert
+      certbot_create_if_missing: yes
+      certbot_create_method: standalone
+      certbot_create_standalone_stop_services: []
+      certbot_admin_email: "{{ developer_email }}"
+      certbot_certs:
+        - domains:
+          - "{{ domain }}"
+      become: yes
       when: protocol == "https"
       tags: cert
 
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -10,4 +10,4 @@
   version: v2.7.4
 
 - src: geerlingguy.certbot
-  version: 1.0.1
+  version: 3.0.0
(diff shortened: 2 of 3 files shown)


enricostano:
Does this also solve the issue related to `certbot` trying to use port `443` while nginx is using it? I'm investigating the `--nginx` option in another project with pretty similar stack. Will report back once done.

playbooks/provision.yml
```diff
@@ -48,12 +48,16 @@
...
-      become: yes
+      certbot_install_from_source: yes
       certbot_auto_renew_user: "{{ unicorn_user }}"
-      when: protocol == "https"
-      tags: cert
-
-    - role: cert
+      certbot_create_if_missing: yes
+      certbot_create_method: standalone
+      certbot_create_standalone_stop_services: []
```
enricostano: we should add `nginx` here, right?

elf-pavlik: I think the *standalone* runs only once when cert doesn't exist yet. In playbooks here nginx gets installed after later one. I think only the renewal command might need to use --nginx since it shouldn't use *standalone* any more but rely on certbot *nginx* plugin. In that case nginx using ports 80 and 443 makes sense since certbot should not try to spin a standalone we server any more.

elf-pavlik:
> Does this also solve the issue related to `certbot` trying to use port `443` while nginx is using it?

Could you drop a link to that issue?

enricostano:
So, it seems that the `nginx` plugin is not supported by this ansible role https://github.com/geerlingguy/ansible-role-certbot/issues/2#issuecomment-365369540

I'm planning to just write a simple custom role that installs `certbot` and the `nginx` plugin.

In this way we'll probably solve a bunch of problems that we currently have with `certbot`.

Here is a link to what I'll propose https://github.com/coopdevs/timeoverflow-provisioning/pull/28/files#diff-558aaf2e3af5ea64c2ee8d1bee47756c

elf-pavlik:
superseded by #124

PR #122, opened 2018-02-14
Move seed data to the app

sauloperez:
Moving seed data script to the app as being totally related to the app logic, there is where it belongs. It does not make sense having to open pull requests on the ansible scripts to deal with business logic. Besides, the seed data is also needed in development.

This removes one piece of this entropy puzzle we have with seed data in OFN. From now on, we'll have a single script to manage seed data.

This is the first step of https://github.com/openfoodfoundation/openfoodnetwork/issues/2071 and https://github.com/openfoodfoundation/openfoodnetwork/issues/2072


roles/app/files/seeds.rb
```diff
@@ -1,29 +0,0 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
-require 'yaml'
-require 'csv'
-
-# -- Spree
-unless Spree::Country.find_by_iso(ENV['DEFAULT_COUNTRY_CODE'])
```
sauloperez: this script feels a bit more decoupled from Australia, so I might bring it to the app as well

enricostano:
The whole point of having seed data in `ofn-install` was to have instance specific data. For instance countries and states. Until now there was a silent agreement on using AUS defaults in development.

I agree on moving these files to the app repo but we need to take in account that different instances will need different seed data. If not to poulate `Enterprise` or `Spree::User`, surely to feed `Spree::Country` or (hopefully) `MailMethod`, `Taxonomy` and so on.

sauloperez:
> I agree on moving these files to the app repo but we need to take in account that different instances will need different seed data.

That's the whole point of https://github.com/openfoodfoundation/openfoodnetwork/issues/2071

sauloperez:
I think I got the whole picture now. This is probably written down thousand times already but bear with me again; it'll help me shape my thoughts.

So, we have this [roles/app/files/seeds.rb](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/files/seeds.rb) that is executed when deploying and also its app's counterpart [db/seeds.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/db/seeds.rb). They are almost the same but while the first one is localized the second one is not and it simply inserts the Australian states into the DB.

I noticed that in the app we already have a file with the Australian states in [db/default/spree/states.yml](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/db/default/spree/states.yml), located where [roles/app/files/seeds.rb](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/files/seeds.rb) loads them from.

What I suggest in order to simplify all this and sucessfully get rid of the [roles/app/files/seeds.rb](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/files/seeds.rb) is to fix `db/default/spree/states.yml` so that it has the exact same structure as [l10n_au/blob/master/states.yml](https://github.com/openfoodfoundation/l10n_au/blob/master/states.yml). Then, we can move `roles/app/files/seeds.rb` to `app/db/seeds.rb`. That would simplify things quite a lot.

Next step could be to symlink [db/default/spree/states.yml](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/db/default/spree/states.yml) to wherever the l10n repo might be on your machine.

enricostano:
Make sure `db/default/spree/states.yml` is not used by Spree

https://github.com/spree/spree/blob/2-0-stable/core/db/default/spree/states.rb

mkllnk:
Sounds logical to keep the states in the l10n repos and use them. A default should still stay in the ofn project so that a dev doesn't need an l10n repo to run the site.

sauloperez:
I provisioned and deployed https://github.com/openfoodfoundation/openfoodnetwork/pull/2123 and it works. I just some other instance that has the l10n repo in order to give it a look. We don't have our own l10n repo in Katuma yet.

pacodelaluna:
Finally tested too, working fine for me too.

sauloperez:
:tada:

PR #123, opened 2018-02-23
Added AVAILABLE_LOCALES env var in ofn-install

mllocs:
Added the AVAILABLE_LOCALES env var in the application.yml template so it can be defined in every instance. I also set the available_locales to "es,ca,pt" in Katuma (es.yml).

 - [x] Test without `available_locales` to see what happens (OR should work)

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -9,6 +9,7 @@ TIMEZONE: {{ timezone }}
 DEFAULT_COUNTRY_CODE: {{ country_code }}
 # Locale for translation.
 LOCALE: {{ locale }}
+AVAILABLE_LOCALES: {{ available_locales }}
 
 # Find currency codes at http://en.wikipedia.org/wiki/ISO_4217.
 CURRENCY: {{ currency }}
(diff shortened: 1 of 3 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -9,6 +9,7 @@ TIMEZONE: {{ timezone }}
 DEFAULT_COUNTRY_CODE: {{ country_code }}
 # Locale for translation.
 LOCALE: {{ locale }}
+AVAILABLE_LOCALES: {{ available_locales }}
```
enricostano: What about group vars that don't specify `available_locales`?

mllocs:
so, if we want english in Katuma we need to add it here https://github.com/openfoodfoundation/ofn-install/pull/123/files#diff-cf20333b737bfac9bf3fece3370e7a3eR7

sauloperez:
I believe so

enricostano:
I tested this in our staging, it works. Many :heart: to @mllocs

mllocs:
boaaaaaaaaaaaa

sauloperez:
Is that portuguese :trollface:?

mllocs:
screen shot 2018-03-14 at 15 59 42

sauloperez:
Lol

Issue #86, opened 2017-10-24
Change crontab user for Letsencrypt renewal


enricostano:
## WAT!? Why!?
Right now we [pass](https://github.com/openfoodfoundation/ofn-install/blob/master/playbooks/provision.yml#L52) `openfoodnetwork` to the [cerbot](https://github.com/geerlingguy/ansible-role-certbot) role as the crontab user for Letsencrypt renewal cron job.
```
openfoodnetwork@staging:~$ crontab -l
#Ansible: Certbot automatic renewal.
30 3 * * * /opt/certbot/certbot-auto renew --quiet --no-self-upgrade
```
The problem is that `openfoodnetwork` user doesn't have permissions to run the certbot binary. Give a look at the CRON log:
```
oct 24 03:30:01 scw-ecce22 CRON[10274]: (openfoodnetwork) CMD (/opt/certbot/certbot-auto renew --quiet --no-self-upgrade)
oct 24 03:30:01 scw-ecce22 sudo[10278]: pam_unix(sudo:auth): auth could not identify password for [openfoodnetwork]
oct 24 03:30:01 scw-ecce22 sudo[10278]: openfoodnetwork : user NOT in sudoers ; TTY=unknown ; PWD=/home/openfoodnetwork ; USER=
oct 24 03:30:03 scw-ecce22 sudo[10299]: pam_unix(sudo:auth): auth could not identify password for [openfoodnetwork]
oct 24 03:30:03 scw-ecce22 sudo[10299]: openfoodnetwork : user NOT in sudoers ; TTY=unknown ; PWD=/home/openfoodnetwork ; USER=
oct 24 03:30:03 scw-ecce22 CRON[10270]: (CRON) info (No MTA installed, discarding output)
oct 24 03:30:03 scw-ecce22 CRON[10270]: pam_unix(cron:session): session closed for user openfoodnetwork
```

## Proposal
Let's use `ofn-admin` user instead since it has permissions to run it. There is still a pending problem about the `standalone` web server but it will be covered in another issue. Will link to this one once created.

enricostano:
> On the server I just provisioned, the root user already had the crontab entry.
Who created it? Are you using latest `ofn-install`?

mkllnk:
https://certbot.eff.org/all-instructions/#ubuntu-16-10-yakkety-nginx
> The Certbot packages on your system come with a cron job that will renew your certificates automatically before they expire.

mkllnk:
[...]

PR #124, opened 2018-03-08
Use coopdevs role to install certbot and nginx plugin

enricostano:
### WAT
So, it seems that in the role that we're using now there is no support for the nginx plugin for renewal https://github.com/geerlingguy/ansible-role-certbot/issues/2#issuecomment-365369540

Since this is something that we need to address ASAP we created an [Ansible role](https://github.com/coopdevs/certbot_nginx) to just:
1. Install `certbot` and `nginx` plugin
2. Generate the certificate if needed

We also remove our `cert` role since it's just creating confusion. It was generating the certificate, let's do that in the external role (same strategy as geerling's role)

Solves https://github.com/openfoodfoundation/ofn-install/issues/86
Solves https://github.com/openfoodfoundation/ofn-install/issues/105
Supersedes https://github.com/openfoodfoundation/ofn-install/pull/117

### DIFF
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -1,4 +1,3 @@
-
 # from galaxy
 - src: geerlingguy.security
   version: 1.5.0
@@ -9,5 +8,5 @@
 - src: jdauphant.nginx
   version: v2.7.4
 
-- src: geerlingguy.certbot
-  version: 1.0.1
+- src: coopdevs.certbot_nginx
+  version: 0.0.1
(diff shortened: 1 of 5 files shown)


bin/requirements.yml
```diff
@@ -9,5 +8,5 @@
 - src: jdauphant.nginx
   version: v2.7.4
 
-- src: geerlingguy.certbot
-  version: 1.0.1
+- src: coopdevs.certbot_nginx
```
sauloperez: where can I see this role?

enricostano: Sorry, added https://github.com/coopdevs/certbot_nginx to description

elf-pavlik:
:+1:

elf-pavlik:
Seeing #150, what stays in the way of getting this one merged?

For me it solves other issues which #150 doesn't and if this one lands the other PR becomes unnecessary.

Strong :+1: from me to get this PR here merged ASAP

PR #134, opened 2018-03-15
Add mail settings' env vars to unicorn_user

sauloperez:
## What?

these ENV vars provide the data required for https://github.com/openfoodfoundation/openfoodnetwork/pull/2148 to work and need to be specified per instance and environment.

**MAKE SURE YOU SPECIFY AND PROVISION THEM IF YOU MANAGE AN OFN INSTANCE**

## Missing
- [x] Update https://github.com/openfoodfoundation/ofn-install/wiki/Sending-emails-from-OFN

### DIFF
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -4,3 +4,8 @@ domain: example.com
 rails_env: production
 
 admin_email: admin@example.com
+
+mail_host: smtp.mandrillapp.com
+mail_port: 587
+
+smtp_username: Coopdevs
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -5,6 +5,7 @@ rails_env: production
 
 admin_email: admin@example.com
 
+mail_domain: example.com
 mail_host: smtp.mandrillapp.com
 mail_port: 587
 
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -6,7 +6,3 @@ rails_env: production
 admin_email: admin@example.com
 
 mail_domain: example.com
-mail_host: smtp.mandrillapp.com
-mail_port: 587
-
-smtp_username: Coopdevs
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -25,4 +25,7 @@ newrelic_license_key:
 skylight_authentication: xxx
 
 # Mail settings
+mail_host:
+mail_port:
+smtp_username:
 smtp_password:
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -23,3 +23,6 @@ newrelic_license_key:
 
 # Skylight's API key. This enables performance instrumentation through Skylight. See https://github.com/openfoodfoundation/openfoodnetwork/pull/2070 for details.
 skylight_authentication: xxx
+
+# Mail settings
+smtp_password:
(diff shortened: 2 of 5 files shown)


myriamboure:
@sauloperez shall we add that somewhere in the wiki? 
Ping @pacodelaluna for information that will be useful for you to fix staging email things.

sauloperez:
That's a good point. We'll have to update https://github.com/openfoodfoundation/ofn-install/wiki/Sending-emails-from-OFN

elf-pavlik:
Using `ofn-install` with commit from this PR and `master` of `openfoodnetwork` I ended up in situation that when I try to log in with `admin_email` i get message that i need to confirm my account and option to re-send verification (but it doesn't have any mail method yet). Does this PR plus  openfoodfoundation/openfoodnetwork#2148 provide enough solution or we should make extra PR to make user with `admin_email` verified?

sauloperez:
Nope, https://github.com/openfoodfoundation/openfoodnetwork/pull/2148 only creates a mail method ready to be used and nothing more. What you might want instead is https://github.com/openfoodfoundation/openfoodnetwork/pull/2117. As you can see `admin_email` is not confirmed automatically, however I noticed that when I deploy no user with `admin_email` gets created, at least in my case. I'm afraid https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/seed.sh.j2 might not be working. Is that your case as well @elf-pavlik ?

inventory/host_vars/_example.com/config.yml
```diff
@@ -4,3 +4,9 @@ domain: example.com
 rails_env: production
 
 admin_email: admin@example.com
+
+mail_domain: example.com
+mail_host: smtp.mandrillapp.com
+mail_port: 587
```
enricostano: For security reasons I would move host and port to secrets

inventory/host_vars/_example.com/config.yml
```diff
@@ -4,3 +4,9 @@ domain: example.com
...
+
+smtp_username: Coopdevs
```
enricostano: Also this one

elf-pavlik:
>  As you can see `admin_email` is not confirmed automatically, however I noticed that when I deploy no user with `admin_email` gets created, at least in my case.

I just run a fresh deployment and it did create and account using `admin_email` but not confirmed. Since openfoodnetwork `master` doesn't have  openfoodfoundation/openfoodnetwork#2148 merged yet and i couldn't login without confirmation, I ended up confirming account from rails console.

I'll create an issue to auto confirm account created with `admin_email` so that after deployment one doesn't have to deal with and can log in right away.

elf-pavlik:
Ok, I've created #139 to address auto confirming account with `admin_email`

During my deployment I also encountered an issue related to the mailer, if it doesn't relate to this PR I can also create separate issue.

Since current `master` of openfoodnetwork doesn't include openfoodfoundation/openfoodnetwork#2148 I've created mailer method 'the old way' using UI. After doing that it still couldn't send emails since it tried using `no-reply@example.com` as `From`. As workaround I just used rails console to set `mails_from` in Spree config
```ruby
Spree.config do |config|
  config.mails_from = 'hola@ofn.example'
end
```

After that it could send account confirmation email. I also restarted unicorn and delayed_job after changing config above. Before changing Spree config manually I also restarted both after creating new mailer method from the UI but it still used `no-reply@example.com` as `From`

sauloperez:
> Before changing Spree config manually I also restarted both after creating new mailer method from the UI but it still used no-reply@example.com as From

Sorry, I don't quite understand the exact steps you took. In any case, I'm not entirely sure that's the way to directly update the email settings that get stored in the cache. Is there any change if you do the following instead?

```ruby
Spree::MailMethod.current.update_attributes(preferred_mails_from: 'hola@ofn.example')
```

That will go through Spree's preferences API.

inventory/host_vars/_example.com/config.yml
sauloperez: Done in https://github.com/coopdevs/katuma_secrets/pull/3

sauloperez:
Tested again with a fresh provision and deploy and working successfully. I got the test email.

PR #136, opened 2018-03-19
Set master as the default branch to deploy

### LINKED ISSUE #111: git_version set to 1.8.6 in all.yml
https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/group_vars/all.yml

```yml
git_version: v1.8.6
```

this post provides some suggestions https://stackoverflow.com/a/29157519

I would lean towards having a branch in `openfoodnetwork` repo named something like `production`, `stable`, `release` (whatever) and always have it pointing to the latest tag.

In [gitflow](http://nvie.com/posts/a-successful-git-branching-model/) one could just use `master` but ofn seems to use `master` like gitflow `development`

sauloperez:
## What?

Closes #111 . This a more useful default than an old tag

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -15,7 +15,7 @@ ssh_pubkey: "{{ lookup('file', '~/.ssh/{{ ssh_pubkey_name }}') }}"
 protocol: https
 
 git_repo: https://github.com/openfoodfoundation/openfoodnetwork.git
-git_version:  v1.8.6
+git_version: master
 
 l10n_git_version: HEAD
 


inventory/group_vars/all.yml
```diff
@@ -15,7 +15,7 @@ ssh_pubkey: "{{ lookup('file', '~/.ssh/{{ ssh_pubkey_name }}') }}"
 protocol: https
 
 git_repo: https://github.com/openfoodfoundation/openfoodnetwork.git
-git_version:  v1.8.6
+git_version: master
```
enricostano: Did you mean `'master'`? :grimacing:

enricostano: Ignore me. Good to go.

sauloperez: Lol

PR #137, opened 2018-03-19
Provide SKYLIGHT_AUTHENTICATION to unicorn user

### LINKED ISSUE #120: Provision SKYLIGHT_AUTHENTICATION in production
We need to provision the `SKYLIGHT_AUTHENTICATION` env var in order for https://github.com/openfoodfoundation/openfoodnetwork/pull/2070 to work. Otherwise, no metrics are sent to Skylight.

sauloperez:
## What?

Closes #120 

As described in https://www.skylight.io/support/advanced-setup#enabling-additional-environments, by setting `SKYLIGHT_AUTHENTICATION`, we'll enable skylight to gather instrumentation data in production. 

`skylight_authentication` has been set as secret already in https://github.com/coopdevs/katuma_secrets/pull/1 and https://github.com/coopdevs/katuma_secrets/pull/2 for Katuma.

Once we get this working other instances can set their own.

### DIFF
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -4,3 +4,5 @@ domain: example.com
 rails_env: production
 
 admin_email: admin@example.com
+
+skylight_authentication: xxx
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -4,5 +4,3 @@ domain: example.com
 rails_env: production
 
 admin_email: admin@example.com
-
-skylight_authentication: xxx
--- a/roles/unicorn_user/templates/defaults.j2
+++ b/roles/unicorn_user/templates/defaults.j2
@@ -1,2 +1,5 @@
 RAILS_ENV={{ rails_env }}
-SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
+
+{% if skylight_authentication is defined %}
+  SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
+{% endif %}
--- a/roles/unicorn_user/templates/defaults.j2
+++ b/roles/unicorn_user/templates/defaults.j2
@@ -1,1 +1,2 @@
 RAILS_ENV={{ rails_env }}
+SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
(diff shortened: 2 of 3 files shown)


inventory/host_vars/_example.com/config.yml
```diff
@@ -4,3 +4,5 @@ domain: example.com
 rails_env: production
 
 admin_email: admin@example.com
+
+skylight_authentication: xxx
```
danypr92: Why this var is in `config.yml`? It should be declared in `secrets.yml`, no?

roles/unicorn_user/templates/defaults.j2
```diff
@@ -1 +1,2 @@
 RAILS_ENV={{ rails_env }}
+SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
```
enricostano: What happens if the environment variable is not defined? Ruby crash?

And what if the ansible variable is not declared anywhere? Ansible crash?

sauloperez: good points

sauloperez: When `SKYLIGHT_AUTHENTICATION` is not defined you'll see the following in `log/skylight.log`:

```
[SKYLIGHT] [1.5.0] Unable to start Instrumenter; msg=authentication token required; class=Skylight::ConfigError
```
When its value is empty instead () you see the following in the `log/staging.log`:

```
[SKYLIGHT] [1.5.0] Unable to start, see the Skylight logs for more details
```

Skylight seems pretty robust. So far I've seen meaningful error messages in every situation :muscle: .

sauloperez: If the ansible var `skylight_authentication` is not declared you get:

```
(...)
TASK [unicorn_user : Write OFN environment variables defaults] *****************************************************************************
Tuesday 20 March 2018  12:28:35 +0100 (0:00:03.414)       0:00:25.459 ********* 
fatal: [staging.katuma.org]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'foo' is undefined"}
        to retry, use: --limit @/home/pau/dev/ofn-install/playbooks/provision.retry
(...)
```

enricostano: OK for the Skylight warning.

Not so OK for the Ansible failure. Since it's an optional feature some instance could be not interested and shouldn't bother at all about it. `when` conditional in Ansible could help?

sauloperez:
No crash when you `skylight_authentication` in the secrets file now :ok_hand:

PR #140, opened 2018-03-23
Pull translations while deploying

pacodelaluna:
We have always to update translations manually while deploying on french side. I guess the other way to do this is to merge directly translations to the master branch, but should we really add a new commit each time a translation changed?

By the way, it is my first task with Ansible, so I am lacking best practices I guess, tell if if any issue at this side.

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -141,6 +141,14 @@
   become_user: "{{ unicorn_user }}"
 
 #--------------------
+# Update translations
+
+- name: get latest translations from Transifex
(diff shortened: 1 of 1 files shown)


sauloperez:
That's an interesting approach. I wonder what needs to happen with the git branch then? I guess the translation will be available to the app but they will have to be brought to master as well. So should we keep opening a Transifex PR?

pacodelaluna:
Well, it was just an idea. I will push a new PR with french translations to have an updated master branch.

sauloperez:
As long as we keep master up to date this seems good to me.

sauloperez:
The latest changes in our Transifex integration should solve this kind of issues.

PR #141, opened 2018-03-23
Restart delayed job after running db:seed

sauloperez:
This ensures new Spree configuration values will be picked up. In any case, it's safer to restart the services at the end of the deployment so we don't miss any configuration or code to be reloaded.

Apart from that, it's simpler to make it restart rather stopping and relying on monit to bring it up.

### DIFF
--- a/roles/deploy/handlers/main.yml
+++ b/roles/deploy/handlers/main.yml
@@ -14,7 +14,7 @@
 
 
 - name: restart delayed job workers
-  command:  bash -lc "./script/delayed_job -i 0 stop RAILS_ENV={{ rails_env }}" chdir="{{ current_path }}"
+  command:  bash -lc "./script/delayed_job -i 0 restart RAILS_ENV={{ rails_env }}" chdir="{{ current_path }}"
 
 
 - name: restart unicorn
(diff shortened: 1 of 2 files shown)


sauloperez:
I just investigate a bug raised by Bugsnag that requires this

screenshot from 2018-03-22 18-42-48

The problem and solution are mentioned in https://github.com/Mange/roadie-rails/issues/62

sauloperez:
Working

before deploy:
```
$ ps auxf | grep delayed
ofn-adm+  1851  0.0  0.0  15476  1004 pts/0    S+   13:50   0:00  |           \_ grep --color=auto delayed
openfoo+  1789  0.6 11.6 1219808 239368 ?      Sl   13:48   0:00 delayed_job
```
after deploy:
```
ofn-admin@staging-katuma:~$ ps auxf | grep delayed
ofn-adm+  1855  0.0  0.0  15476   908 pts/0    S+   13:50   0:00  |           \_ grep --color=auto delayed
openfoo+  1789  0.5 11.6 1219808 239376 ?      Sl   13:48   0:00 delayed_job
```

PR #143, opened 2018-03-24
make unicorn a user service

elf-pavlik:
**WIP** - early work to discuss this approach

Fixes #82 

Initial work on making unicorn service a user service and allow `unicorn_user` user to control it using [ansible systemd module](http://docs.ansible.com/ansible/latest/systemd_module.html)

It needs more testing and improving how `uid` of `unicorn_user` gets registered (currently twice - once in provision and once in deploy)

So far I tested it manually first making changes and re-running `provision.yaml` and later the same with 'deploy.yaml' until all tasks succeeded. Even though `restart nginx` also succeeded I needed to restart it manually before the site started working.

---
This change is [Reviewable](https://reviewable.io/reviews/openfoodfoundation/ofn-install/143)

PR #144, opened 2018-03-24
manage services as root

### LINKED ISSUE #82: Deploy fails to restart unicorn

elf-pavlik:
Fixes #82 

Another approach than #143 , this one IMO much simpler and works fully including restarting nginx

I tested it together with #124 and `site.yaml` run fully without any fails resulting in functioning site (minus delayed jobs running)

---
This change is [Reviewable](https://reviewable.io/reviews/openfoodfoundation/ofn-install/144)

### DIFF
--- a/roles/deploy/handlers/main.yml
+++ b/roles/deploy/handlers/main.yml
@@ -32,7 +32,7 @@
     name: unicorn_{{ app }}
     state: restarted
   become: yes
-  become_user: "{{ unicorn_user }}"
+  become_user: root
   when: table_exists.stderr.find('does not exist') == -1
   # If unicorn isn't actually started yet we probably need this:
   notify: start unicorn
@@ -42,11 +42,11 @@
     name: unicorn_{{ app }}
     state: started
   become: yes
-  become_user: "{{ unicorn_user }}"
+  become_user: root
 
 - name: restart nginx
   service:
     name: nginx
-    state: started
+    state: restarted
   become: yes
-  become_user: "{{ unicorn_user }}"
+  become_user: root


roles/unicorn_user/tasks/main.yml
```diff
@@ -34,9 +34,3 @@
     group: "{{ unicorn_user }}"
     mode: 0644
 
-- name: Create sudoers configuration for unicorn user
-  become: true
-  template:
-    src: sudoers.j2
-    dest: "/etc/sudoers.d/{{ unicorn_user }}"
-    mode: 0440
```
sauloperez: If we do this we won't be able to restart the unicorn when logging in as the unicorn user, which is something we do often.

elf-pavlik:
In https://github.com/openfoodfoundation/ofn-install/issues/82#issuecomment-362321105 @enricostano suggests as a work around to restart unicorn service manually as `{{ user }}` using `sudo` which has `{{ user }} ALL=(ALL) NOPASSWD:ALL`.

Running unicorn as user service #143 would work different but still the super user has to manage nginx service. @sauloperez where do you see 'giving up on security` in this PR?

sauloperez:
I'm not a sysadmin myself but having to use root instead of a user with specific sudo permissions feels a bit too broad and dangerous.

elf-pavlik:
`{{ user }}` (`ofn-admin` by default) has `{{ user }} ALL=(ALL) NOPASSWD:ALL` and `deploy.yaml` playbook includes `remote_user: "{{ user }}"` so it runs with a power to act as superuser without need to provide password.

Systemd unit still includes `User=openfoodnetwork` and runs process as `openfoodnetwork`. It seems like we compare
1. SSH manually as `{{ user }}` (`ofn-admin`) and manage unicorn service manually using `sudo` - current workaround suggested in https://github.com/openfoodfoundation/ofn-install/issues/82#issuecomment-362321105
2. SSH automatically using ansible with `remote_user: "{{ user }}"` (`ofn-admin`) and in playbook use
```yaml
become: yes
become_user: root
```
which actually might use `sudo` as default `become_method`

Given above, to my understanding the current practice https://github.com/openfoodfoundation/ofn-install/issues/82#issuecomment-362321105 and approach in this PR seem equivalent in security.

Besides @enricostano should we ask anyone else to address this security related concern?

sauloperez:
> SSH manually as {{ user }} (ofn-admin) and manage unicorn service manually using sudo - current workaround suggested in #82 (comment)

just to bring context, we've been using `openfoodnetwork` user to SSH manually since https://github.com/openfoodfoundation/ofn-install/commit/62a812bdb8bfc6a8dbb4f97564415ef3e07557b2. When that falls short we do use `ofn-admin`. That's why I still don't why we can't restart the unicorn with that user from Ansible...

Let's see what @enricostano thinks about all this.

elf-pavlik:
> That's why I still don't why we can't restart the unicorn with that user from Ansible...

In https://github.com/openfoodfoundation/ofn-install/issues/82#issuecomment-375816905 I shared link to https://gist.github.com/nanobeep/3b3d614a709086ff832a which talks about that issue.

> we've been using `openfoodnetwork` user to SSH manually since 62a812b.

`deploy.yml` still SSH uses `remote_user: "{{ user }}"` - `ofn-admin`, why does it have to use `openfoodnetwork` to restart unicorn service then?

BTW in #143 (alternative to this PR) I run and manage unicorn as `opetrnfoodnetwork` user, but then still superuser has to restart nginx.

> Agree though, that while we don't the ideal solution, this might be a proper tradeoff that lets us move.

I don't think this PR lowers security in any way, if someone has clear requirements on how to increase security I think on can follow up with another PR or at least Issue tracking those requirements and providing place to discuss it before making another PR.

> Let's see what @enricostano thinks about all this.

:+1:

mkllnk:
In Aus we like to use the `openfoodnetwork` user to restart unicorn. In the past, it has been an init script which could be executed as user. If it was executed as root, it still executed it as the ofn user. I noticed that using a service to run that script requires root. I thought that https://github.com/openfoodfoundation/ofn-install/pull/143 would solve that.

Background: we deploy with the Git post-receive hook. It runs as the user and uses the init script at the moment to reload the unicorn (via signal). This could also be handy if we wanted to offer an update UI like Discourse does.

elf-pavlik:
As requested by @sauloperez in https://github.com/openfoodfoundation/ofn-install/pull/144#discussion_r177011579 this PR now keeps the sudoer file which allows `openfoodnetwork` to manage unicorn with
```shell
/bin/systemctl status unicorn_{{ app }}.service
```
same for `start`, `stop` and `restart`

It does work when done from shell, not by ansible.

Approach in #143 would work with unicorn but not with nginx. I would propose to merge this one here to get the whole playbook to successful run. IMO it doesn't lower security in any way. After that we could still consider using approach from #143 as possible improvement in security.

We could also discuss later if `deploy.yml` playbook has to restart nginx or maybe enough to restart it after `provision.yml`. Also if for that playbook we can set `remote_user: "{{ unicorn_user }}"` instead of `"{{ user }}"`. Still I see benefit of getting the whole playbook first to succeed on fresh VM (or container) without making things worst and then systematically improve things.

sauloperez:
:recycle: there are conflicts

elf-pavlik:
> If you could just remove those extra new lines would be nice, but nothing important.

done

> :recycle: there are conflicts

done

Issue #139, opened 2018-03-21
user account crated with admin_email should get auto confirmed


elf-pavlik:
Currently user account gets created but on first login it complains that accounts requires confirmation and offers to re-send confirmation email.

suggested in https://github.com/openfoodfoundation/ofn-install/pull/134#issuecomment-374691331

elf-pavlik:
Requires change to the app, tracked in openfoodfoundation/openfoodnetwork#2182

*(we can close it here once that issue gets resolved)*

elf-pavlik:
Done with: https://github.com/openfoodfoundation/openfoodnetwork/pull/2288

PR #146, opened 2018-03-25
auto confirm admin user after seeding database

elf-pavlik:
Fixes: #139

---
This change is [Reviewable](https://reviewable.io/reviews/openfoodfoundation/ofn-install/146)

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -195,6 +195,14 @@
   tags:
     - seed
     - skip_ansible_lint
+
+- name: confirm admin user
+  command: bash -lc "bundle exec rails runner 'Spree::User.first.confirm!'"
+  args:
+    chdir: "{{ current_path }}"
+  tags:
+    - seed
+    - skip_ansible_lint
   notify:
     - precompile assets
     - restart unicorn


elf-pavlik:
I don't know what this has to do with `dev:load_sample_data`?

In staging and production environment it also comes helpful to have the admin confirmed and not have to use email confirmation steps.

sauloperez:
We do all the sample data setup from that rake task, and we already autoconfirm the spree user there. See https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/tasks/dev.rake#L206

elf-pavlik:
Do you have clear suggestion of how to extract https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/tasks/dev.rake#L206 into something that both `dev.rake` and playbook can reuse?

Last years I work mostly with javascript, gulp etc. and don't have a that good of a grip on ruby and rake.

Would extracting confirmation into a separate task, which this playbook would run (instead of `rails runner` used in this PR)  and then also using it in `dev.rake` as suggested here: https://coderwall.com/p/no2w6g/invoke-a-rake-task-from-another-task make sense for you?

mkllnk:
I also think that this is not the right approach. Sample data is not right either, because that's not for production servers. We should probably look into the generation of the user in Spree (rake db:setup) and do it there.

elf-pavlik:
Maybe someone could make PR with an alternative approach? This one just automates what otherwise I have to do manually to log in to admin interface. Otherwise this one could land as imperfect but working and new Issue could track ideas for better approach until someone comes up with PR.
I really find it important that one can run this playbook, it runs successful and then open a browser and log in right away. Once there it can take a lot of improvements while always making sure to keep smooth deployment experience.

mkllnk:
Yes, I think you are right. It's not the best approach, but it is a solution and we should use it until we have a better one.

enricostano:
We're making an effort to decouple server provisioning and deploy from application logic, let's try to move all together in that direction if you agree.

I think that this should be in the `db:seed` data rake task, that is called every time you call `db:setup` rake task.

May we close this in favour of https://github.com/openfoodfoundation/openfoodnetwork/issues/2182 ?

sauloperez:
And executed on deploy:

https://github.com/openfoodfoundation/ofn-install/blob/92ca2af690412cd64a4f8d46f36ac8d42714ec12/roles/deploy/tasks/main.yml#L189-L200

enricostano:
@sauloperez 

> And executed on deploy

I think we should be able to remove it from there, could you please open an issue to investigate that?

elf-pavlik:
> May we close this in favour of openfoodfoundation/openfoodnetwork#2182

PR #147, opened 2018-03-27
Add URL to vagrant image

mkllnk:
The default catalogue moved and this image is not found on
atlas.hashicorp.com any more. Older versions of vagrant still try
to access the old server and need the URL explicitely.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   config.vm.provider :virtualbox do |vbox|
     config.vm.box = "ubuntu/trusty64"
+    # The default catalogue moved and this image is not found on
+    # atlas.hashicorp.com any more. Older versions of vagrant still try
+    # to access the old server and need the URL explicitely.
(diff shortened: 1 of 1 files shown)


sauloperez:
Do we still maintain this Vagrantfile? I didn't know we had one.

enricostano:
I never used it, but from my understanding was used to "test" a whole provisioning. To be clear, we have:
- this (old?) VM configuration to "test" the provisioning for all the environments
- the LXC configuration that is used only for OFN development purposes

mkllnk:
I think it's nice to test the whole thing on a virtual machine. I'm fine with a container as well, but that isn't implemented yet.

sauloperez:
It is @mkllnk https://github.com/openfoodfoundation/ofn-install/blob/bd9fbf9057de4754ab69911dcf050cb0d8c49b5d/scripts/create-container.sh

mkllnk:
Hm, that script needs sudo rights. :-(

enricostano:
I wrote down that distinction in my previous comment exactly to avoid this confusion... and I failed :(

Vagrantfile and LXC conf serve 2 different goals. Please don't mix them.

Let's merge this as it will be maintained from people that use it.

Issue #114, opened 2018-02-03
Fetch ssh keys from URL (e.g. github profile)


elf-pavlik:
http://docs.ansible.com/ansible/latest/authorized_key_module.html#examples
```yaml
- name: Set authorized key took from url
  authorized_key:
    user: charlie
    state: present
    key: https://github.com/charlie.keys
```

I would prefer to only include URL `https://github.com/elf-pavlik.keys` in this repo, this way when I change SSH key I only update it on github and don't have to make another commit in this repo.

enricostano:
I like the idea but I'm worried about what happens when we try to remove some user as a sys admin.

[Right now](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/default_user/tasks/main.yml#L13) we don't pass the `state` attribute for each user to the `authorized_key` module. That means that we don't have a way to remove users... So we can just add support for SSH key by URL pretty easily.

But we'll need to think about user removal while implementing this.

mkllnk:
Closing as this would just increase dependency on another service. And we would still need to add a URL for each user. The SSH key is not that much bigger.

PR #149, opened 2018-03-27
Vars for new Aus prod2

mkllnk:
This is the config used for our new production server in Australia. The group var file had to be renamed to follow the conventions.

I'm aware that I listed some admin users without providing their keys. I keep them locally. Call me paranoid, but I prefer not to publish my public key. I would also not publish my `/etc/shadow` file or load my encrypted backups to the public cloud. It just lowers the chance of someone brute-forcing it.


enricostano:
> Call me paranoid, but I prefer not to publish my public key

This one @mkllnk ? https://github.com/mkllnk.keys

:trollface:

mkllnk:
Lol! Yes! I didn't know Github was doing that. Well, I saw an interesting security talk lately that recommended having different keys for different projects. So if Github publishes my key, maybe I should use other keys for everything else...

Thanks Enrico for pointing that out.

elf-pavlik:
>  Well, I saw an interesting security talk lately that recommended having different keys for different projects.

Very likely 'damage control' in case on of the **private** keys gets compromised.

I can't recall seeing from any credible security source a recommendation not to publish **public** keys.
4096 bits RSA keypair or 256 bit Ed25519 (unless you believe that [this curve has a back door](https://www.youtube.com/watch?v=nybVFJVXbww)) should keep most of us save.
If I would try to tighten my security, I would focus on further restricting access to private keys, for example by storing it on hardware like - https://www.nitrokey.com/

I guess you will not like #114 :cat:

mkllnk:
Well, it's not that I'm scared that my public key got published. It's more that I prefer not to do it if there is no good reason for it. And since easy managing of admin users is a good reason, I'm actually leaning towards your approach. #114 is a good solution, much easier then keeping the keys in the repo up-to-date.

My concerns come more from cases like screwed random number generation. But if we have easy processes to replace keys, we can react quickly to those.

PR #150, opened 2018-03-27
Run certbot as root

mkllnk:
Running it as the openfoodnetwork user fails since it doesn't have sudo
permissions.

Fixes https://github.com/openfoodfoundation/ofn-install/issues/86

This would become obsolete with #124, but is a smaller change and could be merged before.


enricostano:
Closing since https://github.com/openfoodfoundation/ofn-install/pull/124 got merged.

PR #151, opened 2018-03-28
Gitignore vendor/

sauloperez:
## What?

`vendor/` seems to be Ansible's default location for plugins. As this wasn't ignored by git, it always shows up as untracked files when I run `git status` and it's quite annoying. I'm just ignoring it now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ community
 inventory/host_vars/*/secrets.yml
 *.log
 **/*.retry
+vendor/


danypr92:
I think that we use `community/` instead of `vendor/`, no?

https://github.com/openfoodfoundation/ofn-install/blob/fd0bfc967f16ab5906bf0459ed75395802b20e29/ansible.cfg#L3

And `community/` is already in `gitignore`: https://github.com/openfoodfoundation/ofn-install/blob/fd0bfc967f16ab5906bf0459ed75395802b20e29/.gitignore#L5

enricostano:
@danypr92 is right:

> This can be set to a series of directories (i.e. /etc/ansible/roles:~/.ansible/roles), in which case the first writable path will be used.

http://docs.ansible.com/ansible/latest/reference_appendices/galaxy.html#roles-path

elf-pavlik:
@sauloperez do you still see need for this PR?

PR #153, opened 2018-04-05
Specify SKYLIGHT_AUTHENTICATION in systemd unit

sauloperez:
Otherwise, Unicorn cannot see this env var and Skylight reports a "authentication token required".

### DIFF
--- a/roles/webserver/templates/unicorn.service.j2
+++ b/roles/webserver/templates/unicorn.service.j2
@@ -8,6 +8,11 @@ Type=simple
 User=openfoodnetwork
 WorkingDirectory={{ current_path }}
 Environment=RAILS_ENV={{ rails_env }}
+
+{% if skylight_authentication is defined %}
+  Environment=SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
+{% endif %}
+
 SyslogIdentifier=openfoodnetwork-unicorn
 PIDFile={{ unicorn_pid }}
 


sauloperez:
Now I can find the following in the `log/staging.log`

```
[SKYLIGHT] [1.6.0] Skylight agent enabled
```

and nothing shows up in the `log/skylight.log`.

roles/webserver/templates/unicorn.service.j2
```diff
@@ -8,6 +8,11 @@ Type=simple
 User=openfoodnetwork
 WorkingDirectory={{ current_path }}
 Environment=RAILS_ENV={{ rails_env }}
+
+{% if skylight_authentication is defined %}
+  Environment=SKYLIGHT_AUTHENTICATION={{ skylight_authentication }}
```
danypr92: We can check https://www.freedesktop.org/software/systemd/man/environment.d.html to set environment variables only in a one place

enricostano: Please open an issue to cover this later.

Matt-Yorkley:
@sauloperez What's the best way to handle having two different keys for production and staging? Can we do something like this?

`skylight_authentication_production = xxxxx`
`skylight_authentication_staging = yyyyy`

`{% if skylight_authentication_{{ rails_env }} is defined %}`
`    Environment=SKYLIGHT_AUTHENTICATION={{ ( skylight_authentication + '_' + rails_env ) }}`

Matt-Yorkley:
Ah, we just have a unique secrets file for each domain. That makes more sense...

PR #155, opened 2018-04-15
set values for variables expected to be defined in roles defaults

### LINKED ISSUE #93: Avoid the need for s3_backups_bucket: none
## What

we shouldn't need to define vars that we don't use and check for their existence wherever they are used. Otherwise, the more vars are used the more annoying it'll be for instances that don't use them.

We should do the same as in https://github.com/openfoodfoundation/ofn-install/pull/92 for `s3_backups_bucket`.

### LINKED ISSUE #109: 666 values in es & us group_vars
both files include

```yaml
s3_backups_bucket: none
s3_access_key: 666
s3_secret: 666

stripe_client_id: 666
stripe_instance_secret_key: 666
```

this seems like some kind of workaround. i just tried to provision a server using playbooks from this repo and my first attempt failed with: `AnsibleUndefinedVariable: 's3_access_key' is undefined`

i also wonder why those 'secrets' don't go to git ignored `secrets.yaml` in `host_vars`?

elf-pavlik:
fixes #109
fixes #93 

:imp: --

It relies on: http://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-default-variables

deploy role already has the one it needs defined this way: https://github.com/openfoodfoundation/ofn-install/blob/8fc7f3e2b21e6fbdbe3a3d24820416cf6331956f/roles/deploy/defaults/main.yml#L10

---
This change is [Reviewable](https://reviewable.io/reviews/openfoodfoundation/ofn-install/155)


enricostano:
Can't we just get rid of them? We should track them back to the OFN code and make them optional.

In that way we'll be able to completely remove them from `ofn-install` unless in some host they use them and they can add them in their host variables. And in any case these variables should go into the secret vaults... not with normal variables.

@elf-pavlik would you please mind to check if it's easy to make them optional in OFN? Thanks!

elf-pavlik:
Based on how #146 ended up, I would prefer make a small improvement first in this PR. You could right away create an issue with how you see further improvements which can reference this PR. Once someone has time and capacity to make that further improvement, it can happen in another PR.

elf-pavlik:
I must admit not understanding strategy of not accepting improvements already done and ready to merge, because someone could possibly improve it even further. This PR doesn't prevent or make further improvements any harder, while it does provide small but IMO obvious improvement. At the same time no one submitted alternative PR which would attempt to improve even more.

enricostano:
Hi, sorry for the delay. The only reason this PR hasn't been merged is that we didn't have time to test it on a OFN instance staging environment yet. I hope we can prioritize testing pending PRs ASAP.

sauloperez:
I assume @enricostano 's approval so moving to test ready.

elf-pavlik:
I see 2 approvals, could someone please merge it? :cat:

enricostano:
Can anyone please test this in a staging environment?

myriamboure:
Ok I'm happy to ask @pacodelaluna to stage on French staging, but I don't get what we are supposed to test... can you explain me @enricostano or @elf-pavlik ?

pacodelaluna:
I have deployed from this branch, didn't see any issue, looks good for me.

PR #163, opened 2018-06-02
Add missing Stripe keys in application.yml template

pacodelaluna:
Following the doc https://github.com/openfoodfoundation/openfoodnetwork/wiki/Setting-up-Stripe-on-an-OFN-instance, we were missing 2 variables on Ansible side to provide the specified way to enable Stripe into the OFN instance.

### DIFF
--- a/roles/app/defaults/main.yml
+++ b/roles/app/defaults/main.yml
@@ -5,3 +5,5 @@ s3_secret: none
 
 stripe_client_id: none
 stripe_instance_secret_key: none
+stripe_instance_publishable_key: none
+stripe_endpoint_secret: none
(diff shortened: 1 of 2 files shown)


roles/app/defaults/main.yml
```diff
@@ -1 +1,3 @@
 ---
+stripe_instance_publishable_key: none
+stripe_endpoint_secret: none
```
pacodelaluna: Following the style defined in #155, I may have to rebase before merging.

Matt-Yorkley:
I've just realised this solution has been replicated in #187 alongside other things...

elf-pavlik:
Yes #187 also included those keys, closing this one since the other one landed.

PR #165, opened 2018-06-10
Public key

Matt-Yorkley:
Adding my public key to the ofn-install repo. 

I've added a line in .gitignore as well to ignore files created by my IDE.


.gitignore
```diff
@@ -6,3 +6,4 @@ community
 inventory/host_vars/*/secrets.yml
 *.log
 **/*.retry
+.idea/
```
mkllnk: Hey Matt, do you have a global git ignore file as well? If you use that for all your IDE specific files, you don't need to change all projects you are working on. :-)

elf-pavlik: :+1:  https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Issue #112, opened 2018-02-03
install python using ansible raw module


elf-pavlik:
https://github.com/openfoodfoundation/ofn-install/wiki/Setup#setup-a-server-instance

> You will need to ensure Python2.7 is installed on your server, and that /usr/bin/python points to Python2.7.

Ansible can ensure that and this way leave less to worry about for person using it.

Probably we could just add snippet from this gist
https://gist.github.com/gwillem/4ba393dceb55e5ae276a87300f6b8e6f

I could make PR, which playbook file do you think I should add it to?

enricostano:
We use this [script](https://github.com/openfoodfoundation/ofn-install/blob/master/scripts/provisioning-dev.sh#L14) for now. I'm sure this can be better, I'm not sure an Ansible playbbok would be a good solution.

elf-pavlik:
Why do you think using ansible raw module as in [this gist](https://gist.github.com/gwillem/4ba393dceb55e5ae276a87300f6b8e6f) doesn't make sense?

enricostano:
I just think that adding an Python/YAML (Ansible) playbook to just do what one line in Bash would do maybe is too much. But maybe that's just me 😂

Matt-Yorkley:
I'll look at making a new task for this now. It holds up fresh provisioning currently, and Aus and others might be provisioning new 16.04 servers soon with ofn-install. Also had someone on Slack setting up a new staging server, so lets iron out these creases...

PR #166, opened 2018-06-11
Vagrant provisioning updates

Matt-Yorkley:
I've made a few adjustments to get Vagrant deploys working smoothly. I've just been doing this over the weekend as a learning exercise for working with Ansible and Vagrant. 

There were a lot of errors and some package issues with the previous version, but now it all runs smoothly in ~20 minutes with:

`$ vagrant up`
`$ ansible-playbook site.yml -i inventory/dev --limit=vagrant`

screenshot from 2018-06-11 17-32-50

I'd appreciate any feedback/comments. If we merge this we can also update the Vagrant wiki page in ofn-install, with: "Tested with Vagrant 2.1.1 and Virtualbox 5.2.10" and the updated command to provision it properly (it needs `--limit=vagrant` and doesn't need `-u vagrant`).


roles/unicorn_user/tasks/main.yml
```diff
@@ -16,6 +16,17 @@
...
   with_flattened: "{{ users_sysadmin }}"
   become: yes
+  when: ssh_key_path is undefined
+
+# Add ssh key path to run playbook command with --extra-vars
+# ex.: --extra-vars "ssh_key_path=.ssh/id_rsa.pub"
+- name: add ssh key alternative
+  authorized_key:
+    user: "{{ unicorn_user }}"
+    key: "{{ lookup('file', lookup('env','HOME') + ssh_key_path) }}"
```
sauloperez: It would be a bit more obvious for the user if `ssh_key_path` behaved as an actual path allowing `ssh_key_path=~/.ssh/id_rsa.pub` and also more flexible; it would allow public keys from any location.

So I suggest this to be

```sh
key: "{{ ssh_key_path }}"
```

Matt-Yorkley:
@sauloperez @enricostano

roles/unicorn_user/tasks/main.yml
Matt-Yorkley: I copied this from the other ssh key setup for the default user in https://github.com/openfoodfoundation/ofn-install/blob/530741c5d82f6c47b0e196666b0b13920310fedf/roles/default_user/tasks/main.yml#L20

Is it worth changing that one as well?

Matt-Yorkley: It looks like they have to match for it to run, so I guess yes is the answer...

Matt-Yorkley: On testing it, it looks like it won't accept paths that start with `~/`, which I guess is why the same method of using `{{ lookup('file', lookup('env','HOME') + ssh_key_path) }}` was used previously in /roles/default_user/tasks/main.yml.

sauloperez: Ah ok, nevermind then... Although I find it really weird...

Matt-Yorkley: The answer was somewhere in between: `{{ lookup('file', ssh_key_path) }}` can handle the tilde in the filepath, the previous addition of `lookup('env','HOME')` wasn't needed.

Vagrantfile
```diff
@@ -10,13 +16,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   config.vm.provider :virtualbox do |vbox|
-    config.vm.box = "ubuntu/trusty64"
-    # The default catalogue moved and this image is not found on
-    # atlas.hashicorp.com any more. Older versions of vagrant still try
-    # to access the old server and need the URL explicitely.
-    config.vm.box_url = "https://app.vagrantup.com/ubuntu/trusty64"
```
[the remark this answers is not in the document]
mkllnk: Okay, scrap that. My backported vagrant knows the new URLs.

mkllnk:
What do you think about splitting `inventory/dev` into `inventory/vagrant` and `inventory/lxc`? I don't think that anyone of us wants to run both at the same time.

mkllnk:
Yay, I just ran this and it worked beautifully. Thank you. I would just like to know about the Ubuntu image.

Matt-Yorkley:
Nice catch @mkllnk, after investigating a bit more I can see there's an unaddressed issues here which could interfere with provisioning new servers. It's related to this: https://github.com/openfoodfoundation/ofn-install/issues/112

I'll fix that issue first then update the vagrant file to use the official xenial64 image. :+1:

Matt-Yorkley:
Should we scrap `lxc`? I think it's outdated/unmaintained...

Matt-Yorkley:
I've fixed the xenial/python issue in #196. Once that's merged we can use this PR which is now set to use the official xenial64 image. I've tested the both together with vagrant using the official image and it's working fine.

mkllnk:
Oh, I just got this error:
```
fatal: [local_vagrant]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: /home/maikel~/.ssh/id_rsa.pub"}
```

That file lookup doesn't seem to work properly.

mkllnk:
Ah, no, forget that. I didn't have this merged. :-)

PR #187, opened 2018-07-09
Missing environment variables for Stripe and S3

### LINKED ISSUE #170: Stripe ENV vars
We currently are setting the `STRIPE_CLIENT_ID` and `STRIPE_INSTANCE_SECRET_KEY` environment variables in https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/application.yml.j2

But it looks like we also need `STRIPE_INSTANCE_PUBLISHABLE_KEY` and `STRIPE_ENDPOINT_SECRET` as well: https://github.com/openfoodfoundation/openfoodnetwork/blob/f2ad087be5374f1192243c87e8b45d833ac211cd/config/initializers/stripe.rb

On UK production it looks like they have been set manually afterwards.

Matt-Yorkley:
Closes #170.

Making sure we actually provision the env vars we need, and updating the example `secrets.yml` file to make it clearer what should be in there.

### DIFF
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -17,7 +17,7 @@ secret_token: 'xxxxxx'
 # Create a Google Maps API key and enable Javascript and Places libraries
 google_maps_api_key:
 
-# Amazon S3 backups with db2fog
+# Set these if using Amazon S3 backups with db2fog
 s3_backups_bucket:
 s3_access_key:
 s3_secret:
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -17,6 +17,20 @@ secret_token: 'xxxxxx'
 # Create a Google Maps API key and enable Javascript and Places libraries
 google_maps_api_key:
 
+# Amazon S3 backups with db2fog
+s3_backups_bucket:
+s3_access_key:
+s3_secret:
+
+# Stripe Connect API keys
+# Find these under 'API keys' and 'Connect' in your Stripe account dashboard -> Account Settings
+# Under 'Connect', the Redirect URI should be set to https://YOUR_SERVER_URL/stripe/callbacks (e.g. https://openfoodnetwork.org.uk/stripe/callbacks)
+# Under 'Webhooks', you should set up a Connect endpoint pointing to https://YOUR_SERVER_URL/stripe/webhooks e.g. (https://openfoodnetwork.org.uk/stripe/webhooks)
+# Different keys can be made for testing or live environments
+stripe_client_id: # ca_xxxx
+stripe_instance_secret_key: # sk_xxxx
+stripe_instance_publishable_key: # pk_xxxx
+stripe_endpoint_secret: # whsec_xxxx
 
 # Set this if using New Relic
 newrelic_license_key:
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -57,7 +57,7 @@
   tags:
     - app_templates
     - db2fog
-  when: s3_backups_bucket != "none"
+  when: s3_backups_bucket != ""
 
 - name: get l10n repo
   git:
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -75,7 +75,7 @@
     force=yes
     owner={{ unicorn_user }}
     mode=0775
-  when: s3_backups_bucket != "none"
+  when: s3_backups_bucket != ""
   tags: symlink
 
 - name: symlink the newrelic file into the config folder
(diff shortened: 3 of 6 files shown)


inventory/host_vars/_example.com/secrets.example.yml
```diff
@@ -17,6 +17,20 @@ secret_token: 'xxxxxx'
 # Create a Google Maps API key and enable Javascript and Places libraries
 google_maps_api_key:
 
+# Amazon S3 backups with db2fog
```
elf-pavlik: those get set by default to none
https://github.com/openfoodfoundation/ofn-install/blob/530741c5d82f6c47b0e196666b0b13920310fedf/roles/app/defaults/main.yml

i guess  a comment should clarify that deployments not using S3 should delete those line from secrets or at least comment them out?

Matt-Yorkley: Our check for including the S3 settings is: `when: s3_backups_bucket != "none"`. Do we need this defaults file or can we just check if `s3_backups_bucket` is set instead?

Matt-Yorkley: We have variables initialized in `app/main/defaults.yml` and `deploy/defaults/main.yml` with the string "none". Is there a reason it was done like that? We could just have them blank and instead check `when: s3_backups_bucket != ""`, and that way blank variables in secrets files won't have to be commented out. Is that a better solution?

Matt-Yorkley: I think it makes more sense if people don't have to manually comment things out. I've added a new commit for it.

sauloperez: I'd like to investigate why we used `none` in the first place and go one way or another (`none` or `empty`). I agree with your approach @Matt-Yorkley

elf-pavlik: I see `none` used already 2 years ago https://github.com/openfoodfoundation/ofn-install/blame/master/roles/app/tasks/main.yml#L60
I don't recall discussion about `none` or `empty`, I just worry that someone might get impression that shouldn't leave any values empty, but I think we could clarify that one can lave certain values empty or delete them all together (which what I would do in my configuration file)

Matt-Yorkley: I've updated the wording to: `# Set these if using Amazon S3 backups with db2fog` to clarify that they are optional.

Matt-Yorkley: I've tested both deployment and provisioning (with and without the keys set in secrets.yml) with this branch on UK staging, and these were no errors. So it looks like we don't need to initialize variables with random strings like "none" or "this_is_an_empty_variable". :+1:

roles/app/tasks/main.yml
```diff
@@ -57,7 +57,7 @@
   tags:
     - app_templates
     - db2fog
-  when: s3_backups_bucket != "none"
+  when: s3_backups_bucket != ""
```
enricostano: You should be able to use the `s3_backups_bucket is defined` syntax here.

roles/deploy/tasks/main.yml
```diff
@@ -75,7 +75,7 @@
     force=yes
     owner={{ unicorn_user }}
     mode=0775
-  when: s3_backups_bucket != "none"
+  when: s3_backups_bucket != ""
```
enricostano: You should be able to use the `s3_backups_bucket is not defined` syntax here.

roles/app/tasks/main.yml
Matt-Yorkley: That's a separate approach, which might be even cleaner. We could scrap the variables in defaults.yml entirely and check if the variable is defined when running this task. Ansible variables are empty strings if the variable is defined but blank.

enricostano: Yep, I would do that. Maybe in the example configuration we can put them commented out.

enricostano: I was talking about the placeholders in our example files... let's keep them, but commented out.

Matt-Yorkley: What do you think @enricostano ?

Matt-Yorkley: Sorry, I didn't refresh the page so I missed your new comments. There's an issue here in that we are also writing the s3 variables into application.yml in a way that is not conditional or separate from our other variables. So if we comment them out in defaults.yml and are optionally setting them in our secrets files, we have to reimplement how we write to application.yml to avoid this:
screenshot from 2018-07-11 11-40-45

enricostano: I see, feel free to use `!= ""` for now but please open an issue to tackle this later on, or at least to discuss it. We don't need empty env var for instance, etc.

Matt-Yorkley: Cool, I've captured the need to assess and refactor this at some point in issue #191. 

Can we merge this some time soon? UK prod backups have been offline fro months and I can't re-provision UK without breaking Stripe payments, which are in use. :scream:

Issue #179, opened 2018-07-05
Get rid of monit


myriamboure:
## Need

Now that server monitoring agent has been installed and is monitored by Wormly, we can get rid of monit

## Acceptance criteria

Monit has been removed and moniroting service still work as expected

Matt-Yorkley:
Note: before closing this issue, make sure to update the ofn-install Wiki.

https://github.com/openfoodfoundation/ofn-install/wiki/Server-monitoring
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Delayed-job

Matt-Yorkley:
Closed by #192

Issue #119, opened 2018-02-11
Delayed_job gets stuck and does not process jobs


sauloperez:
## What? Why?
A customer complained about the shop page being empty although the order cycle and the products where correctly populated. We then found out it was due to cache not being refreshed.

We then manually triggered a cache refresh executing `OpenFoodNetwork::ProductsCacheRefreshment.refresh(enterprise, order_cycle)` from the Rails console, but the job kept enqueued but not processed and the log had the following contents:

```
openfoodnetwork@scw-ecce22:~/apps/openfoodnetwork/current$ tail -f log/delayed_job.log                                 
2018-02-10T04:59:48+0000: [Worker(delayed_job host:scw-ecce22 pid:4035)] Error while reserving job: PG::Error: connection not open    
: UPDATE "delayed_jobs" SET locked_at = '2018-02-10 04:59:48.869453', locked_by = 'delayed_job host:scw-ecce22 pid:4035' WHERE id IN (SELECT
  id FROM "delayed_jobs"  WHERE ((run_at <= '2018-02-10 04:59:48.866440' AND (locked_at IS NULL OR locked_at < '2018-02-10 04:44:48.866533')
 OR locked_by = 'delayed_job host:scw-ecce22 pid:4035') AND failed_at IS NULL) ORDER BY priority ASC, run_at ASC LIMIT 1 FOR UPDATE) RETURNI
NG *                                                                                                                                  
2018-02-10T04:59:53+0000: [Worker(delayed_job host:scw-ecce22 pid:4035)] Error while reserving job: PG::Error: connection not open
: UPDATE "delayed_jobs" SET locked_at = '2018-02-10 04:59:53.876770', locked_by = 'delayed_job host:scw-ecce22 pid:4035' WHERE id IN (SELECT
  id FROM "delayed_jobs"  WHERE ((run_at <= '2018-02-10 04:59:53.873866' AND (locked_at IS NULL OR locked_at < '2018-02-10 04:44:53.873958')
 OR locked_by = 'delayed_job host:scw-ecce22 pid:4035') AND failed_at IS NULL) ORDER BY priority ASC, run_at ASC LIMIT 1 FOR UPDATE) RETURNI
NG *
2018-02-10T04:59:58+0000: [Worker(delayed_job host:scw-ecce22 pid:4035)] Error while reserving job: PG::Error: connection not open
[...]

PR #192, opened 2018-07-14
Delayed job systemd unit

### LINKED ISSUE #84: Unify how to start/stop services
I found myself stopping unicorn and delayed_job to release all DB connections in Katuma's staging server and I found a bit annyoing that not all services are managed in the same way.

I'd be good to have a single way to manage them. That means that we should manage `delayed_job` with systemd like all the others.

Matt-Yorkley:
Closes #84 and #179. And probably #119.

This PR adds a new systemd unit for managing our delayed_job processes. It also removes the monit package which is now no longer needed, as systemd is now configured to restart the service when it is not running.

It is usable as a regular service e.g: `sudo systemctl restart delayed_job_openfoodnetwork.service`. The tasks that start/reload delayed_job have been switched to use this service instead of the directly running the script. 

screenshot from 2018-07-13 21-42-22

I've also added a new task to run `systemctl daemon-reload` to reload the config after any systemd files have been updated, because they won't work otherwise.

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -1,0 +1,13 @@
+- name: stop monit service
+  service:
+    name: monit
+    state: stopped
+  become: yes
+  become_user: root
+
+- name: remove monit package
+  apt:
+    pkg: monit
+    state: absent
+  become: yes
+  become_user: root
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -33,13 +33,18 @@
     - libxslt1-dev
     - memcached
     - ranger
-    - monit
     #- fontconfig # needed for phantomjs
     #- nfs-common # make virtualbox faster
   tags: packages
 
-- name: configure monit
-  template:
-    src=monit.j2
-    dest=/etc/monit/conf.d/{{ app }}
+# Remove monit if still installed. The notified handlers will run only once
+# (when the main task triggers a change by removing the config file)
+- name: clean up monit
+  file:
+    state: absent
+    path: /etc/monit/conf.d/{{ app }}
   become: yes
+  become_user: root
+  notify:
+    - stop monit service
+    - remove monit package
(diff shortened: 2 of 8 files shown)


Matt-Yorkley:
Tested on UK staging with both provisioning and deployment, as it affects both. It cleanly removes monit, fixes the daemon-reload bug, restarts the delayed_job service with a new process on deployment, automatically restarts delayed_job after 60 seconds if the process is dead.

roles/common/handlers/main.yml
```diff
@@ -0,0 +1,13 @@
+- name: stop monit service
```
mkllnk: This is confusing. It's called "stop monit service", but stops delayed job.

roles/common/tasks/main.yml
```diff
@@ -33,13 +33,19 @@
     - libxslt1-dev
     - memcached
     - ranger
-    - monit
+    #- monit
```
sauloperez: why not simply :fire: ?

roles/common/handlers/main.yml
Matt-Yorkley: Nice catch, that was a mistake.

Matt-Yorkley:
@mkllnk The service can be restarted as `ofn-admin` with `sudo systemctl restart delayed_job_unicorn.service`. In theory you could still insteract with the script directly as `openfoodnetwork` with `bash -lc "./script/delayed_job restart --pid-dir=../shared/pids"` as we were doing before. Is that what you mean?

sauloperez:
Let's check if we have any documentation worth updating :warning:

Matt-Yorkley:
There's lots of places in the Wiki that will need updating after all the current open PRs are merged. I'll add a note in the sysadmin epic.

mkllnk:
@Matt-Yorkley Yes, that's what I mean. I thought we could interact via `/etc/init.d` in the past, but no, we only had `/etc/init.d/unicorn_openfoodnetwork` which could be run as `openfoodnetwork` user. So all good.

Matt-Yorkley:
I'm going to merge this delayed_job systemd PR now, if that's okay. The current implementation of monit leaving orphaned processes around that either block each other or eat up resources is still a regular pain in the ass for the UK.

Matt-Yorkley:
Warning: when you update your master branch, you'll probably need to re-provision before deploying otherwise ansible might try to restart the non-existent service.

PR #196, opened 2018-07-16
Ensure Python2.7 is available for ansible

### LINKED ISSUE #112: install python using ansible raw module

Matt-Yorkley:
Closes #112. 

Most of our servers are now using 16.04 Xenial, but our scripts don't actually run successfully on a fresh Xenial box :see_no_evil:. This is due to changes to the default installed python packages in the distro.

This PR adds a new playbook, role and tasks which are run by `site.yml` on a new install to ensure `python2.7` is installed and available to ansible before running the rest of the scripts.


Matt-Yorkley:
The role uses the `gather facts: no` setting which stops ansible using it's fancy modules (which would immediately cause a fatal erorr because python2.7 isn't there).

Matt-Yorkley:
P.S. I've tested this with a vagrant box using the official xenial64 image. It was definitely broken before, giving stdout/python errors, and it's now working perfectly with this fix.

roles/python/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+--- # This role ensures Python2.7 is installed and accessible for ansible
+- name: Install python 2.7
+  raw: apt-get update -qq && apt-get install -qq python2.7
```
enricostano: Why can't we use `apt` module?

playbooks/python.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Ensure python is set up and accessible on the server before starting (ansible requires python2.7)
+- name: python
+  hosts: ofn_servers
+  gather_facts: no
+  become: yes
+  roles:
+    - role: python
+      become: yes
+      tags: setup
```
enricostano: Is a separate playbook really needed?

roles/python/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  tags: skip_ansible_lint
+
+- name: check if softlink exists
+  stat:
+    path: /usr/bin/python
+  register: python_file
```
enricostano: There is no need to create the symlink, just declare the python interpreter variable. Link we do in https://github.com/coopdevs/timeoverflow-provisioning/blob/master/inventory/group_vars/all.yml#L2

roles/python/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- name: create softlink for python 2.7
+  raw: ln -s /usr/bin/python2.7 /usr/bin/python
+  when: python_file.stat.islnk is not defined
+  tags: skip_ansible_lint
```
enricostano: Same as above

roles/python/tasks/main.yml
Matt-Yorkley: Ansible's modules use python 2.7, and at this point it isn't available.

playbooks/python.yml
Matt-Yorkley: We could merge it with `/playbooks/default_user` into a single `/playboks/setup` that does both jobs. Would that be better?

Matt-Yorkley: After investigating and testing this, it looks like we have to do it this way. 

- We can't use the `apt` module or any other modules before we have run these tasks, or we hit the fatal error.
- We have to use the `gather_facts: no` option or we hit the fatal error. 
- We can only set `gather_facts: no` on a playbook, and all our other playbooks need this option to be on for modules to work, so we need a separate playbook and we can't mix them together.

screenshot from 2018-07-16 15-01-01

roles/python/tasks/main.yml
Matt-Yorkley: I think you're right about this one. I'll double check.

Matt-Yorkley: Yep, we already specify it in `/group_vars/all.yml` and it runs fine without the symlink. :fire:

roles/python/tasks/main.yml
elf-pavlik: https://docs.ansible.com/ansible/2.5/modules/raw_module.html
> Executes a low-down and dirty SSH command, not going through the module subsystem. This is useful and should only be done in two cases. The first case is installing python-simplejson on older (Python 2.4 and before) hosts that need it as a dependency to run modules, since nearly all core modules require it.

PR #201, opened 2018-07-18
Specify the Rails env when updating cron jobs

sauloperez:
Although we were passing RAILS_ENV when updating cron jobs through
`whenever --update-crontab`, it turns out that `production` is the
default value for the `:environment` argument.

I spotted this while investigating why we had 9 failed `db2fog:backup`
cron jobs reported in Bugsnag for production. That was impossible as we
didn't provision production in a very long time although we did in
staging.

The problem was that `crontab -l` in staging returned something like:

```
0 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=production bundle exec rake openfoodnetwork:cache:check_products_integrity --silent'
```

Thus, these jobs were executed as if they were in production.

### DIFF
--- a/roles/deploy/handlers/main.yml
+++ b/roles/deploy/handlers/main.yml
@@ -52,6 +52,6 @@
   become_user: root
 
 - name: update whenever
-  command: bash -lc "bundle exec whenever --update-crontab RAILS_ENV={{ rails_env }}"
+  command: bash -lc "bundle exec whenever --set 'environment={{ rails_env }}' --update-crontab"
   args:
     chdir: "{{ current_path }}"


sauloperez:
:warning: this PR might manual intervention once applied :warn

After testing this both manually from the server and through the deploy playbook I ended up with `crontab -l` having all jobs defined twice and delimited by the two following comments:

```
# Begin Whenever generated tasks for: RAILS_ENV=staging
...
# End Whenever generated tasks for: RAILS_ENV=staging

# Begin Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb  
...
# End Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb
```

After digging into whenever's source code I found out that the value after the colon is the identifier, which can be specified when running `whenever` commands.

To rely only on the state crontab ends with when running the deploy playbook, I first cleared all jobs with:

```
$ bundle exec whenever --set 'environment=staging' --clear-crontab /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb
$ bundle exec whenever --set 'environment=staging' --clear-crontab RAILS_ENV=staging
```

Then, after deploying I only had

```
# Begin Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb
0 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake openfoodnetwork:cache:check_products_integrity --silent'

5 0 * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/rails runner lib/open_food_network/integrity_checker.rb'

45 2 * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake db2fog:clean --silent'

0 0,4,8,12,16,20 * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake db2fog:backup --silent'

0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue HeartbeatJob 0'

0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue SubscriptionPlacementJob 0'

0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue SubscriptionConfirmJob 0'

0 1 * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake openfoodnetwork:billing:update_account_invoices --silent'

30 1 2 * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake openfoodnetwork:billing:finalize_account_invoices --silent'

# End Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb
```

Matt-Yorkley:
Do we have to require manual intervention from instance managers, or can we add something here to clear all the previous whenever tasks before updating it?

sauloperez:
If this was needed (I can't tell whether it was me messing with whenever), it would be a one-time fix that doesn't deserve an ansible task. If we did, we would spend time running a task that is not needed after the first time.

In any case, the manual intervention would be from a sysadmin.

Matt-Yorkley:
UK staging crontab just looks like this: 

screenshot from 2018-07-18 10-58-05

I see what you mean about the RAILS_ENV var.

Matt-Yorkley:
Ah I see, the last section in the command acts as the identifier, but it defaults to the file name being used. So when we switch to correctly passing the RAILS_ENV like this: `--set 'environment={{ rails_env }}'` and remove `RAILS_ENV={{ rails_env }}` from the end of the command, it defaults to displaying `# Begin Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb`

Ok, so a sysadmin will have to clean it up after this is provisioned the first time. I'm slightly uncomfortable about leaving random jobs for instance managers in the notes of some PR though.

sauloperez:
> I'm slightly uncomfortable about leaving random jobs for instance managers in the notes of some PR though.

This will require us to sync the provisioning of all production environments and fix it ourselves. It needs to be planned.

Matt-Yorkley:
The initial change to cron jobs only happened 7 days ago, and I think it's only UK that's been re-provisioned in that time? I think it's safe to merge this and I can sort out the UK servers as soon as it goes in.

PR #203, opened 2018-07-18
Add mail method settings

### LINKED ISSUE #202: UK prod: email settings get wiped on deploy
Mail methods settings for instances are currently being wiped with each deploy. The two settings are: `preferred_mails_from` and `preferred_mail_bcc`. We use them in the UK, but I guess nobody else is currently using them.

screenshot from 2018-07-18 15-32-32

I've made two PRs to cover this, one in main the repo, and one in ofn-install. I've used conditional assignment with defaults on both ends, so that if they are missing in either ofn-install or in the main repo due to things being merged or updated at different times there shouldn't be any errors.

Matt-Yorkley:
Closes #202. 

Enables mail settings for from_address and bcc to be set. These are currently able to be set in the instance config section, but get wiped after each deploy.

This PR has a corresponding PR in the main repo: https://github.com/openfoodfoundation/openfoodnetwork/pull/2482 

They should both be able to be merged separately without any problems.

### DIFF
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -43,3 +43,7 @@ mail_host:
 mail_port:
 smtp_username:
 smtp_password:
+
+# Optional mail settings
+# mails_from: from_address@example.com
+# mail_bcc: bcc_me@example.com
--- a/roles/unicorn_user/templates/defaults.j2
+++ b/roles/unicorn_user/templates/defaults.j2
@@ -9,5 +9,9 @@ MAIL_HOST={{ mail_host }}
 MAIL_PORT={{ mail_port }}
 SMTP_USERNAME={{ smtp_username }}
 SMTP_PASSWORD={{ smtp_password }}
-MAILS_FROM={{ mails_from | default("no-reply@{{ mail_domain }}") }}
-MAIL_BCC={{ mail_bcc | default('') }}
+{% if mails_from is defined %}
+  MAILS_FROM={{ mails_from }}
+{% endif %}
+{% if mail_bcc is defined %}
+  MAIL_BCC={{ mail_bcc }}
+{% endif %}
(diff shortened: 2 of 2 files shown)


inventory/host_vars/_example.com/secrets.example.yml
```diff
@@ -43,3 +43,5 @@ mail_host:
 mail_port:
 smtp_username:
 smtp_password:
+#mails_from:
```
sauloperez: I think my comment https://github.com/openfoodfoundation/openfoodnetwork/pull/2482#discussion_r203403257 would also apply here.

roles/unicorn_user/templates/defaults.j2
```diff
@@ -9,3 +9,5 @@ MAIL_HOST={{ mail_host }}
 MAIL_PORT={{ mail_port }}
 SMTP_USERNAME={{ smtp_username }}
 SMTP_PASSWORD={{ smtp_password }}
+MAILS_FROM={{ mails_from | default("no-reply@{{ mail_domain }}") }}
+MAIL_BCC={{ mail_bcc | default('') }}
```
sauloperez: IMO the responsibility of `ofn-install` here should only be to provide the appropriate values of these two vars to the app. Otherwise, we're duplicating logic between two repos, which will make it harder and error-prone to change.

Matt-Yorkley: The values are optional, but will override the from_address of all mails from the instance. So the default is to ignore this setting, but if we leave it as an empty string I think the from_address will be set to an empty string, which is not what we want.

Matt-Yorkley: I also wanted the values not to cause errors if they were missing, both in ofn-install and the main repo, because we can't know if the updates on either side will both be present at the same time.

sauloperez: woudln't we have a win-win if we simply do:

```
{% if mails_from is defined %}
  MAILS_FROM={{ mails_from }}
{% endif %}
{% if mails_bcc is defined %}
  MAIL_BCC={{ mail_bcc }}
{% endif %}
```

This way the app will work in either case: providing the right value if the env vars are set or falling back to the default if they are not, while not duplicating any logic between repos. Am I right?

Matt-Yorkley:
Note: tested this by provisioning UK staging with this branch and deploying latest master after openfoodfoundation/openfoodnetwork#2482 was merged in the main repo. It works! Thanks for the quick turnaround, we will have two happy sysadmins in the UK. :tada:

Issue #198, opened 2018-07-17
Fail to install nginx


mkllnk:
Installing a fresh Ubuntu 16.04 server, I got the following error:
```
        "Setting up nginx-common (1.10.3-0ubuntu0.16.04.2) ...", 
        "Setting up nginx-core (1.10.3-0ubuntu0.16.04.2) ...", 
        "Job for nginx.service failed because the control process exited with error code. See \"systemctl status nginx.service\" and \"journalctl -xe\" for details.", 
        "invoke-rc.d: initscript nginx, action \"start\" failed.", 
        "\u001b[0;1;31m*\u001b[0m nginx.service - A high performance web server and a reverse proxy server", 
        "   Loaded: loaded (/lib/systemd/system/nginx.service; enabled; vendor preset: enabled)", 
        "   Active: \u001b[0;1;31mfailed\u001b[0m (Result: exit-code) since Mon 2018-07-16 09:03:29 UTC; 11ms ago", 
        "  Process: 6715 ExecStart=/usr/sbin/nginx -g daemon on; master_process on; \u001b[0;1;31m(code=exited, status=1/FAILURE)\u001b[0m", 
        "  Process: 6714 ExecStartPre=/usr/sbin/nginx -t -q -g daemon on; master_process on; (code=exited, status=0/SUCCESS)", 
        "", 
        "Jul 16 09:03:27 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] bi...", 
        "Jul 16 09:03:28 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] bi...", 
        "Jul 16 09:03:28 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] bi...", 
        "Jul 16 09:03:28 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] bi...", 
        "Jul 16 09:03:28 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] bi...", 
        "Jul 16 09:03:29 staging.openfoodnetwork.org.au nginx[6715]: nginx: [emerg] st...", 
        "Jul 16 09:03:29 staging.openfoodnetwork.org.au systemd[1]: \u001b[0;1;39mnginx.service: Con...\u001b[0m", 
        "Jul 16 09:03:29 staging.openfoodnetwork.org.au systemd[1]: \u001b[0;1;31mFailed to start A ...\u001b[0m", 
[...]

PR #205, opened 2018-07-19
Add optional playbook to remove unwanted packages

mkllnk:
The packages in this commit are present in Rimu Hosting images.

This solves #198. The errors I saw were caused by nginx not being able to run, because apache was using port 80 already.

I thought about including this in the main provisioning process. The packages could be defined as variables. But the packages don't depend on the OFN instance we want to install, they depend on the target machine and if other software is supposed to run there.

Do you have better ideas how to handle this problem?


sauloperez:
agree with @Matt-Yorkley

PR #206, opened 2018-07-19
Enable copy-paste of PR name from Github

Matt-Yorkley:
This isn't really part of an issue or an epic, but we talked about how it would be nice to have.

This enables copy-pasting of a PR's name from a Github page so you don't have to type the repo and branch name out separately in two places, for example when staging things for testing.

**Copy:** 
screenshot from 2018-07-19 19-25-22

**Paste:**
screenshot from 2018-07-19 19-30-31

P.S. I'm not invoicing for this one, just wanted to add this to make staging PRs a bit nicer. :heart:

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -12,8 +12,11 @@ user: ofn-admin
 # to connect using SSH instead of just passwords.
 ssh_pubkey: "{{ lookup('file', '~/.ssh/{{ ssh_pubkey_name }}') }}"
 
-git_repo: https://github.com/openfoodfoundation/openfoodnetwork.git
-git_version: master
+# You can paste the repo/branch straight from a github PR here, like: "Contributor22:new_feature"
+github_key: openfoodfoundation:master
+
+git_repo: "https://github.com/{{ github_key.split(':')[0] }}/openfoodnetwork.git"
+git_version: "{{ github_key.split(':')[1] }}"
 
 l10n_git_version: HEAD
 


mkllnk:
I was wondering how you do this. You may also be interested in https://gist.github.com/mkllnk/182dcac856fc8eebe154b48148018300.

inventory/group_vars/all.yml
```diff
@@ -12,8 +12,11 @@ user: ofn-admin
 # to connect using SSH instead of just passwords.
 ssh_pubkey: "{{ lookup('file', '~/.ssh/{{ ssh_pubkey_name }}') }}"
 
-git_repo: https://github.com/openfoodfoundation/openfoodnetwork.git
-git_version: master
+# You can paste the repo/branch straight from a github PR here, like: "Contributor22:new_feature"
```
mkllnk: While this is really cool, we lose some flexibility here. If somebody named their repo differently or hosted it somewhere else (because Github is now Microsoft and evil), they can't do it. Can we make both possible?

I also think it would be better to stage the pull request merged into master to test the final result. Github provides that commit: `git fetch origin pull/2479/merge`

I think instead of `git_version: master`, you could define `git_version: pull/2479/merge`. That gives you a copy and paste solution already. I haven't tested that yet. I would actually like to define that on the command line so that we don't have to change the config to stage.

Matt-Yorkley: If someone had a pull request that wasn't  from a fork of openfoodnetwork or used a non-github repo (has that ever happened in the history of ofn?) we can completely ignore the github_key and change git_repo and git_version, which is what we would do anyway.

sauloperez: I think the `--extra-vars` argument would allow us doing both. So if you don't go with the `repo:branch` you can do `--extra-vars "git_repo=foo git_version=bar" and that would have precedenve over the ones in this file. You can check it out at [Passing Variables On The Command Line](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#passing-variables-on-the-command-line).

mkllnk: Ah, okay. That's cool. I still think it would be good to have a process to deploy the merge commit of a pull request, but this a good step in the right direction already.

Matt-Yorkley:
I'm pretty sure you can pass the hash of a commit in the `version` paramater of the git module, which is what our `git_version` variable is being passed to, so you could probably do `github_key=openfoodfoundation:df29d4d6117c1ff7b4833556fc4148884309a7c9`.

mkllnk:
What about `openfoodfoundation:pull/2479/merge`?

Matt-Yorkley:
It looks like ansible can't do that one...

PR #208, opened 2018-07-20
Speed up apt

mkllnk:
The official documentation recommends this approach of passing all packages to apt at once which is much more efficient than calling apt 20 times.

I also removed two packages from the list. They are not on my dev machine and I never needed them. I think they were for a previous admin.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -9,32 +9,29 @@
   #TODO remove unnecesary packages from this list.
 - name: install base packages
   apt:
-    pkg="{{ item }}"
-    state=present
+    name: "{{ packages }}"
   become: yes
-  with_items:
-    - acl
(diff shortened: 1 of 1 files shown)


roles/common/tasks/main.yml
```diff
@@ -2,39 +2,46 @@
...
+      - python3-psycopg2
+      - python-software-properties
+
+      # unknown why or if needed
+      # The playbooks run without these dependencies,
+      # but they may be hidden run time dependencies
+      # like imagemagick.
+      - acl
+      - nodejs
+      - curl
```
sauloperez: it might be just because it's a nice-to-have

Matt-Yorkley:
Ok, I think we can merge this one. It's going to be quiet in August but we should try to keep things moving...

PR #212, opened 2018-07-20
Allow unicorn user deal with delayed job

sauloperez:
Allow the openfoodnetwork user issuing systemd commands for the delayed
job service without password.


elf-pavlik:
Only meant to use during sysadmin's manual ssh session and not by any playbook automation?
I recall that this doesn't work with playbooks and they need to use ansible privilege escalation.

Matt-Yorkley:
I think the idea is to use it during ssh sessions, so the user with least privileges can be used as much as possible and the only use the admin user when it's necessary...

PR #219, opened 2018-07-23
Allow running default_user task multiple times

sauloperez:
This task was meant to be executed just in the first provisioning and hence used `root` as `remote_user`.

By removing that argument we can now also use it to add extra authorized keys to the `ofn-admin` user by simply passing `-u ofn-admin` as follows:

```
$ ansible-playbook playbooks/default_user.yml -u ofn-admin --limit=es-staging
```

To run a first provisioning we will need then do it like:

```
$ ansible-playbook playbooks/default_user.yml -u root --limit=es-staging
```

### DIFF
--- a/playbooks/default_user.yml
+++ b/playbooks/default_user.yml
@@ -3,7 +3,6 @@
 # default user (ubuntu) and wont let this be run as root. But digital ocean images don't.
 - name: default_user
   hosts: ofn_servers
-  remote_user: root
   roles:
     - role: default_user
       become: yes


sauloperez:
Without this I got the following error:

```
➜  ofn-install (master) ansible-playbook playbooks/default_user.yml -u ofn-admin --limit=es-staging -e "@../katuma_secrets/staging.yml" --ask-vault-pass               ✱
Vault password:

PLAY [default_user] *****************************************************************************************************************************************************

TASK [Gathering Facts] **************************************************************************************************************************************************
Monday 23 July 2018  09:40:58 +0200 (0:00:00.042)       0:00:00.042 ***********
fatal: [staging.katuma.org]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: root@51.15.216.10: Permission denied (publickey).\r\n", "u
nreachable": true}
        to retry, use: --limit @/home/pau/dev/ofn-install/playbooks/default_user.retry

PLAY RECAP **************************************************************************************************************************************************************
staging.katuma.org         : ok=0    changed=0    unreachable=1    failed=0

Monday 23 July 2018  09:40:59 +0200 (0:00:00.838)       0:00:00.881 ***********
===============================================================================
Gathering Facts -------------------------------------------------------------------------------------------------------------------------------------------------- 0.84s

Matt-Yorkley:
We'll have to update the wiki in that case. We could automate it a bit more by checking if the ofn-admin user exists and using it by default, or root if it isn't present.

Matt-Yorkley:
That way the playbook would run correctly regardless of whether its the first time via site.yml or a subsequent use to add a user.

sauloperez:
For me it's fine to provide the user as an argument. Being a sysadmin you should be aware of the users present in the system and their responsibilities. Besides, it's nothing we'll run everyday.

Matt-Yorkley:
The `default_user.yml` playbook also gets run from `site.yml` when setting up a new server. It does need to be `root` to add the default user...

sauloperez:
can we document running those two playbooks then? I don't think is a big deal.

Matt-Yorkley:
We actually don't ever need to run the tasks for adding the ssh keys using `remote_user: root` if we stop running the ssh key tasks inside the `default_user` role, which we probably shouldn't do anyway, as they are clearly different roles. Then we don't need to pass different options on the commandline depending on the situation.

I've made an alternative suggestion here, tell me what you think: #222

sauloperez:
I like it! what about you @mkllnk and @danypr92 ?

Matt-Yorkley:
I've just tweaked it a bit more, so that adding ssh keys can go through it's own playbook instead of being combined into the new `setup.yml`, so it's not necessary to filter the roles with tags when provisioning new keys.

sauloperez:
Shall we close this in favor of #222 then @Matt-Yorkley ? Could you please copy over that checklist to your PR?

PR #211, opened 2018-07-20
Remove NewRelic configuration

mkllnk:
Decided on Slack: https://openfoodnetwork.slack.com/archives/C0DNLAZC7/p1531802988000065

PR #221, opened 2018-07-24
Unify inventory to simplify vagrant commands

mkllnk:
To execute playbooks against a vagrant machine, we have to type:

  ansible-playbook site.yml -i inventory/dev -l vagrant

With this change, we just need:

  ansible-playbook site.yml -l vagrant

And that's the same as with all others. I think the chances that we want
to provision all OFN servers at the same time are quite low. We could
introduce another group for that if we needed it.

I also removed some duplicate variables.


mkllnk:
> Do we need local under [ofn-servers:children]?

Yes. All our playbooks have this line: `hosts: ofn_servers`. If `local` isn't in there, these playbooks wouldn't run for vagrant. Maybe we should set `hosts: all` in all our playbooks...

inventory/host_vars/local_vagrant/config.yml
```diff
@@ -11,7 +11,6 @@ db_password: 'vagrant123'
 admin_password: 'spree123'
 secret_token: '511a3d0fa1551b9fa75a1aef5b47684905c64807963fa2c190272878366365'
 google_maps_api_key: 'xxx'
-newrelic_license_key: 'xxx'
```
sauloperez: why this?

mkllnk: Because we merged https://github.com/openfoodfoundation/ofn-install/pull/211 and it's not used anymore.

mkllnk: I should have separated that into a different commit to explain it.

sauloperez: Oops, I read something else. I didn't notice it was newrelice 🙈

PR #214, opened 2018-07-20
Add backup sysadmins to katuma servers

sauloperez:
In case either Dani or Enrico have a "stuck at the dentist" situation
while I'm on holidays.

Jokes aside, we decided with @Matt-Yorkley to let him use our staging
servers from now on as a proof-of-concept for our goal of
granting access for all sysadmins to all servers. This will let us spot
the bottlenecks and areas we will need to work on.

PR #222, opened 2018-07-24
SSH keys

Matt-Yorkley:
This is a proposal for changing the way we run the tasks for adding ssh keys, related to #219.

- Moved ssh key tasks into their own role instead of running them inside the `default_user` role.
- Added ssh keys role to `provision.yml` and run ssh keys tasks with `remote_user: {{ user }}` instead of `root`.
- Refactored our `python.yml` and `default_user.yml` playbooks into a single `setup.yml` playbook that runs through both "plays".
- Updated travis checks and docs for playbooks.

### Documentation Updates

- [ ] Update https://github.com/openfoodfoundation/ofn-install/wiki/Provisioning#default-user.
- [ ] Explain in https://github.com/openfoodfoundation/ofn-install#playbooks how keys are added to `ofn-admin`.
- [ ] Document how to add a new sysadmin as suggested in https://github.com/openfoodfoundation/ofn-install/pull/214#issuecomment-406857937.


playbooks/add_ssh_keys.yml
```diff
@@ -0,0 +1,9 @@
+---
```
sauloperez: now it's even more straightforward to add new keys

playbooks/setup.yml
```diff
@@ -0,0 +1,21 @@
...
+  tasks:
+    - name: Install python 2.7
+      raw: apt-get update -qq && apt-get install -qq python2.7
+      tags: skip_ansible_lint
+
+# Add the default user as root. AWS ubuntu images generally have a
+# default user (ubuntu) and wont let this be run as root. But digital ocean images don't.
+- name: set up default_user
+  hosts: ofn_servers
+  remote_user: root
```
mkllnk: You are assuming `root` again. But not all images come with root access. Can we set root only if no remote user is defined?

playbooks/add_ssh_keys.yml
```diff
@@ -1,9 +0,0 @@
----
```
mkllnk: Wouldn't it be good if we could run this on its own?

sauloperez: You can by specifying the `ssh_keys` tag.

playbooks/setup.yml
Matt-Yorkley: I havn't changed this role, just moved it. 

I think in this case where we don't have our default user yet, if the image doesn't allow root user access or has an alternate root user name, we can pass a different value for `remote_user` in the command line, with `-e "remote_user=ubuntu"` for example.

So it can already be overridden in those cases. Is there a more satisfying way of doing this?

playbooks/add_ssh_keys.yml
Matt-Yorkley: I think it's good that we're now separating things that should only be run once on the initial setup from things that would reasonably be re-provisioned. 

In theory you can just run the whole playbook for `provision.yml` at any time and only the things that need changing will be changed, but yes you can run individual parts of it using `--tags ssh_keys` for example.

sauloperez: > In theory you can just run the whole playbook for provision.yml at any time and only the things that need changing will be changed

playbooks/setup.yml
mkllnk: @sauloperez Does this work for you? Your solution was to take the root user out in https://github.com/openfoodfoundation/ofn-install/pull/219/files.

PR #224, opened 2018-07-27
Only reload systemd when init scripts change

Matt-Yorkley:
Just a quick improvement to the way we refresh systemd configs. 

It should happen: 
- immediately after updating init scripts
- before we call any more services
- only if our init scripts have been changed.


Matt-Yorkley:
`ansible-lint` complained on the build, so I added the `skip_ansible_lint` tag to this task. It suggested this task could be a handler, as it's using a conditional. In this case we need it to run immediately after the previous task and before any others, so if we used a handler here we would need to flush all handlers at this point, which is definitely not optimal. Also we need to ensure this runs before any other handlers are triggered, and I'm not sure if we could ensure they run in the right order if this was a handler.

PR #228, opened 2018-08-02
FR language packs

Matt-Yorkley:
Setting the default OS language to English for the new French server, and making sure French is installed so it is available.

Also added some notes to the `example.yml` file for reference.


pacodelaluna:
Much better to have the system in english if we want global management! Voilà!

PR #229, opened 2018-08-03
Add option to create a swapfile

mkllnk:
This can be very useful to avoid running out of memory. It may be that a
server only needs more memory during deployment.

I have been setting this up manually in the past. It would be nice to automate.

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -18,6 +18,7 @@
 
     - role: oefenweb.swapfile
       become: yes
+      when: swapfile_size != false
       tags: swapfile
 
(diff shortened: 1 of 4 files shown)


playbooks/provision.yml
```diff
@@ -12,6 +12,10 @@
       become_user: root
       tags: security
 
+    - role: oefenweb.swapfile
+      become: yes
+      tags: swapfile
+
```
Matt-Yorkley: Can we use this role conditionally so that it doesn't install it's dependencies unless it's being used? So using `when: swapfile_size != false` or similar.

mkllnk: Great idea. I added that.

PR #234, opened 2018-08-11
US update

mkllnk:
We have access to the server already. It would be good to include that here. Once we get access to the secrets file, we can do another deployment and update the config if necessary.


daniellemoorhead:
@mkllnk do you need more than one review for this? If so can you ping Matt as he's the only other person onboard at the moment :)

PR #237, opened 2018-08-17
Increase app timeout for UK server

mkllnk:
Workaround for https://github.com/openfoodfoundation/openfoodnetwork/issues/2545.

This pull request does not adjust the timeout in nginx. I'm not sure if that is required. An active unicorn could send keepalive packets to signal that it's still there, but busy. I haven't tested that. This may be enough.


mkllnk:
Superseeded by https://github.com/openfoodfoundation/ofn-install/pull/238.

PR #238, opened 2018-08-17
Increase UK timeouts

Matt-Yorkley:
Increasing timeouts for UK servers due to issue https://github.com/openfoodfoundation/openfoodnetwork/issues/2545.

### DIFF
--- a/inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
@@ -4,3 +4,5 @@ domain: staging.openfoodnetwork.org.uk
 rails_env: staging
 
 admin_email: dev.ofnuk@gmail.com
+
+unicorn_timeout: 120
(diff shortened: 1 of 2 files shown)


sigmundpetersen:
Partly duplicate of #237 ? Or is it complimentary @Matt-Yorkley ?

Matt-Yorkley:
Ah yes, I just noticed that. @mkllnk I've added the timeout to UK staging here as well, given the servers both have the same specs and usually have similar data sets. What do you think?

mkllnk:
In that case, it could go into the group vars, but it doesn't really matter.

daniellemoorhead:
@luisramos0 can you have a look at this please, it would be good to get it merged ASAP so the UK team can get to the bottom of the problem.

daniellemoorhead:
Moved this to ready to go as I assume no testing is required @Matt-Yorkley? Move into testing column if it does, or it just needs to be merged by whoever has the power.

mkllnk:
@luisramos0 I think that the `unicorn.rb` you found is not used actually used. It's referenced in `Procfile` and that seems to be a Heroku configuration file. We don't support Heroku any more. We should remove that file as well. Are you happy to create a pull request?

luisramos0:
yes I will. I love deleting dead code!

daniellemoorhead:
@luisramos0 😆

luisramos0:
PR created: [#2636 on ofn](https://github.com/openfoodfoundation/openfoodnetwork/pull/2636).

Issue #169, opened 2018-06-14
USA mail domain is not defined


enricostano:
```
$ git grep mail_domain
inventory/host_vars/_example.com/config.yml:mail_domain: example.com
inventory/host_vars/alpha.katuma.org/config.yml:mail_domain: katuma.org
inventory/host_vars/openfoodnetwork.de/config.yml:mail_domain: openfoodnetwork.de
inventory/host_vars/staging.katuma.org/config.yml:mail_domain: katuma.org
inventory/host_vars/staging.openfoodfrance.fr/config.yml:mail_domain: openfoodfrance.org
inventory/host_vars/www.openfoodfrance.org/config.yml:mail_domain: openfoodfrance.org
roles/unicorn_user/templates/defaults.j2:MAIL_DOMAIN={{ mail_domain }}
```

We just need to create the `mail_domain` variable in the US host vars file as done in other hosts.

PR #240, opened 2018-08-27
Add required mail_domain to us host config

mkllnk:
Closes: #169

The mail domain is needed for provisioning to run. Otherwise we get:

```
TASK [unicorn_user : Write OFN environment variables defaults] ****************************************************************************************************************************************************
Monday 27 August 2018  12:17:18 +1000 (0:00:00.034)       0:03:17.723 ********* 
fatal: [openfoodnetwork.net]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'mail_domain' is undefined"}
```

### DIFF
--- a/inventory/host_vars/openfoodnetwork.net/config.yml
+++ b/inventory/host_vars/openfoodnetwork.net/config.yml
@@ -1,4 +1,20 @@
 ---
+
 domain: openfoodnetwork.net
 rails_env: production
+
 admin_email: ofn-usa@openfoodnetwork.net
+
+mail_domain: openfoodnetwork.net
+
+# Optional list of subdomains to register in letsencrypt certificate
+# Defaults to {{ domain }} if list is undefined
+#
+#certbot_domains:
+#  - example.com
+#  - www.example.com
+#  - info.example.com
+
+# Size in bytes. You can also use units like 1G, 512MiB or 1000KB. See: `man fallocate`
+# The default is `false`, not installing a swapfile.
+#swapfile_size: 1G


inventory/host_vars/openfoodnetwork.net/config.yml
```diff
@@ -1,4 +1,20 @@
 ---
+
 domain: openfoodnetwork.net
```
luisramos0: hey @mkllnk from a link you sent me some days ago I thought we should always have www in our domains. Any reason for not using in?

mkllnk: That wasn't my decision. I'm not sure who set it up initially, but I guess we just kept it the same as it was. Don't change a running system.

While I understand the points for having a non-apex domain, I never encountered a real issue with OFN servers because of this. We are not using aliases, we are not using Cloudflare or similar providers, we just have one server with a static IP address.

sauloperez: would you mind sharing it again :trollface: ?

luisramos0: @sauloperez do you mean the link? I think it was this one: https://www.yes-www.org/why-use-www/

daniellemoorhead:
Peeps, it's been 11 days since this hit the code review column and only @luisramos0 has done this. 

@sauloperez @Matt-Yorkley @enricostano can one of you please do the second review so that we can get this to done?

PR #247, opened 2018-10-03
Add public key of Kristina and append to UK sysadmin group

kristinalim:
I need to inspect some data in the UK production database.


kristinalim:
I'm not sure - `users_sysadmin` is for general server access, right, and not necessarily superuser access? For what I need to do now, I don't need superuser access.

sauloperez:
I think you're right.

PR #256, opened 2018-10-08
Install compatible Ansible version on Travis

mkllnk:
The current version of jdauphant.nginx is not compatible with Ansible
2.7.0. The error was:

> 
>   ERROR! 'always_run' is not a valid attribute for a Handler
> 
>   The error appears to have been in
>   '/home/travis/build/openfoodfoundation/ofn-install/community/jdauphant.nginx/handlers/main.yml':
>   line 17, column 3, but may
>   be elsewhere in the file depending on the exact syntax problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python: "2.7"
 install:
   # - pip install dopy
   # - pip install httplib2
-  - pip install ansible
+  - pip install 'ansible<2.7.0'
   - pip install ansible-lint
 
 script:


mkllnk:
I'm merging this to fix the current master build. It's not changing production code.

Matt-Yorkley:
It looks like in the line in question, the command `always_run: yes` is deprecated in favour of the new syntax `check_mode: no`, and has been since Ansible 2.2. 

Maybe we should look at updating some of our community modules? It looks like the versions in our repo havn't been bumped for a while and they are a bit stale.

sauloperez:
that's a great point. We don't want to hit the same problem as we did with gem versions...

PR #262, opened 2018-10-22
Add Hugo to french group & public key

HugsDaniel:
I need to inspect some behavior in the FR production server


sauloperez:
ping @pacodelaluna @myriamboure

myriamboure:
We need to fix that bug so it seems necessary to add Hugo so that he can investigate what happens. As discussed in the retro on Tuesday we said: 1- reproduce in local env, 2- reproduce in staging, 3- reproduce in prod.
I just tried to reproduce in Aus staging and it works, so I guess Hugo needs production data to understand what's going on. So I approve it !

sauloperez:
You'll need to reprovision production @pacodelaluna

PR #266, opened 2018-11-09
Install swap file on US server

mkllnk:
Asset compilation failed, probably because it ran out of memory.


sauloperez:
Did this work @mkllnk ? because I did set this for Katuma due to the same reason and it didn't seem to work.

mkllnk:
Yes, it worked. I can't be 100% that it was due to the swap file, because there are many other processes affecting the current memory consumption. So on busy server 2GB memory plus 1GB swap may still not be enough. I think we have 8GB memory on Aus production.

sauloperez:
I actually added 4gb of RAM but then in the machine it showed 0kb of swap in use. It could as well be that the server didn't reach 100% mem usage.

PR #286, opened 2018-11-22
Serve multiple subdomains

sauloperez:
## Description

This change gives support for serving multiple subdomains from the same nginx server using HTTPS. 

This solves the problem of having to migration from alpha.katuma.org to app.katuma.org. This has been long overdue and we need it to improve the perception potential customers have about the project. It's been already a year passed the alpha!

This will be useful for other instances facing a similar problems due to increased flexibility.

## Changes

This introduces the `server_name` var as a way to provide a nginx server name other than the domain plus the `certbot_cert_name` var to allow also setting a certificate name other than the domain.

Note this depends on https://github.com/coopdevs/certbot_nginx/pull/7.

I carefully ordered the commits so you can follow easily. Check them if there's something you don't understand.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -118,7 +118,7 @@ nginx_sites:
     - |
       listen 80;
       listen [::]:80;
-      server_name {{ domain }};
+      server_name {{ server_name | default(domain) }};
 
       add_header X-Content-Type-Options nosniff always;
       add_header X-Xss-Protection "1; mode=block" always;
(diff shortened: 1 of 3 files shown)


inventory/group_vars/all.yml
```diff
@@ -118,30 +118,29 @@ nginx_sites:
     - |
       listen 80;
       listen [::]:80;
-      server_name {{ domain }};
+      server_name {{ server_name | default(domain) }};
```
Matt-Yorkley: :heart:

Issue #273, opened 2018-11-09
Investigate a quick solution for system monitoring


sauloperez:
We can't delay this any further. We need a way to know how OFN's instances are behaving in terms of Memory, CPU Load, Network and disk usage. Without these, we can't troubleshoot any performance problems or perform any [operations](https://stackify.com/web-operations/) task. This should allow us to start dealing with the latter which will eventually enable us to work on the former.

There are two options here: a hosted or a saas solution. While the first one gives us full control it also brings an extra maintenance and configuration overhead alhtough the software might be free.

On the other hand, a Saas solution will have a monthly cost and might not give us full control but it will have very little maintenance overhead compared to a hosted solution.

sauloperez:
I did a bit of investigation and the options seem to be

## Hosted
[Prometheus](https://prometheus.io/) with https://github.com/prometheus/node_exporter 

## Saas
[Datadog](https://www.datadoghq.com/) or [New Relic Infrastructure](https://newrelic.com/products/infrastructure) seem to be the biggest ones in this space.

Matt-Yorkley:
I would vote for a simple saas solution for all instances. Maybe Datadog.

sauloperez:
I'd go straight with datadog. There's a free option we could start using. Check it out in https://www.datadoghq.com/pricing/

deanmarc25:
I can confirm the datadog installation script will install all of the dependencies for you; assuming we are linux based. The latest datadog agent is golang based, I think (version 6). I believe this is the script here: https://github.com/DataDog/dd-agent/blob/master/packaging/datadog-agent/source/install_agent.sh. You simply pass it your datadog api key and it does the rest in most cases.

I'd be happy to answer any questions on datadog. I used newrelic very briefly in the past but it became expensive and we discontinued it in favor of datadog and more recently, prometheus. It does have more integrations that when setup properly, you can really hone in on specific areas of the application cause problems.

I'm wondering if our infrastructure provider gives us the info we are looking for?

mkllnk:
We are using Wormly: https://www.wormly.com/help/metrics-collectd/metrics-overview

The good thing about them is that they use collectd, free software for monitoring and you can monitor anything using their API. If you have an account with them already, you get the basic metrics for free. A basic plan is $25. That is for monitoring (like Happy Apps) and metrics. It's pricey, but an excellent tool.


They are a nice small business in Melbourne.

sauloperez:
[...]

PR #288, opened 2018-11-22
Install datadog agent in provisioning

sauloperez:
## Description

This uses https://github.com/DataDog/ansible-datadog to install the agent with the provided `datadog_key`. Note this variable needs to specified in the secrets file for the agent to work.

Note you will need to execute `ansible-galaxy -r bin/requirements.yml install` before provisioning to fetch this new role.

## Production instances provisioned

- [x] Katuma
- [ ] UK
- [ ] France
- [ ] US

## Missing

- [x] Entry in example

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -69,3 +69,10 @@
 
     - role: webserver # Build the unicorn webserver.
       tags: webserver
+
+    - role: Datadog.datadog
+      become: yes
+      vars:
+        datadog_api_key: "{{ datadog_key }}"
+      when: datadog_key is defined
+      tags: datadog
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -13,3 +13,6 @@
 
 - src: oefenweb.swapfile
   version: v2.0.7
+
+- src: Datadog.datadog
+  version: 2.4.0
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -47,3 +47,9 @@ smtp_password:
 # mail_secure_connection: 'TLS'
 # mails_from: from_address@example.com
 # mail_bcc: bcc_me@example.com
+
+# Datadog monitoring. See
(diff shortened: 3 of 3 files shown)


sauloperez:
So fast @sigmundpetersen :joy: !

playbooks/provision.yml
```diff
@@ -69,3 +69,10 @@
 
     - role: webserver # Build the unicorn webserver.
       tags: webserver
+
+    - role: Datadog.datadog
+      become: yes
+      vars:
+        datadog_api_key: "{{ datadog_key }}"
```
enricostano: In the PR descriptio you talk about `datadog_api_key` variable, which one is correct?

In any case would be nice to add it to https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/host_vars/_example.com/secrets.example.yml

sauloperez: Update the description, sorry

enricostano: Could you please update the secrets example? thanks!

sauloperez:
@Matt-Yorkley @mkllnk @pacodelaluna we'll need you to add the api key to your corresponding secret files :warning:

playbooks/provision.yml
sauloperez: Done!

sauloperez:
IMO we should be together on this; that's what makes us strong and what we agreed on multiple times as a community. I don't care about the tool we use but I do care about the limited resource of OFN. Keep in mind that my proposal is to use a single Datadog account we all share, which I think it'll pay off compared to using different tools, with the frictions that brings when trying to share knowledge and efforts.

Also, I'm afraid that if you stick to a different solution for Aus, we perpetuate the single-point of failure of your instance.

mkllnk:
I like your thinking @sauloperez. I'm not planning on running the Australian instance as rogue outlaw forever. ;-)

I explained my thoughts a bit more in the other thread on the original issue: https://github.com/openfoodfoundation/ofn-install/issues/273#issuecomment-441854842

PR #296, opened 2018-12-03
Enable 1G swap file on Katuma

sauloperez:
Until we don't set up a new 8Gb-RAM new server, which is going to happen this month, this will let us sleep.

screenshot from 2018-12-03 13-25-10

:point_up: This is its usage under regular conditions


sauloperez:
Sí noi, sí. Estem ben cardats.

PR #299, opened 2018-12-14
Enable deployment with Semaphore

Matt-Yorkley:
I've set up UK staging to enable deployment via Semaphore. Here's the playbook I used to do it. 

It sets up Ansible on the remote server and adds an unprivileged `ofn-deploy` user that can run `ansible-playbook` with sudo. A copy of ofn-install is added to /home/ofn-deploy/ofn-install, which the deployment user cannot edit. 

The setup also creates an SSH key pair and adds the public key to the new ofn-deploy user's known keys. The private key needs to be copy-pasted into Semaphore when adding a new server, you can fetch it with: 
`ssh ofn-admin@example.instance.org "sudo cat keys/semaphore"`.

The commands added to the Semaphore build options look like this: 

`echo "staging.openfoodnetwork.org.uk ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBNRC38ZIlwmK3Zx3hn7lhbUaCj5uuxAx2FOQuiNVSxG6G/ppOeWVEyFn18McIGuxDuwkb5O3nlvotaoWb4GsPN4=" >> "$HOME/.ssh/known_hosts"`

`ssh ofn-deploy@staging.openfoodnetwork.org.uk -o LogLevel=ERROR "$BRANCH_NAME"`

The first one adds the signature to the known hosts. You can get the fingerprint using `ssh-keygen -H -F example.instance.org` if you've connected to the server before. The second command uses Semaphore's custom environment variable for the branch being deployed.

Ansible then runs the full `deploy.yml` playbook from within the server itself and shows the output in Semaphore: 

screenshot from 2018-12-14 12-16-49

The secrets file is currently pulled from /home/ofn-admin/secrets.yml. If we later move to a private repo for secrets storage, we can probably just pull it from there using another SSH key pair? Possibly encrpyting/decrypting it with the same key pair?


Matt-Yorkley:
I've added two tags to the tasks for fetching ofn-install to the server and running the community module setup scripts, so they can be run on their own with `playbook/remote_ansible.yml --tags fetch_ofn_install` for example. This could even be added to Semaphore as another command, so it always checks/fetches the latest ofn-install on the remote machine before doing the deployment.

Matt-Yorkley:
I did look at the `command` option for ssh. It's good for restricting to a single command that doesn't change it's options, but not so good if you want to pass different options like branch name for example.

What do you think can be done with the `ansible-playbook` command? Any examples?

Matt-Yorkley:
@mkllnk In the current Aus staging setup, Semaphore has access to an unrestricted shell for the `openfoodnetwork` user, right? 

I've just had a look on Aus staging, and the permissions aren't set restrictively on your secrets file, so the `openfoodnetwork` user can currently read your `secrets.yml` using `cat`...

mkllnk:
> It's good for restricting to a single command that doesn't change it's options, but not so good if you want to pass different options each time (like branch name for example.)

I agree that it's not as convenient, but you can use STDIN.
```sh
echo "$branchname" | ssh ofn-deploy@ofn.org
```

> What do you think can be done with the ansible-playbook command that worries you?

Everything. It can execute an attacker's playbook with root permissions. I know that it may be difficult to get the attacker's code onto the server. But with the extra variables evaluation an attacker can just load their own repository from Github with malicious code and then execute it as root.

The lshell project also doesn't seem to have a huge audience for a security critical component.

> In the current Aus staging setup, Semaphore has access to an unrestricted shell for the openfoodnetwork user, right?

That is correct and it's not ideal. That's why I said that I would like to use a separate deploy user. I just haven't had the time to work on it. :-)

Matt-Yorkley:
> I agree that it's not as convenient, but you can use STDIN.
> 
> ```shell
> echo "$branchname" | ssh ofn-deploy@ofn.org
> ```

I think the `command` option ignores all input sent over ssh by default, but we can parse `$SSH_ORIGINAL_COMMAND` in a separate script :+1: 

> with the extra variables evaluation an attacker can just load their own repository from Github with malicious code and then execute it as root.

That's a good point. And we don't even need to specify the repo actually. I guess this is equally true though if we let Semaphore build and push the branch that's being deployed.

Matt-Yorkley:
I've removed `lshell` and moved to using the `command` and `restrict` options in authorized
_keys.

roles/deploy/tasks/main.yml
```diff
@@ -22,12 +22,13 @@
...
+- name: build codebase from git branch
   git:
-    repo={{ git_repo }}
-    dest={{ build_path }}
-    version="{{ git_version }}"
-    force=yes
+    repo: "{{ git_repo }}"
+    dest: "{{ build_path }}"
+    version: "{{ git_version }}"
+    refspec: "+refs/pull/*/head:refs/remotes/origin/pull-request-*"
```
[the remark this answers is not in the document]
sauloperez: Just for me to understand, where does that merge commit come from? AFAIK the branch we're deploying from is not merged yet.

Matt-Yorkley:
@sauloperez Yes, I added it as a separate playbook for now, because we might not want to add this to every server, maybe just a few staging servers selectively. This way it's optional. We could add it to the provision playbook later if we wanted it everywhere.

I can document it a bit more, but the setup should be possible in about 5 minutes or so. After this PR is merged the steps would be: 

1. run the `setup_semaphore_deployment` playbook against the target server
2. `$ ssh-keygen -H -F example.instance.org` to get the server's fingerprint
3. `$ ssh ofn-admin@exmple.instance.org "sudo cat keys/semaphore"` to get the private key
4. add a new server through the wizard in Semaphore and paste in the details from steps 2 and 3 (the example commands are in this PR's description)

And yes, we can trigger deploys from machines that don't have ofn-install set up. We could also create a distinct `deployers` group var like our current `users_sysadmin` list and/or add individual users' ssh keys to the ofn-deploy user on each server, giving us granular control on who can deploy for each instance.

Matt-Yorkley:
With `restrict` and `command` set on the user that's connecting, passing anything with stdin fails with this message: `Pseudo-terminal will not be allocated because stdin is not a terminal.` shortly followed by the deploy script exiting with this message: `Invalid branch name given.` because `$SSH_ORIGINAL_COMMAND` is blank.

mkllnk:
This is really exciting. I like the decision to keep it separate for now. We probably don't want to run it against production servers just yet and we may make some adjustments.

@Matt-Yorkley About ssh and stdin: it works, but you need to read stdin instead of using the parameters. This is my test setup:

```
# .ssh/authorized_keys
restrict,command="/home/openfoodnetwork/test.sh" ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQ...

# test.sh
#!/bin/sh

echo "This is test.sh"

while read line; do
  echo "$line"
done


# command shell on my computer
$ echo "hello" | ssh ofn-prod
Pseudo-terminal will not be allocated because stdin is not a terminal.
This is test.sh
hello
```

No need to change your script though. Your solution is great.

roles/deploy/tasks/main.yml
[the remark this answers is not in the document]
mkllnk: Github creates that merge commit to test if the pull request has any conflicts or can be merged automatically.

sauloperez:
Yes, @Matt-Yorkley I thought about the same regarding a `deployers` group. That is something we will need as some people don't want to take the responsibility of deploying to prod envs.

Let's get this merged! can you fix CI first?

Matt-Yorkley:
@sauloperez I think the build error is not related to this PR. I made an issue here: #300

roles/deploy/tasks/main.yml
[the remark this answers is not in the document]
Matt-Yorkley: > Can we still deploy the master branch which is not represented in a pull request?

Yes, regular branches and release tags should work as normal. The custom refspec here just adds an additional option to the existing ones.

sauloperez:
Yes, that it's a nice addition. Will this change the way we deploy? Let's say I want to deploy v1.24.0. It will still pick up the commit the tag is pointing to, right?

Matt-Yorkley:
Yes. "Normal" refs like `master` or `v1.24.0` will be unchanged and work in exactly the same way as before. This refspec only applies when we reference a pull request in the format: `pull-request{number}`, in which case it will fetch the merge commit of the pull request.

sauloperez:
ok, so is there something stopping us from merging? It'd be great to be able to stage to any server from Semphore.

roles/semaphore_deployment/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    mode: 0700
+
+- name: check certificate already exists
+  stat:
+    path: "/home/{{ user }}/keys/semaphore.pub"
+  register: semaphore_key
+
+- name: generate semaphore key pair
+  command: "ssh-keygen -f /home/{{ user }}/keys/semaphore -t rsa -b 4096 -C semaphore -N '' "
+  when: not semaphore_key.stat.exists
```
mkllnk: You can replace the previous command with this simple condition:
```suggestion
  args:
    creates: "/home/{{ user }}/keys/semaphore.pub"
```

roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -0,0 +1,17 @@
...
+branch=$@
+allowed=^[a-zA-Z0-9._/\-]+$
+
+# Given branch name must be a single string with no spaces,
+# containing only: ".", "-", "/", "_", or alphanumeric characters.
+
+if [[ ! $branch =~ $allowed ]]
+then
+  echo "Invalid branch name given."
+  exit;
```
mkllnk:
```suggestion
  exit 1
```
That indicates that the script failed. You don't need `;` if you have a newline at the end.

roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -0,0 +1,17 @@
...
+
+branch=$@
+allowed=^[a-zA-Z0-9._/\-]+$
+
+# Given branch name must be a single string with no spaces,
+# containing only: ".", "-", "/", "_", or alphanumeric characters.
+
+if [[ ! $branch =~ $allowed ]]
+then
+  echo "Invalid branch name given."
```
mkllnk:
```suggestions
   echo "Invalid branch name given." >&2
```
Prints to STDERR instead of STDOUT.

roles/semaphore_deployment/templates/sudoers.j2
```diff
@@ -0,0 +1,2 @@
+# Grant {{ deployment_user }} sudo access for for deploy script
```
mkllnk:
```suggestion
# Grant {{ deployment_user }} sudo access for deploy script
```

sauloperez:
Please, rebase onto `master` to get the fixed build.

roles/semaphore_deployment/tasks/main.yml
Matt-Yorkley: Thanks, I hadn't seen that trick before. I can scrap the preceding task as well :heart:

Matt-Yorkley:
Okay, I've rebased on master and added those changes @mkllnk

roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+branch=$@
```
mkllnk:
```suggestion
branch="$1"
```

roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -0,0 +1,17 @@
...
+# Given branch name must be a single string with no spaces,
+# containing only: ".", "-", "/", "_", or alphanumeric characters.
+
+if [[ ! $branch =~ $allowed ]]
+then
+  echo "Invalid branch name given." >&2
+  exit 1
+fi
+
+cd /home/{{ deployment_user }}/ofn-install
```
mkllnk:
```suggestion
cd "/home/{{ deployment_user }}/ofn-install"
```

Just in case someone chooses a weird username. It's unlikely, but good practice to use quotes.

roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
mkllnk:
```suggestion

set -e
```

This makes sure that the script fails if any of the commands fail. In this case it just makes sure that the script fails if ansible-playbook fails.

Matt-Yorkley:
Okay, I've added those last few changes. Maybe we can merge and test the integration on Katuma staging?

sauloperez:
feel free to use it @Matt-Yorkley .

sigmundpetersen:
🤣 Amazing GIF-work @sauloperez! Saved my day

PR #315, opened 2018-12-28
Don’t compare to boolean literal

sauloperez:
That is what ansible-lint complains about. It fixes the build error

```
$ ansible-lint site.yml --exclude=community
[601] Don't compare to literal True/False
/home/travis/build/openfoodfoundation/ofn-install/playbooks/provision.yml:21
      when: swapfile_size != false
The command "ansible-lint site.yml --exclude=community" exited with 2.
```
There is no need to conditionally execute the swapfile role. It will skip all tasks if no `swapfile_size` is defined.

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -18,7 +18,6 @@
 
     - role: oefenweb.swapfile
       become: yes
-      when: swapfile_size != false
       tags: swapfile
 
     - role: unicorn_user # Create unprivileged user to run unicorn


playbooks/provision.yml
```diff
@@ -18,7 +18,6 @@
 
     - role: oefenweb.swapfile
       become: yes
-      when: swapfile_size != false
```
sauloperez: If I type `when: swapfile_size` as ansible-lint's docs suggest, I get

```
fatal: [staging.katuma.org]: FAILED! => {"msg": "The conditional check 'swapfile_size' failed. The error was: template error while templating string: expecte
d token 'end of statement block', got 'G'. String: {% if 2G %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/home/pau/dev/ofn-inst
all/community/oefenweb.swapfile/tasks/main.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending
line appears to be:\n\n---\n- name: install dependencies\n  ^ here\n"}
```

sauloperez:
Would either you @mkllnk or @Matt-Yorkley test this in another staging. It worked in Katuma's staging but better to be sure.

Matt-Yorkley:
I've tested this now and it's working fine.

One thing I discovered whilst testing is that after the swapfile has been created, any subsequent changes are ignored, including if the `swapfile_size` variable is changed (to a higher amount, for instance) or removed. That's part of the module itself though, not this branch.

Good to go.  :+1:

sauloperez:
Sounds like we might need to open a PR to that plugin then. Shall we create an issue in their repo?

sauloperez:
Thanks for opening the issue @mkllnk!

Issue #74, opened 2017-06-29
Use same Bundler version as in OFN `Gemfile.lock`


enricostano:
This is a follow up to the discussion started in https://github.com/openfoodfoundation/openfoodnetwork/pull/1641

Here is where we install Bundler: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/bundler/tasks/main.yml#L6

mkllnk:
This just became really important as the newest bundler doesn't support our Ruby version any more. Ofn-install is broken for new hosts.

PR #320, opened 2019-01-09
Install used bundler version on deploy

mkllnk:
Closes https://github.com/openfoodfoundation/ofn-install/issues/74.
https://openfoodnetwork.slack.com/archives/C0DNLAZC7/p1546618893018000

Installing during deploy instead of provision enables us to read the
Gemfile.lock file which contains the last used bundler version.

The newest bundler v2 doesn't support our Ruby 2.1.5. We need to install
an older bundler.

### Dependencies

This depends on https://github.com/openfoodfoundation/openfoodnetwork/pull/3302 for the install script.

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -78,6 +78,15 @@
   when: s3_backups_bucket != ""
   tags: symlink
 
+- name: install bundler
+  # This needs to be run inside a bash shell to initialise rbenv
+  # See http://stackoverflow.com/questions/22115936/install-bundler-gem-using-ansible
+  command: bash -lc "./script/install-bundler --no-ri --no-rdoc"
+  args:
+    chdir: "{{ build_path }}"
+  register: bundler
+  changed_when: "bundler.stdout != ''"
+
 - name: bundle install app dependencies
   #TODO make the "--without development" part conditional on rails_env
   # Note: the 'LANG=...' is a fix for broken rubygems utf8 handling.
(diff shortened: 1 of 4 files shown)


roles/deploy/tasks/main.yml
```diff
@@ -78,6 +78,13 @@
   when: s3_backups_bucket != ""
   tags: symlink
 
+- name: install bundler
+  # This needs to be run inside a bash shell to initialise rbenv
+  # See http://stackoverflow.com/questions/22115936/install-bundler-gem-using-ansible
+  command: bash -lc "RBENV_VERSION='{{ ruby_version }}' gem install bundler --no-ri --no-rdoc -v \"$(grep -m 1 -A 1 -x -F 'BUNDLED WITH' '{{ build_path }}/Gemfile.lock' | tail -n 1  | tr -d '[:space:]')\""
```
sauloperez: Is it possible to store the parsing of the version in a var? And perhaps that's better suited in a role. It'd increase readability a great deal.

mkllnk: I thought about that. This was just a quick fix to help Shawn, but you are right that we should solve it properly.

I revived a bundler install script that we had in the OFN repo a while ago. That script is much more useful and can be used by devs as well. It also makes sure that it upgrades or downgrades correctly. Have a look at https://github.com/openfoodfoundation/openfoodnetwork/pull/3302 and my new commit here.

PR #323, opened 2019-02-08
Enable Italian in Katuma

sauloperez:
This will allow our friends from Piemonte to give OFN a try. I hope they like it decide to join OFN!

### DIFF
--- a/inventory/group_vars/es.yml
+++ b/inventory/group_vars/es.yml
@@ -4,7 +4,7 @@ checkout_zone: Europe
 country_code: ES
 currency: EUR
 locale: es
-available_locales: es,ca,pt
+available_locales: es,ca,pt,it
 language: es_ES.UTF-8
 language_packages:
   - language-pack-es-base


sauloperez:
> There will be no "English" ("ca" behind the scenes) option after this PR: [openfoodfoundation/openfoodnetwork#3463](https://github.com/openfoodfoundation/openfoodnetwork/pull/3463) Do you plan to create a separate `ofn-install` PR adding "en" to the list?

That was intentional but you made me realize you might want to use the instance from time to time to check things. Do you want me to add English as well?

sauloperez:
I just provisioned it in staging.katuma.org and after rebooting the unicorn it successfully picked up the current Italian translation :ok_hand:

kristinalim:
> That was intentional but you made me realize you might want to use the instance from time to time to check things. Do you want me to add English as well?

@sauloperez I see. :slightly_smiling_face: No issue, I just wanted to check with you if it wasn't intentional and maybe some real users would need it.

PR #329, opened 2019-02-26
Remove all traces of bugsnag initializer

sauloperez:
This has been moved to the app's `config/initializers/` instead where it doesn't need neither templates nor symlink. It's just about providing the BUGSNAG_API_KEY env var, which is set in both the unicorn service and `/etc/default/openfoodnetwork`.

As a result, it'll fix the issue that both Katuma and OFF are facing where the `config/initializer/bugsnag.rb` symlink to `shared/config/bugsnag.rb` does not exist thus, nothing gets notified to Bugsnag since December 3rd (according to customer support).

### DIFF
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -35,18 +35,10 @@
     - { src: "seed.sh.j2", dest: "{{ config_path }}/seed.sh" }
   tags: app_templates
 
-# Note this requires the `bugsnag_key` var to be specified otherwise no
-# template will be copied
-- name: copy the bugsnag's initializer file into the shared folder
-  template:
-    src: "bugsnag.rb.j2"
-    dest: "{{ config_path }}/bugsnag.rb"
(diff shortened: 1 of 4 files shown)


sauloperez:
> Can you remind me again why setting the environment variables in defaults and unicorn service file is better than in application.yml?

I don't know if it's better but that is what I'd normally do. Then I remembered that we have application.yml. So I'll use that file instead to stick to a single solution.

sauloperez:
Tested and it worked.

sauloperez:
AFAIK among the non-ofn-install servers the only one using Bugsnag is AUS, isn't it @mkllnk?

sigmundpetersen:
Scandinavia is using Bugsnag :trollface:

sauloperez:
This is seriously frustrating I wish I could lock an entire week to get over it :sob:

mkllnk:
I applied this change to Aus and Can, easy. It was quicker than running Ansible.

PR #334, opened 2019-03-07
Update Canadian variables

mkllnk:
I created the ofn-admin user on the Canadian server. It's still not ready for full ofn-provisioning because it's an old Ubuntu version but here are at least the basic variables.


inventory/group_vars/ca.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+# See https://github.com/openfoodfoundation/ofn-install/wiki/Setup for more info
+checkout_zone: Canada
+country_code: CA
+currency: CAD
+locale: en
+available_locales: en,es    # Available locales for OFN instance
```
sauloperez: wait. Is `es` right here? shouldn't it be `fr_CA` instead?

mkllnk: Well spotted! It was originally provisioned with es as an example but got changed to fr_CA. I added another commit with the missing timezone as well.

luisramos0:
ready to go?

PR #337, opened 2019-03-12
Enable nl_BE locale for the Belgian instance

sauloperez:
Last time we deployed a custom branch based on v1.27.0 plus nl_BE but it was one-time change that needs to be persisted here.

Now Theo complains the server lost its nl_BE translation. See https://openfoodnetwork.slack.com/archives/CFU9N7TGB/p1552397538005900


luisramos0:
@Matt-Yorkley if you are the second reviewer you should move the PRs to test ready, ok? ;-)

PR #346, opened 2019-03-26
Fix fatal missing database errors when doing a fresh install

### LINKED ISSUE #345: Database missing error on fresh install
When running the playbooks for the first time on a fresh install I get this error:

```
TASK [deploy : create a database rollback version] ***********************************************************
Tuesday 26 March 2019  18:11:18 +0000 (0:00:00.171)       0:07:55.167 ********* 
fatal: [local_vagrant]: FAILED! => {"changed": true, "cmd": ["bash", "-c", "set -o pipefail && pg_dump -h 'localhost' -U 'ofn_user' 'openfoodnetwork' | gzip > '/home/openfoodnetwork/apps/openfoodnetwork/rollback.sql.gz'"], "delta": "0:00:01.924317", "end": "2019-03-26 18:11:23.702867", "msg": "non-zero return code", "rc": 1, "start": "2019-03-26 18:11:21.778550", "stderr": "pg_dump: [archiver (db)] connection to database \"openfoodnetwork\" failed: FATAL:  database \"openfoodnetwork\" does not exist", "stderr_lines": ["pg_dump: [archiver (db)] connection to database \"openfoodnetwork\" failed: FATAL:  database \"openfoodnetwork\" does not exist"], "stdout": "", "stdout_lines": []}
```

failed: FATAL:  database \"openfoodnetwork\" does not exist".

At the point this task is run, the database hasn't been created yet. :see_no_evil: 

I have a feeling the recent refactoring to comply with the latest ansible-lint warnings has changed the way these tasks pass or fail, but I could be wrong.

Anyway, we should move the tasks that create the DB to run before the tasks that start trying to access them.

Matt-Yorkley:
Closes #345

I've changed the order of a few of these tasks so that we don't get fatal error on a fresh install.

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -99,6 +99,21 @@
   set_fact:
     timestamp: "{{ ansible_date_time.date }}-{{ ansible_date_time.hour }}{{ ansible_date_time.minute }}{{ ansible_date_time.second }}"
 
+#--------------------------
+# Check the database status
+
+- name: check database status
+  command: psql -h {{ db_host }} -U {{ db_user }} -d {{ db }} -c "SELECT true FROM pg_tables WHERE tablename = 'order_cycles';"
+  register: table_exists
+  ignore_errors: yes
+  tags:
+    - rake
+    - skip_ansible_lint
+
+- debug: msg="{{ table_exists.stdout }}"
+  tags: rake
+- debug: msg="{{ table_exists.stderr }}"
+  tags: rake
 
 #--------------------------
 # Create a rollback release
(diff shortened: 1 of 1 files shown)


roles/deploy/tasks/main.yml
```diff
@@ -99,37 +99,8 @@
...
-- name: move build to current
-  command: mv {{ build_path }} {{ current_path }} removes={{ build_path }}
-  become: yes
-  become_user: "{{ unicorn_user }}"
-
 #--------------------
-# Update the database
+# Create database if it doesn't exist
 
 - name: check database status
```
Matt-Yorkley: The diff isn't very clear here, but basically the tasks "check database status", "create database if it doesn't exist" and "load schema if it isn't loaded" have been shifted upwards, so they run before the task "create a database rollback version".

PR #348, opened 2019-03-26
Bionic beaver

### LINKED ISSUE #343: Package "python-software-properties" has been obsoleted
The `python-software-properties` package (listed in our packages in ofn-install) has been obsoleted in 18 and it looks like it's been replaced/superceded by package: `software-properties-common`.

```
Package python-software-properties is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  software-properties-common

E: Package 'python-software-properties' has no installation candidate

```

It looks like `software-properties-common` is also available in 16, so maybe we can just switch it out...

### LINKED ISSUE #344: Ruby 2.15 install/build fails
We install ruby via the community plugin `zzbet.rbenv`. When it gets to the install task, it fails in Ubuntu 18.

Output:

```
Downloading ruby-2.1.5.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.1/ruby-2.1.5.tar.bz2
Installing ruby-2.1.5...
WARNING: ruby-2.1.5 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.
BUILD FAILED (Ubuntu 18.04 using ruby-build 20190320)
Inspect or clean up the working tree at /tmp/ruby-build.20190326160938.15756
Results logged to /tmp/ruby-build.20190326160938.15756.log
Last 10 log lines:
make[1]: *** [ext/openssl/all] Error 2
make[1]: *** Waiting for unfinished jobs....
compiling callback-7.c
compiling callback-8.c
installing default callback libraries
linking shared-object dl/callback.so
make[2]: Leaving directory '/tmp/ruby-build.20190326160938.15756/ruby-2.1.5/ext/dl/callback'
make[1]: Leaving directory '/tmp/ruby-build.20190326160938.15756/ruby-2.1.5'
uncommon.mk:180: recipe for target 'build-ext' failed
```

Matt-Yorkley:
Closes #343 
Closes #344 

This PR bring compatibility for Ubuntu 18 (Bionic Beaver) to ofn-install. I've tested the `setup`, `provision`, and `deploy` playbooks with these changes on both Ubuntu 16 and Ubuntu 18 boxes. Both are running fine, and the fatal errors are fixed for Ubuntu 18.

Additional notes are in the 2 issues above.

Note: this requires an update of the community roles with:
`ansible-galaxy install -r bin/requirements.yml --force`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,7 +2,7 @@
 # vi: set ft=ruby :
 
 # Run the following command to set up this Vagrant box:
-# ansible-playbook site.yml -i inventory/dev --limit=vagrant
+# ansible-playbook site.yml --limit=vagrant
 
 # It takes around 20 minutes to complete the first setup/provision/deploy.
 # Your new local OFN instance will then be available at: http://localhost:8080
@@ -16,7 +16,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   config.vm.provider :virtualbox do |vbox|
-    config.vm.box = "ubuntu/xenial64"
+    config.vm.box = "ubuntu/bionic64"
 
     # Set box memory.
     vbox.customize ["modifyvm", :id, "--memory", "1792"]
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -17,7 +17,6 @@
       - python-pycurl
       - python-psycopg2
       - python3-psycopg2
-      - software-properties-common
 
       # unknown why or if needed
       # The playbooks run without these dependencies,
@@ -43,12 +42,6 @@
       - libxslt1-dev
       - memcached
   tags: packages
-
-- name: define additional build dependencies for Ruby 2.1.5 on Ubuntu 18
-  set_fact:
-    rbenv_extra_depends:
-      - libssl1.0-dev
-  when: ansible_distribution_major_version == '18'
 
 # Remove monit if still installed. The notified handlers will run only once
 # (when the main task triggers a change by removing the config file)
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -17,7 +17,7 @@
       - python-pycurl
       - python-psycopg2
       - python3-psycopg2
-      - python-software-properties
+      - software-properties-common
 
       # unknown why or if needed
       # The playbooks run without these dependencies,
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -44,6 +44,12 @@
       - memcached
   tags: packages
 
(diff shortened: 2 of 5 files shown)


roles/common/tasks/main.yml
```diff
@@ -17,7 +17,7 @@
       - python-pycurl
       - python-psycopg2
       - python3-psycopg2
-      - python-software-properties
```
mkllnk: Can we make this change only for Ubuntu 18, maybe with `when: ansible_distribution_major_version == '18'`? I'm worried we break our current setups when merging this to master.

Vagrantfile
```diff
@@ -16,7 +16,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   config.vm.provider :virtualbox do |vbox|
-    config.vm.box = "ubuntu/xenial64"
+    config.vm.box = "ubuntu/bionic64"
```
mkllnk: Is there a way we can configure this? It would be nice to test both Ubuntu versions.

roles/common/tasks/main.yml
Matt-Yorkley: Yeah, we could do that. I see this dependency was added in 2014, and probably to support Ansible. I'm not sure it's needed any more?

Matt-Yorkley:
I've tested this thoroughly on a Vagrant box, but I think the configuration skips a few bits like certbot setup. It would be good to test it on a live server before merging.

roles/common/tasks/main.yml
Matt-Yorkley: I'll take a look at extracting these bits out into a separate role where the compatibility-specific code can live.

Matt-Yorkley: Done.

Vagrantfile
Matt-Yorkley: You can change that line at any time, and do `$ vagrant destroy && vagrant up`. I could look at having multiple Vagrant setups, but does anyone else use it much?

roles/common/tasks/main.yml
Matt-Yorkley: I just found an Ansible note here: https://docs.ansible.com/ansible/2.4/intro_installation.html#latest-releases-via-apt-ubuntu that says:

`On older Ubuntu distributions, “software-properties-common” is called “python-software-properties”.`

It looks like it was around in Ubuntu 12, but has been slowly deprecated since then.

Matt-Yorkley:
@sauloperez Yeah, that sounds like a good idea. We could use French or Katuma staging maybe?

mkllnk:
I'm sorry, I accidentally merged this. I reverted it again and opened another pull request.

sauloperez:
Sounds good @Matt-Yorkley. We had to renew that machine anyway.

PR #352, opened 2019-03-27
Change languages in staging FR so we have english for english testers

luisramos0:
Changing languages in staging FR as per request from @RachL 
we will now have english in staging FR.

this is done in staging AUS [here](https://github.com/openfoodfoundation/ofn-install/blob/c55042af0cc313f27c16da9e664dd8babe5079b7/inventory/host_vars/staging.openfoodnetwork.org.au/config.yml#L11).

But I tried to deploy this from my local ofn-install and although the application.yml file was rewritten with the deploy in /shared, the languages were not updated and remained the default fr, it specified in group_vars/fr.yml

I thought [this](https://github.com/openfoodfoundation/ofn-install/blob/c55042af0cc313f27c16da9e664dd8babe5079b7/roles/app/templates/application.yml.j2#L12) would pick up stuff from inventory_vars as well.

### DIFF
--- a/inventory/host_vars/staging.openfoodfrance.org/config.yml
+++ b/inventory/host_vars/staging.openfoodfrance.org/config.yml
@@ -7,4 +7,6 @@ admin_email: admin@openfoodfrance.org
 
 mail_domain: openfoodfrance.org
 
+available_locales: fr,en
+
 unicorn_timeout: 120


Matt-Yorkley:
It's normally set in the corresponding file under `/inventory/group_vars/fr.yml`. It's already set there as: `fr,it`, so I think that value is not being overridden by your addition.

In Australia's case they don't have it set in group_vars at all, soI think that's why it works for their for staging.

Matt-Yorkley:
Yep: https://docs.ansible.com/ansible/2.3/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

I'm not sure whats' the best way to override it per environment in this case...

luisramos0:
yeah, thanks for feedback and link Matt.
I think I'll have to try again.

sigmundpetersen:
So is this going back to In Dev @luisramos0 ?

luisramos0:
yes :+1:

mkllnk:
@luisramos0 It has been almost a year. Is it time to pick it up again?

luisramos0:
it's hasnt worked and it has not been a priority, I will close it for now.

PR #353, opened 2019-03-27
Set staging servers to update themselves

Matt-Yorkley:
I havn't tracked my time for this as it's not been prioritised, it was a 5-minute job, but I think this could save a lot of time when we make important updates to ofn-install.

Staging servers that have Ansible installed locally for deploying themselves via Semaphore can now also update their local copy of ofn-install by pulling from the repo, and upgrade their community roles if the versions have been bumped in the `requirements.yml` file.


Matt-Yorkley:
I pushed this to UK staging and tested it. It works nicely! I've updated French and Katuma Staging with the new version of this script as well. :+1:

PR #356, opened 2019-03-28
Triple Unicorn Rainbow Explosion

### LINKED ISSUE #355: Increase number of unicorn workers for UK prod
UK (like other instances) is experiencing intermittent site failures. When this happens any user that tries to access any page on [openfoodnetwork.org.uk](openfoodnetwork.org.uk) gets an error meaning that to the user is appears that OFN UK is totally down. This goes away on a refresh (or possibly two refreshes).

**Possible Cause**
After many conversations with @luisramos0 and @sauloperez I have come to deduce a probable cause. UK (like most instances) has two unicorn workers. When both of these workers are occupied with other requests no other users can access the site at all. This is increasingly common due to the many performance issues we face in OFN (and we won't go into these here). Two large reports or data heavy pages loading at the same time can take minutes and stop others accessing OFN during this time.

**Relevant Background**
We have two unicorn workers based on experimentation in Australia (according to @mkllnk). The number of unicorn workers is limited by the server memory. Another possible limitation is the number of DB connections, though I am sure that our connection pools have more than 2 connections so this shouldn't be a direct limitation in the same way that the number of workers is.

@sauloperez informs me that, based on Datadog monitoring, UK is not using close to its maximum memory. Therefore we feel confident we can try increasing the number of unicorn workers to 3 on UK prod to see if this helps the issue.

**Possible Solution**
Updating configuration such that unicorn starts with 3 workers in UK servers only. If this is successful other instances might be interested in doing this too.

Matt-Yorkley:
Closes #355

This increases the `unicorn_workers` to 3 on UK production and sets 1GB of swap. Hopefully the catchy title will get it approved and merged quicker. :wink: 

:unicorn: :unicorn: :unicorn:

### DIFF
--- a/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
@@ -3,4 +3,5 @@
 domain: www.openfoodnetwork.org.uk
 rails_env: production
 
+unicorn_workers: 3
 unicorn_timeout: 120
--- a/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
@@ -5,3 +5,5 @@ rails_env: production
 
 unicorn_workers: 3
 unicorn_timeout: 120
+
+swapfile_size: 1G


lin-d-hop:
https://gph.is/1ON38cv

lin-d-hop:
https://giphy.com/gifs/3o85xBO5Rt4031bH44/html5

Matt-Yorkley:
Note: this has now been provisioned to UK Prod.

mkllnk:
@Matt-Yorkley I don't see an improvement for the UK server. I checked the `unicorn.rb` on the server and it now says `3`. But then I checked how many processes are running and there are only two workers. The modification of the config didn't trigger a unicorn reload.

Now, I just did `sudo systemctl restart unicorn_openfoodnetwork.service` and there are three workers. Let's watch the server performance.

You may want to check ofn-install for the missing unicorn reload.

Matt-Yorkley:
Good catch! I thought it would get restarted when it was provisioned, I'll take a look at notifying the unicorn handler for a restart. :+1:

sauloperez:
Yes! we only notify the handler on deploy AFAIK. 

> I'll take a look at notifying the unicorn handler for a restart.

But just on unicorn.rb changes, if possible.

mkllnk:
It looks like the UK server improved but there is still a little bit of downtime.
Screenshot from 2019-04-09 17-48-24

I would go up to 5 workers in the UK. Aus is almost without downtime with 4 workers:
Screenshot from 2019-04-09 17-50-23

lin-d-hop:
Thanks Maikel! Really useful. @Matt-Yorkley would you make the switch to 5 workers. Looks like we have available memory without upgrading servers
 \ (•◡•) /

sauloperez:
Agree that adding workers seems to be solving the issue but watch out. We need some room for spikes in RAM and we will also increase Postgres memory usage :warning:.

So I'd like to check RAM in Datadog and probably go with 4 workers first.

lin-d-hop:
I love it when anyone comments on this issue. Nothing like a triple unicorn rainbow explosion inbox notification.
Cool, quadruple unicorn sounds explosively rainbow for now... thanks for monitoring RAM!

RachL:
Hey guys, is this something we should consider doing on French production as well?

RachL:
@sauloperez sure, no rush! I was just asking :)

PR #358, opened 2019-03-28
Remove unused playbooks

### LINKED ISSUE #341: Where is the Ruby Update playbook used
I am starting a task to cover the #157 in the Katuma servers and we are looking the playbooks that we are using.

I found the playbook `ruby_update.yml` and I don't understand where this playbook is used. 

https://github.com/openfoodfoundation/ofn-install/blob/master/playbooks/ruby_update.yml

The tasks that are done in this playbook also are done in the `provision.yml` playbook.

https://github.com/openfoodfoundation/ofn-install/blob/master/playbooks/provision.yml#L32

Is this playbook needed?

Thanks!

### LINKED ISSUE #347: Remove old playbooks
We've got some old playbooks lying around that can probably be removed.

- update_ruby.yml
- verify.yml

This was mentioned as part of the coopdevs team looking at the upgrade to Ubuntu 18. I've had a look and I think we can scrap them.

Matt-Yorkley:
Closes #347
Closes #341 

Removes outdated playbooks (pending consensus).


playbooks/development.yml
```diff
@@ -1,37 +0,0 @@
----
```
sauloperez: This one is used by `scripts/provisioning-dev.sh` but I'm not sure anyone is using it but it'll come in handy when we start changing monitoring and logging settings for Postgres for instance.

Matt-Yorkley: Ah I see, it's used with the LXC dev setup. Do we know who actually uses it?

sauloperez: I know @enricostano and @danypr92 used to but they don't do any OFN since ages. I just wonder if I could make use of it while playing with postgres.conf soon.

Matt-Yorkley: The whole `/scripts` folder is just for setting up LXC containers as dev environments. If nobody uses it, we can remove the folder.

Matt-Yorkley: Ah, ok. Well lets leave it then.

sauloperez: yep, let me give it a try and see if it still works.

Matt-Yorkley: I've reverted the deletion of `development.yml`. :+1:

mkllnk: If I remember correctly, that was a community contribution a few years ago. I don't think any core devs ever used it.

sigmundpetersen:
Guess this is closing #341 as well

Issue #227, opened 2018-07-27
Refactor postgresql tasks


Matt-Yorkley:
We have lots of hacky postgresql tasks, including one that runs an old bash script from 4 years ago to "fix pg encoding": https://github.com/openfoodfoundation/ofn-install/blob/98d35374ab99620ef95cc9cec15e470a65156219/roles/dbserver/files/fix_pg_encoding.sh

There are **6** ansible postgresql modules for doing just about anything with postgresql, and we currently don't use any of them. They are listed here: https://docs.ansible.com/ansible/latest/modules/list_of_database_modules.html#postgresql

#techdebt

sauloperez:
I have no experience with it but, at Coopdevs we use https://github.com/geerlingguy/ansible-role-postgresql in all our projects.

Matt-Yorkley:
Closed by #400

Issue #99, opened 2017-12-03
Enforce version of Postgres


sauloperez:
## What?

We should have a way to enforce what version of Postgresql, and generally, any package, gets installed on our servers. When provisioning the machine's state should be totally predictable.

What I've found is that we install the postgresql package without any version specified and so we rely on the latest version available in the ubuntu distribution being used (I guess).

enricostano:
[Here](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/dbserver/tasks/main.yml#L9) is how we install the package. We should specify the version also, but keep in mind that we use vanilla Ubuntu package instead of a custom PPA and we don't enforce which Ubuntu version we should use. This can lead to packages not found.

sauloperez:
> we use vanilla Ubuntu package instead of a custom PPA

that was my point with

>  and so we rely on the latest version available in the ubuntu distribution being used (I guess)

which makes our infra depend on the ubuntu distro we decide to use in every particular server :scream:

mkllnk:
The version is now predictable:

https://github.com/openfoodfoundation/ofn-install/blob/d94e7e5eed3b407b1c38ceb580236f803abd73a7/inventory/group_vars/all.yml#L85

Issue #72, opened 2017-05-22
Postgres authentication question


enricostano:
Hi,
does anybody has any idea about why we need to create a `.pgpass` file in the home of the unicorn user?

I'm talking about this specific task: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/dbserver/tasks/main.yml#L21

I don't think we use it in VPS instances, maybe Heroku or something else?

RohanM:
This would make it easier as a user on the VPS to sign in to postgres for debug/maintenance purposes. That's my guess at the intent. What do you think?

enricostano:
I don't know, on a VPS you can just `sudo su postgres` from `ofn-admin` user. Maybe @pmackay can shade some light on this? If we don't use it I would remove it since it adds noise on how we authenticate to postgres and for what purposes.

pmackay:
I think this might have been added by Raf. I’m not sure I can offer
anything additional on best practice here, whether this is worth keeping.

enricostano:
Can we simplify this step removing the things that we don't know / understand? If not we'll keep stuff that only add confusion.

What do you think @RohanM @pmackay @oeoeaio ?

saimonmoore:
Agreed with @enricostano .

elf-pavlik:
I see it added by @rafaqz in https://github.com/openfoodfoundation/ofn-install/commit/c936b67fd8a886188e6e528fa50ed7ee2745bfe2#diff-df92b892760893a0a7c5736ef1451e02

:+1: to removing clutter, if we ever need it again we can find it in git history

sauloperez:
feel free to open a PR @elf-pavlik

mkllnk:
But even when we get rid of that deployment script I still find that .pgpass file very convenient. It also enables other convenience scripts, for example the mirror-db script that copies that database of a server into your local dev environment. It doesn't need to know the password and it doesn't need to connect as ofn-admin.

Is there any downside of having this file?

enricostano:
> Is there any downside of having this file?

Not really. I don't personally like it, but I'm not against either. The point of this issue was just about understanding who uses it and in which instance. I don't think anybody is using it, but I may be wrong.

PR #362, opened 2019-03-31
Configure postgres for production

### LINKED ISSUE #361: Postgresql tuning
This was mentioned somewhere else, but we don't have an issue for it.

Postgresql sets incredibly conservative configuration values by default that are not well-suited to production servers.

There are quite a few "recommended" tuning settings but most of them are based on the resources of the system being tuned (available RAM, etc). I've anecdotally seen references to ~30% performance increases just by applying recommended settings.

Given that we seem to be hitting some performance issues, maybe we should prioritise a spike on this at some point for some quick wins.

Matt-Yorkley:
Closes #361 

This is a draft playbook to configure PostgreSQL for production servers by setting some recommended values.

It might be a good idea to get the Datadog paid plan first, as discussed here: https://community.openfoodnetwork.org/t/making-operations-a-first-class-citizen/1601/5

Then we could try provisioning this configuration to one production server and get some data on the results.

### DIFF
--- a/playbooks/postgres_tuning.yml
+++ b/playbooks/postgres_tuning.yml
@@ -31,6 +31,7 @@
           shared_buffers = {{ ( ansible_memtotal_mb / 4 ) | int }}MB
           effective_cache_size = {{ ( ( ansible_memtotal_mb / 4 ) * 3 ) | int }}MB
           maintenance_work_mem = {{ ( ansible_memtotal_mb / 16 ) | int }}MB
+          wal_buffers = 16MB
         dest: "/etc/postgresql/{{ pg_current_version.stdout }}/main/conf.d/tuning.conf"
       become: yes
       register: tuning_configuration
--- a/playbooks/postgres_tuning.yml
+++ b/playbooks/postgres_tuning.yml
@@ -32,6 +32,7 @@
           effective_cache_size = {{ ( ( ansible_memtotal_mb / 4 ) * 3 ) | int }}MB
           maintenance_work_mem = {{ ( ansible_memtotal_mb / 16 ) | int }}MB
           wal_buffers = 16MB
+          work_mem = 6MB
         dest: "/etc/postgresql/{{ pg_current_version.stdout }}/main/conf.d/tuning.conf"
       become: yes
       register: tuning_configuration
--- a/playbooks/postgres_tuning.yml
+++ b/playbooks/postgres_tuning.yml
@@ -1,0 +1,50 @@
+---
+- name: PostgreSQL production tuning
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+
+  tasks:
+    - name: Find current postgres version
+      shell: psql postgres -c 'SHOW config_file;' | grep etc | egrep -o '[0-9]{1,}\.*[0-9]{1,}'
+      become: yes
+      become_user: postgres
+      register: pg_current_version
+
+    - debug:
+        msg: "Postgres current version: {{ pg_current_version.stdout }}"
+
+    - name: Create postgres conf.d directory
+      file:
+        dest: "/etc/postgresql/{{ pg_current_version.stdout }}/main/conf.d"
+        state: directory
+      become: yes
+
+    - name: Include conf.d directory in postgresql.conf
+      lineinfile:
+        path: "/etc/postgresql/{{ pg_current_version.stdout }}/main/postgresql.conf"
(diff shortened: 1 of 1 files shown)


playbooks/postgres_tuning.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Add tuning configuration
+      copy:
+        content: |
+          shared_buffers = {{ ( ansible_memtotal_mb / 4 ) | int }}MB
+          effective_cache_size = {{ ( ( ansible_memtotal_mb / 4 ) * 3 ) | int }}MB
+          maintenance_work_mem = {{ ( ansible_memtotal_mb / 16 ) | int }}MB
+          wal_buffers = 16MB
+          min_wal_size = 1GB
+          max_wal_size = 2GB
```
sauloperez: Can we make at least these two relative to `ansible_memtotal_mb` as well? We're working on getting a production server with the standard specs at Katuma but we're not there yet. 

I checked pgtune and still with a 2 GB RAM it sets max_wal_size = 2GB :man_shrugging:

sauloperez:
I also think we will need to hold this until we reach consensus on https://github.com/openfoodfoundation/ofn-install/issues/367

Matt-Yorkley:
> What about using an Ansible role such as https://github.com/geerlingguy/ansible-role-postgresql and just configure things through per-host variables if needs be? IMO it's simpler that way compared to having multiple conf files deviating from the standard.

I think using that postgres galaxy role you mentioned would be a big improvement, and there are several issues around that could be grouped under a general postgres refactor: #99, #72, #227.

On the other hand, that role won't optimise the configuration at all, which still I think we should try to do...

Matt-Yorkley:
I put this up as a general proposal for discussion more than anything. Maybe we could remove the `wal_buffers_min` and `wal_buffers_max`? 

UK would definitely like to try some optimisation from minimal to recommended settings, so what I'm trying to avoid with this PR is:
- manually adding a non-standardised configuration to UK production that doesn't appear in ofn-install (and forgetting it's there)
- having to re-add a non-standardised configuration if we migrate the server
- having to manually re-configure hard coded values higher if we add more ram

sauloperez:
> On the other hand, that role won't optimise the configuration at all, which still I think we should try to do...

Of course not but it would be a great foundation to make it easy to manage and iterate its configuration. If it is affordable I would that first we start from nothing anyway.

sauloperez:
Would you guys mind waiting to have a refactored provision.yml using https://github.com/geerlingguy/ansible-role-postgresql? It does it better than us and I feel it'll make a lot easier to iterate on postgres config.

Matt-Yorkley:
> Would you guys mind waiting to have a refactored provision.yml using https://github.com/geerlingguy/ansible-role-postgresql? It does it better than us and I feel it'll make a lot easier to iterate on postgres config. It was my next step.

Sounds good, and we can define a postgres version at the same time. When I was playing with ofn-install in travis it installed PG 11 by default as we don't specify the version _anywhere_.

Okay, I'll do it this week. We can wait until then to merge this. :+1:

Matt-Yorkley:
I've updated this to be a bit cleaner, and it now depends on #400.

I think use of this `conf.d` pattern for storing custom settings is the cleanest way to manage configurations.

Screenshot-2019-04-16-21-56:48

Matt-Yorkley:
> What do these values mean and why are they better than the default?

Similar to the `unicorn_workers` experiments, the default postgres settings are generally regarded to be a bit low for production servers. The idea here is that we can apply some recommended settings that should enhance performance, and monitor the results with Datadog.

Mostly I'm trying to look for ways we can get better performance from what we have, and avoid the downtime we're starting to see regularly on French and UK staging.

Some further reading here: 
https://www.revsys.com/writings/postgresql-performance.html
https://leopard.in.ua/2014/03/24/pgtune-for-postgresql

The playbook is optional, so we could run it on one server as a test and see the results. The playbook also includes a task for removing the configuration from a server.

sauloperez:
I think the main idea @mkllnk is to increase the PostgreSQL's mem usage so it doesn't hit disk that often. This will improve performance in all sorts of tasks the server performs. And it turns out our servers have a considerable amount of RAM available.

playbooks/postgres_tuning.yml
Matt-Yorkley: `min_wal_size` and `max_wal_size` have been removed, so they will stay at their defaults. :+1:

PR #363, opened 2019-04-01
Delete old files in /releases-old on deploy

### LINKED ISSUE #321: Delete old releases after deploy
Each deploy creates a backup in `releases_old`. That fills up the disk over time. I propose a new playbook that deletes some old releases. For example, we could delete every release that is older than three days.

Matt-Yorkley:
Closes #321 

Keep the last 5 pairs of release files in `/releases-old` (backup and dump) and delete the rest when deploying.

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -137,12 +137,21 @@
   command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }} removes={{ rollback_path }}
   tags: skip_ansible_lint
 
-- name: delete old release files
-  shell: set -o pipefail && ls -1tr | head -n -10 | xargs -d '\n' rm -rf --
+- name: check for old release files
+  shell: set -o pipefail && ls -1tr | head -n -10
   args:
     chdir: "{{ releases_path }}"
+    executable: /bin/bash
   become: yes
-  tags: skip_ansible_lint
+  changed_when: False
+  register: releases_for_deletion
+
+- name: delete old release files if present
+  file:
+    path: "{{ releases_path }}/{{ item }}"
+    state: absent
+  with_items: "{{ releases_for_deletion.stdout_lines | list }}"
+  become: yes
 
 #-------------------------
 # Move new code into place
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -142,6 +142,7 @@
   args:
     chdir: "{{ releases_path }}"
   become: yes
+  tags: skip_ansible_lint
 
 #-------------------------
 # Move new code into place
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -137,6 +137,12 @@
   command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }} removes={{ rollback_path }}
   tags: skip_ansible_lint
 
+- name: delete old release files
+  shell: ls -1tr | head -n -10 | xargs -d '\n' rm -rf --
+  args:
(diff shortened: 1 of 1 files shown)


sauloperez:
Looks like the linter is not so happy

Matt-Yorkley:
Okay, there were two linter failures. I fixed the first one, but I don't think it's possible to fix the second one, so I've used the `skip_ansible_lint` tag.

sauloperez:
Feel free to merge if you agree @mkllnk

roles/deploy/tasks/main.yml
```diff
@@ -138,7 +138,7 @@
   tags: skip_ansible_lint
 
 - name: delete old release files
-  shell: ls -1tr | head -n -10 | xargs -d '\n' rm -rf --
+  shell: set -o pipefail && ls -1tr | head -n -10 | xargs -d '\n' rm -rf --
```
mkllnk: Did you test this? I read that the default shell is `sh` which doesn't support the pipefail option. So I executed bash:
https://github.com/openfoodfoundation/ofn-install/blob/459c3d01495a94dc45650be7822b0ff0deb7ddf6/roles/deploy/tasks/main.yml#L122

roles/deploy/tasks/main.yml
```diff
@@ -142,6 +142,7 @@
   args:
     chdir: "{{ releases_path }}"
   become: yes
+  tags: skip_ansible_lint
```
mkllnk: The lint writer would probably be happier about this version:
```patch
diff --git a/roles/deploy/tasks/main.yml b/roles/deploy/tasks/main.yml
index e4d7d15..f59ac77 100644
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -126,6 +126,7 @@
 
 - name: create a database backup version
   command: cp {{ rollback_sql_path }} {{ releases_path }}/{{ timestamp }}.sql.gz
+  register: backup_creation
   tags: skip_ansible_lint
   when: table_exists.stderr.find('does not exist') == -1
 
@@ -135,7 +136,7 @@
 
 - name: create a repo backup version
   command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }} removes={{ rollback_path }}
-  tags: skip_ansible_lint
+  when: backup_creation is succeeded
 
 - name: delete old release files
   shell: set -o pipefail && ls -1tr | head -n -10 | xargs -d '\n' rm -rf --
```

It connects the two steps. There are many lint skips in this file though. The nature of this deploy script doesn't fit well into the Ansible paradigm. I think most of these commands would actually be better in a shell script on the server (which is installed and executed by Ansible). That's out of scope of this PR though.

I don't mind if you leave this as is.

Matt-Yorkley:
@mkllnk I was trying to avoid splitting the job into multiple tasks as it seemed like such a small thing, but you're right, it needed to be a bit nicer. I've refactored and thoroughly tested. It's now beautiful, idempotent Ansible that correctly reports changes.

Notes on the improvement:
- pipes don't work with the `command` module, which is why I opted for `shell` originally. We can specify `executable: /bin/bash` in the args of the `shell` module though :+1: 
- the first part of the task is now just registering a variable, so we can use `changed_when: False` to tell Ansible this command is not changing the system. That's what the linter was complaining about with the original command, as the task didn't correctly report when it was changing or not changing the system.
- the second part of the command now uses the `file` module to delete the list of files, and that module correctly records if the system was changed or not, so Ansible lint is happy and the code is much better :+1: 

A brief look at the other tasks in our playbooks suggests we can remove all the ansible-lint skips quite easily with some very simple improvements/refactoring.

PR #364, opened 2019-04-02
Rescue and notify Slack when deployment fails

Matt-Yorkley:
We can now get notified when deployment fails as well as when it succeeds, using Ansible's block/rescue format in the `deployment.yml` playbook.

Tested on Vagrant (both success and failure paths).

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -5,5 +5,16 @@
   become: yes
   become_user: "{{ unicorn_user }}"
 
-  roles:
-    - role: deploy
+  tasks:
+    - block:
+      - include_role:
+          name: deploy
+
+      rescue:
+        - name: Notify slack of deployment failure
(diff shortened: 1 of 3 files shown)


Matt-Yorkley:
FYI: I'm pretty sure we could check if the files and database have been changed yet during the current playbook run-though, and optionally include a task to automatically rollback the server to the previous backup if an error occurs. Just an idea...

sauloperez:
> FYI: I'm pretty sure we could also check if the database migration task failed during the current playbook run-though, and optionally include a task to automatically rollback the server to the previous backup in that case. Just an idea...

I'm not sure that would be useful in all cases. Most of the times we will want to go check what failed and decide ourselves what to do. Also, most of the times I bet the failure will not cause data loss and so restoring a backup would we pointless.

PR #366, opened 2019-04-03
Check secrets have been included

### LINKED ISSUE #365: Exit playbook if secrets are not present
We should probably exit the deploy or provision playbooks _early_ if the necessary secrets have not been included due to user error, and show a message.

Matt-Yorkley:
Closes #365 

Provision and deploy playbooks will now exit early if the secrets have not been included correctly.

Tested (both paths) on Vagrant.

### DIFF
--- a/roles/check_secrets/tasks/main.yml
+++ b/roles/check_secrets/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
 - name: Check secrets
   fail:
-    msg: "ERROR: It looks like your secrets.yml file is missing..."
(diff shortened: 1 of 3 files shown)


roles/check_secrets/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Check secrets
+  fail:
+    msg: "ERROR: It looks like your secrets.yml file is missing..."
```
sauloperez: We could be nice to users and explain how to provide them. I would add something like "specify them using providing the argument '-e @path/to/secrets.yml'"

Matt-Yorkley: Done.

Issue #280, opened 2018-11-19
Move config/initializers/db2fog.rb to the app


sauloperez:
## Description

Something as simple as having a `config/initializer/` to configure `db2fog` is overly complex and confusing, which seriously affects our confidence when changing these settings (at least mine). Also, what should be straightforward takes too much time.

## How it works now

Currently our servers end up with a symlink pointing to the config stored in `shared/` as you can see below:

```
$ ll config/initializers/db2fog.rb 
lrwxrwxrwx 1 openfoodnetwork openfoodnetwork 66 nov 16 09:56 config/initializers/db2fog.rb -> /home/openfoodnetwork/apps/openfoodnetwork/shared/config/db2fog.rb*
```

This is done in 2 steps.

### Copy file to shared folder

First, the file is copied to the shared folder

https://github.com/openfoodfoundation/ofn-install/blob/48476c4f012325350f55211820c83b8a362e0f28/roles/app/tasks/main.yml#L51-L60

### Symlink the db2fog file into app config

Then, the config needs to be made available to the app by means of a symlink.

https://github.com/openfoodfoundation/ofn-install/blob/48476c4f012325350f55211820c83b8a362e0f28/roles/deploy/tasks/main.yml#L70-L79

All this was done this way to be able to get `{{ s3_backups_bucket }}`'s value (note it's an ansible var) and print it into the Db2fog config object.

https://github.com/openfoodfoundation/ofn-install/blob/48476c4f012325350f55211820c83b8a362e0f28/roles/app/templates/db2fog.rb.j2#L2-L7

## Expected behavior

We should only have a `config/initializers/db2fog.rb` file in the app that gets the `{{ s3_backups_bucket }}`'s value from an ENV var.

To do that, we need to decide which way we populate said ENV var. Either from [roles/unicorn_user/templates/defaults.j2](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/unicorn_user/templates/defaults.j2) or [roles/app/templates/application.yml.j2](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/application.yml.j2)

We also need to check what happens then with [/config/initializers/0_depricated_db2fog.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/0_depricated_db2fog.rb). Can we remove it? Because as far as I understand, it will clash with the new file.

Matt-Yorkley:
Proposal:
[...]

PR #369, opened 2019-04-03
Fetch secrets when deploying with Semaphore

Matt-Yorkley:
This is a proposal to modify the Semaphore deployment system so that a server will automatically fetch the latest copy of it's own secrets file from the new private repo when it's deploying itself.

There's a couple of contentious bits:
- we provision the vault password into a file under `/home/ofn-admin/`, owned by `root` with 600 permissions as a one-off event when the `setup_semaphore_deployment.yml` is run against the server we're setting up. It's handled via a secure prompt, so the sysadmin provisioning it has to manually put in the password.
- in order to pull from the private repo we use a bot account, give it read access on the secrets repo, and create a Personal Access Token in the bot's account which gives access to read files in that repo and nothing else. We can then use that token programmatically to pull from the secrets repo.
- we save that token into Semaphore, using an encrypted environment variable in the Semaphore secrets UI. The secret can't be read or edited by any Semaphore users after that point (it can be deleted). We then pass it to the server that's being deployed so it can fetch the (encrypted) secrets and decrypt them with the vault-pass file.

Simple, right? :trollface:

### DIFF
--- a/playbooks/setup_semaphore_deployment.yml
+++ b/playbooks/setup_semaphore_deployment.yml
@@ -4,5 +4,10 @@
   remote_user: "{{ user }}"
   become: yes
 
+  vars_prompt:
+    - name: ofn_vault_pass
+      prompt: "Set vault password (or enter to skip)"
+      private: yes
+
   roles:
     - role: semaphore_deployment
--- a/roles/semaphore_deployment/tasks/main.yml
+++ b/roles/semaphore_deployment/tasks/main.yml
@@ -53,6 +53,7 @@
 - name: copy public key
   shell: "cat /home/{{ user }}/keys/semaphore.pub"
   register: semaphore_public_key
+  changed_when: False
 
 - name: add semaphore public key to deployment user's authorized_keys
   authorized_key:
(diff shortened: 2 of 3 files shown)


sauloperez:
Wouldn't storing the vault password in the Semaphore secrets UI make this a bit simpler? It's a 3rd but... We are not security expects either, are we?

Matt-Yorkley:
We could definitely store it in Semaphore as well, and that might be marginally simpler. I felt it was nicer to not keep both the ability to access the secrets and the means to encrypt them in the same place and held by a third party.

sauloperez:
That is legit too :+1:

mkllnk:
Okay, what's the risk? If someone compromised one server, they would have access to all encrypted secrets. If every file has a different password, the attacker can only read the secrets of the host which is already compromised and you would find those secrets in the application.yml as well. I don't think we are increasing the risk by much here. :+1:

Matt-Yorkley:
I've just realised if we finally do this: #280, then the deploy playbook actually has no dependencies on the secrets file at all, and we can deploy without it. :see_no_evil: And we can tidy up and simplify things a lot. :tada:

mkllnk:
Awesome thought!

sauloperez:
I really want to get rid of annoyances (I'm tired of things being stuck forever) so #280 is done. I hope no one thinks I hacked the process. It wasn't my intention.

PR #370, opened 2019-04-04
Remove db2fog initializer from ofn-install

### LINKED ISSUE #280: Move config/initializers/db2fog.rb to the app

Matt-Yorkley:
Closes #280 

Removes the db2fog initializer from ofn-install after it is moved to the main repo in: https://github.com/openfoodfoundation/openfoodnetwork/pull/3694

This also frees up the `deploy.yml` playbook from all dependencies on secrets files, so deploying from the command line will be much simpler and deploying via Semaphore will not involve any secrets handling. :tada: 

#### Dependencies

openfoodfoundation/openfoodnetwork#3694

### DIFF
--- a/roles/app/tasks/main.yml
+++ b/roles/app/tasks/main.yml
@@ -40,17 +40,6 @@
     path: "{{ config_path }}/bugsnag.rb"
     state: absent
 
-- name: copy the db2fog file into the shared folder
-  template:
-    src: db2fog.rb.j2
-    dest: "{{ config_path }}/db2fog.rb"
-    owner: "{{ unicorn_user }}"
-    mode: 0775
-  tags:
-    - app_templates
-    - db2fog
-  when: s3_backups_bucket is defined
-
 - name: get l10n repo
   git:
     repo={{ l10n_repo }}
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -19,7 +19,7 @@ CHECKOUT_ZONE: {{ checkout_zone }}
 GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
 
 {% if s3_backups_bucket is defined %}
-S3_BACKUPS_BUCKET: {{ s3_backups_bucket }}
+S3_BUCKET: {{ s3_backups_bucket }}
 S3_ACCESS_KEY: {{ s3_access_key }}
 S3_SECRET: {{ s3_secret }}
 {% endif %}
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
Note: I just discovered if we're setting the bucket variable in application.yml, it has to be named `S3_BUCKET` instead of `S3_BACKUP_BUCKET`, otherwise Spree::Config will throw a fatal error, as `Spree::Config[:s3_bucket]` is a predefined key, and `Spree::Config[:s3_backup_bucket]` does not exist.

PR #373, opened 2019-04-05
Do not check secrets as a requirement to deploy

### LINKED ISSUE #371: Remove secrets check from deployment
When the changes to the db2fog initializer are finished in https://github.com/openfoodfoundation/openfoodnetwork/pull/3694 and #370 we can update the code added in #366 to only validate the presence of secrets during provisioning and not during deployment.

sauloperez:
Closes #371

As explained in https://github.com/openfoodfoundation/ofn-install/issues/371 we no longer need secrets while deploying so there is no need for this check anymore :tada:.

This should fix the last Semaphore deployment failure https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/servers/fr-staging/deploys/8.

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -5,9 +5,5 @@
   become: yes
   become_user: "{{ unicorn_user }}"
 
-  pre_tasks:
-    - include_role:
-        name: check_secrets
-
   roles:
     - role: deploy


sauloperez:
Still, I don't understand why the `@/home/ofn-admin/secrets.yml` in the deploy script didn't work

PR #374, opened 2019-04-05
Deployment update

Matt-Yorkley:
Removing some more code as deployment is no longer coupled to secrets.

### DIFF
--- a/roles/semaphore_deployment/templates/deploy.j2
+++ b/roles/semaphore_deployment/templates/deploy.j2
@@ -18,13 +18,5 @@ cd "/home/{{ deployment_user }}/ofn-install"
 # Pull any recent ofn-install updates
 git pull
 
-# Update community roles if the requirements have changed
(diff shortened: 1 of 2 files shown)


roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -18,13 +18,5 @@ cd "/home/{{ deployment_user }}/ofn-install"
 # Pull any recent ofn-install updates
 git pull
 
-# Update community roles if the requirements have changed
-if $(ansible-galaxy install -r bin/requirements.yml 2>&1 | grep -q "change version"); then
-    echo "Updating community roles..."
-    ansible-galaxy install -r bin/requirements.yml --force
-else
-    echo "All community roles are up to date"
-fi
```
sauloperez: we do it somewhere else already or what?

Matt-Yorkley: I realised the deploy playbook actually doesn't call any of the community roles, so this isn't needed. :fire:

PR #382, opened 2019-04-07
Specify depth when fetching repo

Matt-Yorkley:
This is equivalent to `--depth 1`, meaning we fetch only the latest version of each file and not the entire history. 

It reduces the execution time of this task from ~2:30 to ~0:40. :rocket:


Matt-Yorkley:
This is causing issues with Semaphore deployment. The Ansible git module has an issue where using depth along with a custom refspec (which we need for Semaphore deployments) fails.

I think we'll have to revert this and lose the speed increase... :sob:

PR #386, opened 2019-04-10
Allow provisioning of different buckets

Matt-Yorkley:
This allows for provisioning of different buckets for images and backups, and enables provisioning of the `Spree::Config['s3_bucket']` preference used for images, which was previously only accessible via the superadmin config page.

This is a partial patch for recent changes in #370 

**Configuration:**
The use of `s3_backups_bucket` variable is unchanged.
The new `s3_images_bucket` will populate the `Spree::Config[:s3_bucket]` preference which is used for images. Unless that variable is explicitly set in the secrets file, no existing manually configured settings will be changed.

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -18,10 +18,17 @@ CHECKOUT_ZONE: {{ checkout_zone }}
 
 GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
 
-{% if s3_backups_bucket is defined %}
-S3_BUCKET: {{ s3_backups_bucket }}
+{% if s3_access_key is defined %}
 S3_ACCESS_KEY: {{ s3_access_key }}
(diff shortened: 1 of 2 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -18,11 +18,18 @@ CHECKOUT_ZONE: {{ checkout_zone }}
 
 GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
 
-{% if s3_backups_bucket is defined %}
-S3_BUCKET: {{ s3_backups_bucket }}
+{% if s3_access_key is defined %}
 S3_ACCESS_KEY: {{ s3_access_key }}
+{% endif %}
+{% if s3_access_key is defined %}
```
sauloperez: I guess you meant `s3_secret` here. Otherwise we can merge the block with previous conditional.

PR #391, opened 2019-04-12
Create europe group

### LINKED ISSUE #313: Enable deployment of mulitple instances at once
## Description

Today it's been the first we deploy the latest release to a bunch of instances together (European ones) instead of having to wait for everyone to update their instances, which results in inefficiencies.

A problem I saw is that I wasn't entirely sure how Ansible would pick up and distinguish between each instance's secrets. To avoid problems I chose to deploy them one by one although what I really wanted was to do deploy them at the same time. In my mind, there shouldn't be any difference between our setup and an app that has multiple app and DB servers. 

In any case, I found the process tremendously error-prone. It was too easy to get French secrets and use them to deploy to UK, for instance.

## Proposed solution

First, we'll need to have a group defined so that we can do something like `--limit=europe`.

Then, we need to check https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#group-and-host-variables and see what's the best for Ansible to differentiate the secrets.

However, https://github.com/openfoodfoundation/ofn-install/issues/307 needs to be taken into account. Any solution we implement needs to work with that approach as well or should be easy to refactor to adapt to it.

Matt-Yorkley:
Closes #313 

Adds a `europe` group in the hosts file, so playbooks can be run against multiple servers at once.

Hot tip: you can test which servers would be included in a command by using `--limit europe` and `--list-hosts`.

### DIFF
--- a/inventory/hosts
+++ b/inventory/hosts
@@ -126,6 +126,16 @@ www.openfoodnetwork.be ansible_host=51.75.124.147
 be-prod
 
 #------------------------------------------------------------------------------
+# Europe
+
+[europe:children]
+be-prod
+es-prod
+fr-prod
+uk-prod
+#de-prod
+
+#------------------------------------------------------------------------------
 # All
 
 [ofn_servers:children]


enricostano:
OMG :scream: :heart_eyes:

PR #396, opened 2019-04-16
Rollbacks

### LINKED ISSUE #394: Fix rollback playbook
I think nobody has used it for a long long time, and it looks like it's not working (error reported by Pau). 

The playbook looks okay, so I have a suspicion it was written with an old version of Ansible that we now don't use. The syntax and logic of how including handers works has changed, and I think that's the issue. 

I'll fix it up soon, probably along with this: #383

Matt-Yorkley:
Closes #394 

The `rollback.yml` playbook was not working at all, for various reasons. I don't think it's been used for a long time, and maybe was only used in development? 

It works now! :tada:

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -1,5 +1,6 @@
 ---
 rollback_path: "{{ app_path }}/rollback"
 rollback_sql_path: "{{ app_path }}/rollback.sql.gz"
+rollback_tmp_path: "/var/lib/postgresql/rollback.sql.gz"
 
 bugsnag_key: none
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -106,7 +106,7 @@
 - name: create a repo backup version
   command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }}
   args:
-    removes: "{{ rollback_path }}"
+    creates: "{{ releases_path }}/{{ timestamp }}"
   tags: skip_ansible_lint
 
 - name: check for old release files
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -128,10 +128,9 @@
 #-------------------------
 # Move new code into place
 
-- name: move build to current
-  command: mv {{ build_path }} {{ current_path }} removes={{ build_path }}
-  become: yes
-  become_user: "{{ unicorn_user }}"
+- name: copy build to current
+  command: cp -TLr {{ build_path }} {{ current_path }}
+  changed_when: True
 
 #--------------------
 # Update the database
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -89,9 +89,19 @@
 # Create a rollback release
 
 - name: create a database rollback version
-  command: bash -c "set -o pipefail && pg_dump -h '{{ db_host }}' -U '{{ db_user }}' '{{ db }}' | gzip > '{{ rollback_sql_path }}'"
+  postgresql_db:
+    name: "{{ db }}"
+    state: dump
+    target: "{{ rollback_tmp_path }}"
+  become: yes
+  become_user: postgres
+
+- name: move database dump
+  command: mv {{ rollback_tmp_path }} {{ rollback_sql_path }}
+  become: yes
+  become_user: root
   args:
-    creates: "{{ rollback_sql_path }}"
+    removes: "{{ rollback_tmp_path }}"
 
 - name: create a database backup version
   command: cp {{ rollback_sql_path }} {{ releases_path }}/{{ timestamp }}.sql.gz
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -1,16 +1,13 @@
 --- # Deploy
 
 # Remove old rollbacks so they aren't used accidentally after errors.
-- name: remove the old rollback
+- name: remove the old rollbacks
   file:
-    dest={{ rollback_path }}
-    state=absent
-  tags: clone
-
-- name: remove the old rollback database
-  file:
-    dest={{ rollback_sql_path }}
-    state=absent
+    dest: "{{ item }}"
+    state: absent
+  with_items:
+    - "{{ rollback_path }}"
+    - "{{ rollback_sql_path }}"
   tags: clone
 
 #----------------
@@ -18,8 +15,8 @@
 
 - name: remove the old build path
   file:
-    dest={{ build_path }}
-    state=absent
+    dest: "{{ build_path }}"
+    state: absent
   tags: clone
 
 - name: build codebase from git branch
@@ -101,11 +98,15 @@
   tags: skip_ansible_lint
 
 - name: create a repo rollback version
-  command: mv {{ current_path }} {{ rollback_path }} removes={{ current_path }}
+  command: mv {{ current_path }} {{ rollback_path }}
+  args:
+    removes: "{{ current_path }}"
   tags: skip_ansible_lint
 
 - name: create a repo backup version
-  command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }} removes={{ rollback_path }}
+  command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }}
+  args:
+    removes: "{{ rollback_path }}"
   tags: skip_ansible_lint
 
 - name: check for old release files
--- a/roles/rollback/tasks/main.yml
+++ b/roles/rollback/tasks/main.yml
@@ -10,6 +10,29 @@
   args:
     creates: "{{ current_path }}"
 
+- name: close active db connections
+  shell: psql -c "REVOKE CONNECT ON DATABASE {{ db }} FROM public; SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='{{ db }}' AND pid <> pg_backend_pid();"
+  become: true
+  become_user: postgres
+
+- name: drop database
+  postgresql_db:
+    name: "{{ db }}"
+    state: absent
+  become: true
+  become_user: postgres
+
+- name: recreate database
+  postgresql_db:
+    name: "{{ db }}"
+    owner: "{{ db_user }}"
+    encoding: "UTF-8"
+    lc_ctype: "{{ postgres_encoding }}"
+    lc_collate: "{{ postgres_encoding }}"
+    state: present
+  become: true
+  become_user: postgres
+
 - name: restore database from backup
   postgresql_db:
     name: "{{ db }}"
--- a/roles/rollback/tasks/main.yml
+++ b/roles/rollback/tasks/main.yml
@@ -10,8 +10,14 @@
   args:
     creates: "{{ current_path }}"
 
-- name: copy rollback db to live db
-  shell: gunzip -c {{ rollback_sql_path }} | psql -h {{ db_host }} -U {{ db_user }} {{ db }}
+- name: restore database from backup
+  postgresql_db:
+    name: "{{ db }}"
+    owner: "{{ db_user }}"
+    state: restore
+    target: "{{ rollback_sql_path }}"
+  become: true
+  become_user: postgres
   notify:
     - precompile assets
     - restart unicorn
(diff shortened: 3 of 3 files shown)


Matt-Yorkley:
This might be needed during the v2 rollout...

roles/deploy/tasks/main.yml
```diff
@@ -100,31 +97,39 @@
...
 
 - name: create a database rollback version
-  command: bash -c "set -o pipefail && pg_dump -h '{{ db_host }}' -U '{{ db_user }}' '{{ db }}' | gzip > '{{ rollback_sql_path }}'"
-  args:
-    creates: "{{ rollback_sql_path }}"
+  postgresql_db:
+    name: "{{ db }}"
+    state: dump
+    target: /tmp/rollback.sql.gz
+  become: yes
```
sauloperez: Feels great to rely on core modules ❤️

roles/deploy/tasks/main.yml
```diff
@@ -100,31 +97,39 @@
...
+  become_user: postgres
   when: table_exists.stderr.find('does not exist') == -1
 
+- name: copy database dump
+  copy:
+    src: /tmp/rollback.sql.qz
+    dest: "{{ rollback_sql_path }}"
+    remote_src: yes
```
sauloperez: Why is this step needed? Can't we use this path as the dump's target?

roles/deploy/tasks/main.yml
```diff
@@ -100,31 +97,39 @@
...
 - name: create a repo rollback version
-  command: mv {{ current_path }} {{ rollback_path }} removes={{ current_path }}
+  command: mv {{ current_path }} {{ rollback_path }}
+  args:
+    removes: "{{ current_path }}"
   tags: skip_ansible_lint
 
 - name: create a repo backup version
-  command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }} removes={{ rollback_path }}
+  command: cp -r {{ rollback_path }} {{ releases_path }}/{{ timestamp }}
```
sauloperez: I don't really understand the point of `rollback_path`. Isn't enough with a releases folder containing the last N releases and `releases/current` being a symlink pointing to the last one? Then, rolling back is just a matter of making that symlink point to an older release. Not to mention how fast that is would be.

roles/deploy/tasks/main.yml
```diff
@@ -146,10 +151,9 @@
 #-------------------------
 # Move new code into place
 
-- name: move build to current
-  command: mv {{ build_path }} {{ current_path }} removes={{ build_path }}
-  become: yes
-  become_user: "{{ unicorn_user }}"
+- name: copy build to current
+  command: cp -TLr {{ build_path }} {{ current_path }}
```
sauloperez: so basically, our `releases/current/` is not a symlink but a copy

roles/rollback/tasks/main.yml
```diff
@@ -2,14 +2,45 @@
...
 - name: copy rollback repo to current path
-  command: cp -r {{ rollback_path }} {{ current_path }} creates={{ current_path }}
+  command: cp -r {{ rollback_path }} {{ current_path }}
+  args:
+    creates: "{{ current_path }}"
 
-- name: copy rollback db to live db
-  shell: gunzip -c {{ rollback_sql_path }} | psql -h {{ db_host }} -U {{ db_user }} {{ db }}
+- name: close active db connections
+  shell: psql -c "REVOKE CONNECT ON DATABASE {{ db }} FROM public; SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='{{ db }}' AND pid <> pg_backend_pid();"
```
sauloperez: AFAIK pg_stat_activity is not enabled at all. I'd like to whether this works at all.

roles/rollback/tasks/main.yml
```diff
@@ -2,14 +2,45 @@
...
+- name: recreate database
+  postgresql_db:
+    name: "{{ db }}"
+    owner: "{{ db_user }}"
+    encoding: "UTF-8"
+    lc_ctype: "en_US.utf8"
+    lc_collate: "en_US.utf8"
+    state: present
+  become: true
+  become_user: postgres
```
sauloperez: Might be worth checking whether our instances use these default `lc_ctype` and `lc_collate` locales or the corresponding local ones such as `es_ES.utf8`.

roles/deploy/tasks/main.yml
Matt-Yorkley: This was a workaround, as we have to use the `postgres` user to dump the db in the preceding step, and that user doesn't have the necessary folder permissions. I agree it seems a bit more convoluted, but it didn't work otherwise. 

I could take another go at it...

sauloperez: Fair enough. Let's iterate

roles/deploy/tasks/main.yml
Matt-Yorkley: Yeah, I was getting weird results when looking at the files copied into `/rollback`. I think it was because a lot of the files in `/build` are symlinks. Strangely, about 50% of the time the files were not all there. Forcing the copy to dereference symlinked files and create real copies fixed the issue and made the outcomes repeatable, with all the files intact each time.

roles/rollback/tasks/main.yml
Matt-Yorkley: I tested it, and it seems to work. I tried a rollback on a Vagrant box with active connections to the app. It failed before I added this, and now it succeeds.

Matt-Yorkley: I think the pg_stat_activiity table does some minimal things by default like track process ids, and can be configured to record additional monitoring data?

roles/rollback/tasks/main.yml
Matt-Yorkley: We force the use of `en_US.utf8` here: https://github.com/openfoodfoundation/ofn-install/blob/98d35374ab99620ef95cc9cec15e470a65156219/roles/dbserver/files/fix_pg_encoding.sh

Matt-Yorkley:
> I'm not entirely sure I would always execute this playbook when having to revert the previous code. In my little experience restoring a DB backup is normally unnecessary, as you just want a code rollback, and doing it by default feels a bit scary. I'd just use this playbook when needing a "full" rollback like we might for v2.

Agreed! This is a pretty destructive operation and should only be used in emergencies. Maybe we should add a comment to explain that? It fully reverts the code and the database to the state before the last deploy.

roles/deploy/tasks/main.yml
Matt-Yorkley: If you have any insights on this one, they would be appreciated!

Matt-Yorkley:
At least an emergency rollback playbook that actually works is better than one that doesn't. Do you think we need more nuanced alternative? Like a soft-rollback type option for less serious cases?

roles/rollback/tasks/main.yml
sauloperez: Must be that. I'll give it a look.

roles/rollback/tasks/main.yml
sauloperez: even with that, I wouldn't give it for granted because I don't really know if we created the DBs with `bundle exec rake db:create`. Could that be forcing diferent values for these two config settings?

sauloperez:
Definitely, I want us to describe that this implements a "full" rollback and maybe the "soft-rollback" can be just a new deploy. I recall hearing a concept like rollforward or something alike in the GitOps tech talk I attended to recently.

roles/deploy/tasks/main.yml
sauloperez: @enricostano you might have a better idea as you played with [ansistrano/deploy](https://github.com/ansistrano/deploy) in the past.

sauloperez: @Matt-Yorkley worth checking https://github.com/ansistrano/rollback

Matt-Yorkley:
Normally I would just do another deploy, but if the database has been migrated that wouldn't remove those changes, would it?

sauloperez:
Exactly. That's the use case this would cover but even then a `db:rollback` might be a better idea. I don't have a clear answer. I guess it depends on how deep the change in DB was.

Matt-Yorkley:
> Exactly. That's the use case this would cover but even then a `db:rollback` might be a better idea. I don't have a clear answer. I guess it depends on how deep the change in DB was.

By default `db:rollback` will only roll back a single migration, but we could have added multiple migrations in the previous deployment. So for that to work I think we would need to figure out how many new migrations ran during the last deploy...? Then we could do `db:rollback STEP=3` for example. I'm not sure how we could calculate that accurately... :thought_balloon:

sauloperez:
In any case that is something, I might manually, and just when we know one or two migrations got in. I don't think it's worth establishing a process for that now. It feels too early to me.

What if we just add another playbook in a separate PR that restores a previous release? That's definitely faster than preparing a new patch release and deploying it. We still need to improve that process.

What do you think?

Matt-Yorkley:
Lets have a call! :phone:

roles/rollback/tasks/main.yml
Matt-Yorkley: I think our existing hack overrides the default template, so any databases created with `db:create` will pick up those settings from `template0`.

roles/deploy/tasks/main.yml
mkllnk: I see. That also means that the backup will be visible to other system users. I don't think it's a big problem with our current setups but could be a security issue when sharing the server with other apps and users. Can we store it in the postgres user's home directory instead of tmp and then move it as root?

Matt-Yorkley: I've changed this to dump the db to the default postgres user's home folder, and used `mv` as root immediately after, so the temporary file is removed.

roles/rollback/tasks/main.yml
sauloperez: Nice! `pg_stat_activity` seems to be enabled by default. 

```
openfoodnetwork=> SELECT * FROM pg_stat_activity;
(...)
datid            | 16386
datname          | openfoodnetwork
pid              | 32597
usesysid         | 16384
usename          | ofn_user
application_name | delayed_job
client_addr      | ::1
client_hostname  | 
client_port      | 38148
backend_start    | 2019-04-15 12:40:28.499155+00
xact_start       | 
query_start      | 2019-04-17 13:51:37.313753+00
state_change     | 2019-04-17 13:51:37.313997+00
waiting          | f
state            | idle
backend_xid      | 
backend_xmin     | 
query            | UPDATE "delayed_jobs" SET locked_at = '2019-04-17 13:51:37.311119', locked_by = 'delayed_job host:scw-ecce22 pid:3
2594' WHERE id IN (SELECT  id FROM "delayed_jobs"  WHERE ((run_at <= '2019-04-17 13:51:37.308423' AND (locked_at IS NULL OR locked_at
 < '2019-04-17 13:36:37.308501') OR locked_by = 'delayed_job host:scw-ecce22 pid:32594') AND failed_at IS NULL) ORDER BY priority ASC
, run_at ASC LIMIT 1 FOR UPDATE) RETURNING *
```

roles/rollback/tasks/main.yml
Matt-Yorkley: I refactored this so the value is stored in a global variable `{{ postgres_encoding }}`, to make it a bit cleaner and so we could potentially override it in the future for different instances.

I also checked around various production databases to check the encoding was the same everywhere.

roles/deploy/defaults/main.yml
```diff
@@ -1,5 +1,6 @@
 ---
 rollback_path: "{{ app_path }}/rollback"
 rollback_sql_path: "{{ app_path }}/rollback.sql.gz"
+rollback_tmp_path: "/var/lib/postgresql/rollback.sql.gz"
```
mkllnk: This is much better. But it's still a world-readable directory. Even though the file is only there for a short time, you can watch the directory with inotify and read the file as soon as it's created.

What do you think about using `mktemp` which creates a private file. `root` will still be able to move it. :thinking: Looking at our current setup, `/home/openfoodnetwork` is actually world-readable as well. So you are not making anything worse here.

sauloperez: Maybe we can just open a separate issue and deal with it in a separate PR? your call

Matt-Yorkley:
i just fixed a merge conflict (moving to shared handlers) and retested this. The odd behaviour I saw previously with files not being copied correctly to the `/rollback` path is back. It needs more investigating before this is merged.

PR #400, opened 2019-04-16
Postgres installation

### LINKED ISSUE #393: Manage Postgres installation with a standardised role
We should switch to managing our installation of Postgres with a standardised role, like this one: https://github.com/geerlingguy/ansible-role-postgresql

We also need to specify a postgres version as well, as currently it will install whatever the OS's repos feel like, which is not a great idea for stability and reproducing environments.

We can do these two things at the same time, and our playbooks will get tidier.

Matt-Yorkley:
Closes #393 

This PR brings a big change to the way we handle Postgres installation, switching to use of an open source role: https://galaxy.ansible.com/geerlingguy/postgresql. This enables us to tidy things up, manage Postgres better, and should improve the situation when it comes to upgrading to Bionic Beaver.

:warning: Requires an update of the community roles before provisioning.


Matt-Yorkley:
I've opted to remove this code as well:

https://github.com/openfoodfoundation/ofn-install/blob/96badf275642293086baf5e04bd88e863b305b81/roles/dbserver/tasks/main.yml#L30-L36

https://github.com/openfoodfoundation/ofn-install/blob/96badf275642293086baf5e04bd88e863b305b81/roles/dbserver/files/fix_pg_encoding.sh#L1-L14

The use of this plugin means we now create the database outside of Rails. It's not constrained by the issue that the above code was designed to fix, and provisioning a fresh box with this PR works fine without the above code. 

The only issue would be if for some reason we wanted to drop and recreate a database from within Rails it wouldn't have our specified encoding defaults, but I don't think we ever any need to do that?

:fire:

roles/dbserver/tasks/main.yml
```diff
@@ -10,11 +10,3 @@
     mode: 0600
   become: yes
   become_user: "{{ unicorn_user }}"
-
-# Rails expects unicode databases. See the script for more details.
-- name: fix postgres db encoding problem
```
mkllnk: This means we can't use `rake db:reset` on staging servers?

I ran into this problem only once. It usually doesn't happen with English data but the application can crash when using other locales and the database is not UTF8.

I'm wondering if using template0 in database.yml could help: https://stackoverflow.com/a/16737073/3377535

enricostano:
> The only issue would be if for some reason we wanted to drop and recreate a database from within Rails (using db:create) it wouldn't have our specified encoding defaults, but I don't think we ever have any need to do that?

@Matt-Yorkley the only time I need this is on a staging machine when I want to wipe the DB and start from scratch again (pretty common IMHO).

roles/dbserver/tasks/main.yml
Matt-Yorkley: Ah, I forgot about `db:reset` on staging :see_no_evil: . Ok, I the template hack will have to go back in for sure, but maybe we can do it with Ansible in a clearer way.

Matt-Yorkley:
Okay, I moved the encoding hack into a reusable set of tasks that handle the job a bit better, and run conditionally.

playbooks/provision.yml
```diff
@@ -54,6 +54,8 @@
       tags: app
 
     - role: dbserver # Set up database server and user for the app.
+      become: yes
+      become_user: root
```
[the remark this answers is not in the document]
Matt-Yorkley: Yes! It depends on the context though, it can change. I encountered some weird issues with fatal permissions errors when testing this PR, and this fixed them. I didn't want to spend too much time investigating it, as it doesn't hurt.

Matt-Yorkley:
I've tested this thoroughly in Vagrant as well as on UK staging. Also double-checked the encoding values on all production servers so there's no surprises. :+1:

roles/deploy/tasks/main.yml
```diff
@@ -167,7 +148,7 @@
   command: bash -lc "bundle exec rake db:schema:load RAILS_ENV={{ rails_env }}"
```
mkllnk: I'm wondering if we should separate *initial deploy* from following *deploy*s. One mistake in the `when` clause and we wipe out production data. But I guess that's not the Ansible way of doing things. We just have to test things well and you seem to be doing that. :-)

Matt-Yorkley: Yes, I was planning on doing something like that after the next few bits are merged. The database creation is now outside of the deploy script's responsibilities, and that's a big step forward to being able to untangle things.

PR #401, opened 2019-04-16
Refactor handlers

### LINKED ISSUE #383: Refactor use of handlers
This was prompted here: https://github.com/openfoodfoundation/ofn-install/pull/381#pullrequestreview-223762467 and is probably long-overdue.

We have handlers that are used across different playbooks and roles, but no definitive way to DRY and reuse them.

We also have some terrible use of handlers in the deploy role.

Matt-Yorkley:
Closes #383 

- Creates a reusable `shared_handlers` role that can be used to store handlers that are used in multiple playbooks
- Tidies up junk tasks, like this one: https://github.com/openfoodfoundation/ofn-install/commit/a8b4ba9513bca86067fb59b70e47c938c27a33a6 :fire:
- Moves some handlers that shouldn't be handlers into tasks in `deploy.yml`


roles/deploy/handlers/main.yml
```diff
@@ -18,25 +18,6 @@
   become: yes
   become_user: root
 
-- name: restart unicorn
```
mkllnk: I see that this is removed but I don't see any `notify` of the shared handler.

roles/deploy/tasks/main.yml
```diff
@@ -196,14 +195,15 @@
...
 
-# tag to trigger a unicorn restart
-- name: reset
-  # needs to do something here
-  command: echo 1
-  tags:
-    - reset
-    - skip_ansible_lint
-  notify:
-    - restart unicorn
```
mkllnk: I don't see a replacement for notifying the unicorn handler.

roles/shared_handlers/handlers/main.yml
```diff
@@ -1,7 +1,26 @@
...
+  become: yes
+  become_user: "{{ unicorn_user }}"
+
 - name: restart unicorn
   service:
     name: unicorn_{{ app }}
     state: restarted
   become: yes
   become_user: root
+  notify: restart nginx
```
mkllnk: I believe that nginx only needs to be restarted if the nginx config changes. Is that wrong?

roles/shared_handlers/handlers/main.yml
```diff
@@ -0,0 +1,26 @@
+# This role holds reusable handlers that can be included in multiple playbooks to keep things DRY
+- name: precompile assets
```
mkllnk: I'm wondering if this should be a task as well.

roles/shared_handlers/handlers/main.yml
Matt-Yorkley: I was thinking that too! I left this in because it's how the current handlers work, but I agree that we can probably cut it.

roles/deploy/tasks/main.yml
Matt-Yorkley: There are multiple `notify`s to the same handler in `deploy.yml`. It's still there!

roles/shared_handlers/handlers/main.yml
sauloperez: I would just create a separate issue. Agree. We almost never change nginx config so if we don't restart it we could already have zero-downtime deployments :tada:

Matt-Yorkley: Removed! :fire: :smile:

Matt-Yorkley:
> we could already have zero-downtime deployments

@sauloperez I think the unicorn `restart` we use is still a slightly harsh one, and not entirely zero-downtime. Instead we could make sure the service is running, then use `reload`, which according to the docs: "reloads config file and gracefully restarts all workers".

Maybe that's for another PR...

sauloperez:
I thought that was done in one of the previous PRs already. At least I remember reading about it.

Matt-Yorkley:
Yes, it's been enabled and used in the provision playbook for when we update unicorn configs, but we don't use it in the deploy playbook.

roles/shared_handlers/handlers/main.yml
mkllnk: @sauloperez Don't we have zero-downtime deployments? What is currently causing downtime?

roles/shared_handlers/handlers/main.yml
mkllnk: Interesting. I thought a rollback is restoring the old files including the assets. So it doesn't need to compile assets... Anyway, I'm happy to keep it as handler. I just like your approach so much. :-)

mkllnk:
Interesting point @Matt-Yorkley. The old init script treated `reload` and `restart` the same:

https://github.com/openfoodfoundation/ofn-install/blob/a769b5c5aaded867e2748abce23daca7ac594c77/roles/webserver/templates/unicorn_init#L52-L55

But the new service doesn't define restart. So I guess that systemd is automatically doing a stop and start.

https://github.com/openfoodfoundation/ofn-install/blob/84fb28360b7b8d17bb860c293ba27690ff8d24c4/roles/webserver/templates/unicorn.service.j2#L19-L21

That's good, we can actually choose to do a restart which is necessary for some gems that load really early. new_relic was one of them, bugsnag and skylight could be affected as well. So it would be nice if we could choose on the command line if we wanted to use reload or restart during the deploy, depending on which gems were updated. Not sure how to detect that automatically. Maybe during the bundle execution?

roles/shared_handlers/handlers/main.yml
Matt-Yorkley: That's a good point! The original rollback role used `notify: precompile assets` so I didn't remove it, but maybe it can be striped out. I think updating unicorn should be enough.

Matt-Yorkley:
> The old init script treated reload and restart the same

That script is now defunct, as `initd` is deprecated and we're not using it, so I guess we can remove it from ofn-install.

I read through the unicorn docs recently and there's various improvements we can make to our setup. I'd like to do that at some point.

PR #402, opened 2019-04-16
Use native modules for bundler tasks

Matt-Yorkley:
Removes a bit more dependence on bash scripts and `bash -lc ...`  in favour of native Ansible modules.

:fire::fire:

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -4,3 +4,12 @@ rollback_sql_path: "{{ app_path }}/rollback.sql.gz"
 rollback_tmp_path: "/var/lib/postgresql/rollback.sql.gz"
 
 bugsnag_key: none
+
+initialize_rbenv:
+  RBENV_ROOT: "/home/{{ unicorn_user }}/.rbenv"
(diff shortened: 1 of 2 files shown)


Matt-Yorkley:
> I'm not convinced that replacing one shell task with three others is an improvement. But I'm also biased because I wrote that shell script. ;-) When I saw the fire signs in the PR description I was hoping for less code.

I was mostly trying to see where we could use more native modules and less `bash -lc ...` in our playbooks. The original script is still useful in dev.

`grep -m 1 -A 1 -x -F "BUNDLED WITH" Gemfile.lock | tail -n 1 | tr -d '[:space:]'` is a fantastic line! We couldn't do without it, but we can use it with `register` and `changed_when: False`, and then handle the real tasks of `gem bundler install ...` and `bundle install` with the `gem` and `bundler` modules.

Matt-Yorkley:
I've retested this and there's some issues, possibly with rbenv's rehash system. Lets leave it for now.

sauloperez:
Agree. This is post v2. We have more urgent matters for that release to be rolled out smoothly.

luisramos0:
moving to in dev.

mkllnk:
Is it a good time to pick this up again, @Matt-Yorkley? Or should we close it?

mkllnk:
Things have changed and this PR doesn't apply any more. Closing.

PR #403, opened 2019-04-18
Zapier makes you happier

### LINKED ISSUE #230: Enable DB integrations such as Zapier
In the UK we've started using Zapier, which is a service that integrates with the database and allows for a wide range of extra functionality, like creating custom reports for enterprises, exporting order data to accounting software, or sending email alerts when something gets added.

It involves creating a postgresql user with restricted permissions, and enabling the service to connect directly to the DB (read-only).

It looks like this could be useful for multiple instances, and since it involves server configuration we should probably add an optional role for it to ofn-install and decide on the best way to do it. There are some security issues to discuss as well.

Matt-Yorkley:
Closes #230 

This one's a beauty...

- Enables integrations with services like Zapier and Metabase
- Enables extensible postgres config management per-host via secrets files
- Integrations can be easily added or removed by updating the variables in secrets.yml and running the new playbook
- Multiple integrations can be added at once. Instance managers can use Metabase and Zapier at the same time to Meta their Zaps (if they feel like it)
- Existing manually-configured Zapier integrations in France and UK will be cleanly updated (and actually improved slightly)

Tested on Vagrant and UK Staging. I also did a quick audit of all the existing `pg_hba.conf` and `postgresql.conf` files across our various production servers to check we didn't have any unexpected manual configurations.

ZAP!


Matt-Yorkley:
Tested on UK Production as well :heavy_check_mark:

roles/integrations/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
+---
```
sauloperez: Might a bit easier to understand what is about if we name it `roles/db_integrations/` instead?

Matt-Yorkley:
An example secrets file entry for enabling both Zapier and Datadog would look like this:

```
db_integrations:
  - { user: zapier, state: present, password: incredibly_strong_password }
  - { user: datadog, state: present, password: incredibly_strong_password }

postgres_listen_addresses:
  - '*'

custom_hba_entries:
  - { type: hostssl, database: "{{ db }}", user: datadog, address: '<datadog's ip>/32', auth_method: md5 }
```

inventory/group_vars/all.yml
```diff
@@ -56,6 +56,31 @@ db_user: ofn_user
 db_host: localhost
 postgres_encoding: en_US.utf8
 
+#----------------------------------------------------------------------
+# Postgresql variables
+
+# These include the default values for postgres, and some empty (extendable) values.
+# Entries marked custom_* can be defined in secrets.yml files, and will be appended
```
sauloperez: we need to update the doc replacing secrets.yml with host_vars.

Matt-Yorkley: I've updated it so it's more generic. I think this'll need a wiki page anyway, and it'll only be configured by core devs anyway, so we shouldn't worry too much.

sauloperez:
> I think this'll need a wiki page anyway

You read my mind.

PR #405, opened 2019-04-19
Australia update

mkllnk:
Some update configuration for the Australian production server. The server has been partly provisioned and its current status is documented at https://github.com/openfoodfoundation/ofn-install/wiki/Australia.

There will be more updates while progressing on this task. But since I'm going on holiday for two weeks I didn't want to change too much right now.

### DIFF
--- a/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
@@ -1,8 +1,9 @@
 ---
 
-domain: prod2.openfoodnetwork.org.au
+domain: openfoodnetwork.org.au
 rails_env: production
 
 admin_email: maikel@openfoodnetwork.org.au
 
 unicorn_timeout: 120
+unicorn_workers: 6
--- a/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
@@ -5,5 +5,19 @@ rails_env: production
 
 admin_email: maikel@openfoodnetwork.org.au
 
+mail_domain: openfoodnetwork.org
+
+# Optional list of subdomains to register in letsencrypt certificate
+# Defaults to {{ domain }} if list is undefined
+#
+#certbot_domains:
+#  - openfoodnetwork.org.au
+#  - www.openfoodnetwork.org.au
+#  - prod2.openfoodnetwork.org.au
+
+# Size in bytes. You can also use units like 1G, 512MiB or 1000KB. See: `man fallocate`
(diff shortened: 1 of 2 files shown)


inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
```diff
@@ -1,8 +1,23 @@
 ---
 
-domain: prod2.openfoodnetwork.org.au
+domain: openfoodnetwork.org.au
```
sauloperez: I would also rename the folder for the sake of consistency.

inventory/hosts
```diff
@@ -15,7 +15,6 @@ lxc
 # Australia
 
 [au-prod]
-www.openfoodnetwork.org.au
 prod2.openfoodnetwork.org.au
```
sauloperez: here too

inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
luisramos0: agree, the folder name needs to be updated.
 there's another folder www.openfoodnetwork.org.au...

luisramos0:
moving to in dev until @mkllnk is back to this

inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
mkllnk: Is there any other benefit than consistency? I did this intentionally because I thought the host is the actual computer we are installing to which is different to the application it is hosting and under which domain names the application is reachable.

When I added prod2, the old host was still alive and I didn't want to touch it's configuration. And when it comes to installing a new host, I would name it prod3. Initially, it would have the domain name prod3.openfoodnetwork.org.au because we can't get a certificate for a domain that points to another server. When everything works, I would change the domain, copy the SSL certificates, database and files. In case anything goes wrong, I may still want the old host to be available.

Do you think that is too complicated and I should just use different branches for dealing with old and new hosts?

luisramos0: I am not sure what's my opinion here.  I trust you! :-)

From an external perspective, it would be better to keep it simple and have one single domain to connect to au live server, either www.openfoodnetwork.org.au or openfoodnetwork.org.au would be better.

mkllnk: Is it okay if I review this again when moving the server next time? Are you happy to confirm this PR @luisramos0?

luisramos0: yes yes :+1:

PR #406, opened 2019-04-19
Remove obsolete Git post-receive hook

mkllnk:
We are deploying with Ansible now. The hook is also outdated and the current versions in Canada and Australia look different. If we decide to replace this hook we can do that from scratch. At the moment it's just unused code that delays deployments by creating an unused symlink.


sauloperez:
And then where is that different version of the hook kept @mkllnk ?

mkllnk:
@sauloperez 
> And then where is that different version of the hook kept

It's currently only on the servers. So if we lose a server, we lose this configuration. But then we can just provision with ofn-install and don't need it any more.

PR #415, opened 2019-04-22
Separate deploy.yml into two halves

Matt-Yorkley:
This is the last step of refactoring of `deploy.yml` I've been wanting to do.

Now that we've removed some of the database creation tasks from `deploy.yml` and removed a lot of conditionals, we can separate all the remaining tasks in the `deploy.yml` role into two clear groups of tasks: handling rollbacks and deploying the codebase. The tasks were previously all mixed together, but now they can be extracted into separate files, leaving `roles/deploy/tasks/main.yml` looking like this:

Screenshot-2019-04-22-19-45:17

The tasks have moved into sibling files (but still inside the deploy role), like so:
```
roles/
  deploy/
    tasks/
      main.yml
      rollbacks.yml
      deploy.yml
```

Also includes: removing some unnecessary code, tidying up some more syntax, removing some instances of `skip_ansible_lint`, and a small bugfix for an error in the creation of release files on a fresh install when a path doesn't exist yet.


roles/deploy/tasks/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+  tags:
+    - rake
+    - skip_ansible_lint
+  notify:
+    - precompile assets
+    - restart unicorn
+
+- name: seed database
+  # We run a shell script that passes the default email and password to rake with an EOF block, so we don't hang on the prompts.
+  command: bash -lc "{{ config_path }}/seed.sh RAILS_ENV={{ rails_env }}"
```
sauloperez: Aren't we seeding the DB always, including production?

Matt-Yorkley: Yes, and I think the rake task accepts ENV vars, so we don't need the script.

sauloperez:
I agree with you @enricostano . This can serve as a gradual transition towards that.

roles/deploy/tasks/deploy.yml
Matt-Yorkley: I'll do this in a separate PR :+1:

Issue #414, opened 2019-04-21
Set up CI for ofn-install


Matt-Yorkley:
I think this is long overdue. We don't have to do it all at once, but here's what I would suggest:

- [x]  Add a basic CI build that runs though `site.yml` #418 
- [x]  Test other core playbooks #421 
- [x]  Add some actual tests to the process! We can do it really easily with Ansible #421 
- [ ] Use the "test" certificates feature with Certbot so we can check it's working (it broke recently: https://github.com/openfoodfoundation/ofn-install/issues/392) 

Maybe we can take a look at this at the UK devhack?

PR #416, opened 2019-04-23
Ensure correct exit code on deployment

Matt-Yorkley:
We hit this issue the other day when Rachel was testing. It turns out the rescue block we use in deploy.yml to send a failed deployment alert to Slack interferes with the exit code of the playbook because of an Ansible bug. This fixes it.

See Ansible issue for more details: https://github.com/ansible/ansible/issues/48344


Matt-Yorkley:
Also needed for correct build status with CI builds #414

PR #417, opened 2019-04-23
Clean up db:seed task

Matt-Yorkley:
Follow-up from: https://github.com/openfoodfoundation/ofn-install/pull/415#discussion_r277547914

We don't need tasks in multiple playbooks to get this job done :+1: 

The code that handles this rake task is in the `spree_auth_devise` gem, and it either takes input from `ENV` vars, or prompts for user input (which then hangs in Ansible).


Matt-Yorkley:
Tested :heavy_check_mark:

PR #418, opened 2019-04-23
Travis CI build

Matt-Yorkley:
Part one of #414 

This introduces a basic CI build, where we do a real run-through of `site.yml`.

Given that it's the 50th anniversary of the moon landings this year, I think I should point out that this is: one small step for `travis.yml`, one giant leap for `ofn-install`...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,21 +7,28 @@ python: "2.7"
 services:
   - postgresql
 
+env:
+  - task: LINT
+  - task: BUILD
+
 before_install:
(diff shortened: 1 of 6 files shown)


sauloperez:
It would be awesome if we could run these two jobs sequentially. Don't BUILD until LINT passes. It can be done in a separate PR though..

Matt-Yorkley:
I think there's a way to exit the build early if one of the jobs fails, I'll look into it. Lets iterate! 

I've got big plans for this already... :smile:

PR #419, opened 2019-04-23
Fix nginx config check error on Vagrant

Matt-Yorkley:
An error popped up recently from the `jdauphant.nginx : check nginx configuration` handler when runing `provision.yml` on Vagrant. I think the role was updated recently.

This fixes it. :+1:


Matt-Yorkley:
> So we don't serve HTTPS with Vagrant?

Nope, it's a local VM which can be destroyed and recreated often.

PR #421, opened 2019-04-24
Test!

### LINKED ISSUE #414: Set up CI for ofn-install

Matt-Yorkley:
Closes #414 

I've been experimenting with adding some basic tests for playbooks to our new CI build, and this is what I've got so far. I think we can add one or two tests for each playbook that definitively answer whether it's done it's job or not. So far I've added:

- A smoke test that checks the homepage is working after running `site.yml`
- A test to see if `deploy.yml` creates a rollback properly (it was broken before)
- Tests that makes sure the database and files have been reset in `rollback.yml`
- Tests that check `db_integrations.yml` updates postgres configs correctly.

With a few more tests, this could cut out the need for manual testing in VMs every time a change has been made! :rocket:


luisramos0:
@Matt-Yorkley is it in code review or still in wip? can you please move to "in dev" or remove the "pr-wip" label?

Matt-Yorkley:
So this PR is sort of an experiment / proposal. I didn't track my time for it, it was just something I put together on a free Sunday. 

I looked at Ansible testing frameworks but they mostly seem to be based on rigorous unit tests for people packaging individual roles for re-use by third parties, and I think that's not really what we need. They would also require everyone using them to learn a new DSL.

This imposes some arbitrary conventions in terms of the directory structure and a naming convention for the tests. There's a central file called `suite.yml` that runs the tests, which is partly dictated by the limitations of the way `import_playbook` works in Ansible. The other files are kept in folders named by the playbooks they apply to. I think if we run through the playbooks in a build, each task is effectively a test in itself, so we don't need to add much more. There have been a few cases in the last month though where I've discovered playbooks that were not failing but also not doing exactly what they should, and we hadn't realised.

So this PR is waiting for consensus on whether or not people think it's a good idea, and if people agree with the basic structure. If so we could add to it a bit more in the future.

Having tested and re-tested a lot of `ofn-install` changes manually I think it's awesome, but feel free to disagree! :smile:

sauloperez:
> I think if we run through the playbooks in a build, each task is effectively a test in itself, so we don't need to add much more. There have been a few cases in the last month though where I've discovered playbooks that were not failing but also not doing exactly what they should, and we hadn't realised.

IMO these are two key things here. We need to trust Ansible's declarative and idempotent approach but we do commit mistakes. We don't need to add as many tests as we have in the app.

Matt-Yorkley:
> We need to trust Ansible's declarative and idempotent approach but we do commit mistakes. We don't need to add as many tests as we have in the app.

Absolutely! :100:

mkllnk:
So we have two approvals but @Matt-Yorkley talked about consensus here. Whom would you like approve? @sauloperez?

tests/db_integrations/test.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: test
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+
+  tasks:
+    - name: test | db_integrations | expect postgresql.conf to be updated
+      command: grep -q "listen_addresses = '*'" {{ postgresql_config_path }}/postgresql.conf
```
kristinalim: I'm not sure how the PostgreSQL configuration playbook updates the config files. Don't we need to match the _start of the line_ here to make sure that the setting is not commented out?

tests/rollback/setup.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  tasks:
+    - name: setup | rollback | add testable entry to database
+      become: yes
+      become_user: postgres
+      command: |
+        psql {{ db }} -c "
+        UPDATE spree_preferences
+        SET value='rollback_test'
+        WHERE key='spree/app_configuration/privacy_policy_url';"
```
kristinalim: This row would initially not exist, right?

tests/db_integrations/test.yml
Matt-Yorkley: That might be an improvement, yes. It works currently because the initial setup of the test server applies the defaults, which don't contain the string `listen_addresses = '*'`.

tests/rollback/setup.yml
Matt-Yorkley: It's a standard spree preference, so the row will exist at this point, but we're setting a non-default value of `rollback_test`.

Matt-Yorkley: P.S. the build is green, so you can see the Ansible output already in the Semaphore build for this PR. :point_down:

tests/db_integrations/test.yml
Matt-Yorkley: We apply this non-default setting in the setup step for the test, in `tests/db_integrations/setup.yml` 
https://github.com/openfoodfoundation/ofn-install/pull/421/files#diff-2efee1d210c2ed1f315a9ea78be94230

Matt-Yorkley:
> So we have two approvals but @Matt-Yorkley talked about consensus here. Whom would you like approve? @sauloperez?

It's a big step in a new direction, and includes lots of choices about structure and implementation that I essentially just made up, so I thought it would be good to get some general agreement and see if people have any counter-proposals or complaints. :+1:

sauloperez:
I'm very happy to merge but before that I'd carefully document this new approach and the decisions behind it in a wiki page.

tests/rollback/setup.yml
kristinalim: Is this created somewhere in the setup file aside from `script/setup`? I asked because `script/setup` (running Rake task `ofn:sample_data`) does not create a `Spree::Preference` record for this.

Is this why the check passes?

```
failed_when: rollback_test.stdout == '1'
```

Sorry if I'm missing something supposedly obvious here.

Matt-Yorkley:
Wiki page added: https://github.com/openfoodfoundation/ofn-install/wiki/CI-Build

sauloperez:
what do you think @kristinalim ? shall we move forward?

tests/rollback/setup.yml
kristinalim: @Matt-Yorkley I have a pending small concern here. I believe the `Spree::Preference` here does not actually exist yet and the `failed_when` below is wrong (`stdout` output should be 1 if there is a match).

But already happy for this PR to be merged if that is not the case.

kristinalim: @Matt-Yorkley I have a pending small concern here. I believe the `Spree::Preference` here does not actually exist yet and the `failed_when` below is wrong (`stdout` output should be 1 if there is a match).

But already happy for this PR to be merged if that is not the case.

Matt-Yorkley: The `UPDATE` query passes in the build:

exists

It would fail if that record didn't exist, right?

Matt-Yorkley: This setup step takes place after a full provision and deploy. Those Spree preferences are stored in the database by that point.

kristinalim: @Matt-Yorkley A SQL `UPDATE` query does not fail if there is no record to update, and not all Spree preferences are inserted in the database when `script/setup` is run. The Spree preference for privacy policy URL is one the many that are not created - it is just un-set.

Matt-Yorkley: Ah you're right! I must have looked in my dev database instead of my ofn-install Vagrant database when choosing a key. I've changed it to an `INSERT` of a unique record.

Matt-Yorkley: We're still doing `failed_when: rollback_test.stdout == '1'` in the test though, as the record should _not_ exist after the rollback playbook has run.

kristinalim: That sounds right, @Matt-Yorkley! :slightly_smiling_face:

sauloperez:
:tada:

Matt-Yorkley:
:tada: Awesome, this will really help with #442

PR #422, opened 2019-04-25
Datadog config

### LINKED ISSUE #305: Notify deploys to Datadog
## Description

We need to know when deployments happen so that we can correlate issues (or not). While a mem and cpu usage spike is fine in a deploy it is not in other circumstances.

Visually knowing this information in all metrics graphs would allow correlating things at a glance.

## Proposed solution

Investigate what's the way to do this with Datadog and implement it. It shouldn't be hard.

Matt-Yorkley:
Closes #305 

This PR: 
- improves our current Datadog settings
- notifies Datadog on a successful deploy or provision
- enables collection of postgresql stats and logs through the Datadog agent
- also enables nginx stats and logging to Datadog

Screenshot-2019-04-25-13-21:57

:warning: Provisioning the new postgres stats and logging features to a server requires adding or updating these variables in `secrets.yml`:

```ruby
datadog_db_password: <secure_password_goes_here> # use $ openssl rand -hex 128

db_integrations:
  - { user: datadog, state: present, password: "{{ datadog_db_password }}" }
```
And if you havn't done it already, you'll also need to run `playbooks/db_integrations.yml` first to create the database user and grant them permissions.

### DIFF
--- a/roles/datadog/tasks/pg_stats.yml
+++ b/roles/datadog/tasks/pg_stats.yml
@@ -1,0 +1,53 @@
+---
+
+- name: enable postgres logging
+  set_fact:
+    datadog_checks:
+      postgres:
+        init_config:
+        instances:
+          - host: localhost
+            port: 5432
+            username: datadog
+            password: "{{ datadog_db_password }}"
+            dbname: "{{ db }}"
+            tags:
+              - "host:{{ inventory_hostname }}"
+            collect_activity_metrics: true
+        logs:
+          - type: file
+            path: /var/log/pg_log/postgresql-*.log
+            source: postgresql
+            sourcecategory: database
+            service: postgres
+  when: datadog_db_password is defined
+
+- name: check datadog db_integration exists
+  command: psql -tAc "SELECT 1 FROM pg_roles WHERE rolname='datadog'"
+  become: yes
+  become_user: postgres
+  register: integration_exists
+  changed_when: False
+
+- fail:
+    msg: "The database role `datadog` has not been created. Add it via the `db_integrations` playbook."
+  when: integration_exists.stdout != '1'
+
+- name: check datadog views
+  command: psql -c "SELECT * FROM pg_catalog.pg_views WHERE viewname = 'pg_stat_activity_dd';"
+  become: yes
+  become_user: postgres
+  register: stat_views_exist
+  changed_when: False
+
+- name: add pg_stat function and view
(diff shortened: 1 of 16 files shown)


Matt-Yorkley:
I've tested this extensively in Staging, and it's now live on UK production as well :+1:

roles/datadog/tasks/pg_stats.yml
```diff
@@ -0,0 +1,54 @@
...
+            port: 5432
+            username: datadog
+            password: "{{ datadog_db_password }}"
+            dbname: "{{ db }}"
+            tags:
+              - "host:{{ inventory_hostname }}"
+            collect_activity_metrics: true
+        logs:
+          - type: file
+            path: /var/log/pg_log/postgresql-*.log
```
sauloperez: Are we sure about that? As far as I read setting pg_log as the log destination stores them within the Postgres installation directory and not in /var/log.

sauloperez: Check `log_directory` in https://github.com/openfoodfoundation/ofn-install/pull/410/files

Matt-Yorkley: We havn't had any `stderr` output form UK prod yet, so we don't know... I'll check it again though

Matt-Yorkley: https://docs.datadoghq.com/integrations/postgres/#log-collection

Matt-Yorkley: You're right, so I've opted to go with the suggestions you quoted from here: https://logdna.com/postgresql-logging-guide/

Matt-Yorkley:
@sauloperez this turned out to be a lot more complicated than I originally thought. I decided to only apply the logging to servers that will use it, because the logs are massive and we will only be collecting them on 3 servers. I also switched the way we apply the logging config to use the `/conf.d` pattern, because it's difficult to cleanly change or remove things that have written lots of lines into the main config file in a haphazard way.

Matt-Yorkley:
We can get really nice Rails logs collected in Datadog as well and I think it would be really useful, but we need to discuss signing up for the log service to get log retention > 1 day. It's an extra $2-3 per million logs. If we do that we should probably look at raising the log level above `info` for postgres logs.

Matt-Yorkley:
I've reprovisioned UK prod :+1:

PR #423, opened 2019-04-25
Remove old monit tasks

### LINKED ISSUE #368: Remove old monit references
As mentioned in #332, we can now remove the `clean up monit` task and it's two associated handlers, as monit is now gone on current ofn-install servers and will not be added to any new ones if they get reinstalled from scratch.

#smallthings #techdebt

Matt-Yorkley:
Closes #368

Removing old tasks that cleaned up `monit`.


Matt-Yorkley:
> Can we check if this would still be needed for instance that don't use ofn-install yet? If they use monit we better wait a bit more.

Well here's Canada which needs to be completely reinstalled (so it's not a problem in that case), and maybe Aus production? I'm not sure.

sauloperez:
Yes, that's a good point. We will setup the server from scratch so we can remove the check.

PR #426, opened 2019-04-28
Add South Africa configs

Matt-Yorkley:
Add configs for the new/updated South Africa instance.

### DIFF
--- a/inventory/host_vars/ofn.kandu.co.za/config.yml
+++ b/inventory/host_vars/ofn.kandu.co.za/config.yml
@@ -1,6 +1,0 @@
----
-
-domain: ofn.kandu.co.za
-rails_env: production
-
-admin_email:
--- a/inventory/host_vars/www.openfoodnetwork.org.za/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.za/config.yml
@@ -1,0 +1,12 @@
+---
+domain: www.openfoodnetwork.org.za
+rails_env: production
+
+mail_domain: openfoodnetwork.org.za
+
+certbot_domains:
+  - www.openfoodnetwork.org.za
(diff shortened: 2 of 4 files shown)


inventory/host_vars/ofn.kandu.co.za/config.yml
```diff
@@ -1,6 +0,0 @@
----
-
-domain: ofn.kandu.co.za
```
mkllnk: That server still exists. What's happening with it? Is there a handover, is Kandu still alive or do we have two instances now?

inventory/host_vars/www.openfoodnetwork.org.za/config.yml
```diff
@@ -0,0 +1,12 @@
...
+rails_env: production
+
+mail_domain: openfoodnetwork.org.za
+
+certbot_domains:
+  - www.openfoodnetwork.org.za
+  - openfoodnetwork.org.za
+
+unicorn_workers: 2
+swapfile_size: 1G
```
mkllnk: I'm wondering if we should make this the standard now.

inventory/host_vars/ofn.kandu.co.za/config.yml
Matt-Yorkley: They're basically migrating to the new domain, and ofn-install isn't really used on the existing server anyway.

inventory/host_vars/www.openfoodnetwork.org.za/config.yml
Matt-Yorkley: I went for some conservative options as I'm not sure about their specs, but yes, we can probably set some bigger defaults soon for all servers.

sauloperez:
happy with the answers @mkllnk ?

PR #427, opened 2019-04-30
Allow passing a specific AWS region to S3 settings

sauloperez:
No need to rely on the default region Db2fog might specify. Needed for https://github.com/openfoodfoundation/openfoodnetwork/pull/3763.


sauloperez:
Tested in Katuma staging and working

sauloperez:
This needs to make it to a v1.30.1 release so we can get katuma production covered.

PR #428, opened 2019-05-02
Set unicorn_timeout for OFN DE to 120

kristinalim:
Set `unicorn_timeout` for OFN DE to 120.

## Description

This sets the Unicorn timeout for OFN DE to 120. 

It is not confirmed if a Unicorn timeout causes the HTTP 500 when using a large dataset for product import. But, considering that large data imports will be done, this change is appropriate anyway.

This is similar to #238 and #290.

Actually, I'm wondering if we should set the default `unicorn_timeout` to 120 for all instances. But that's outside the scope of this PR.

### DIFF
--- a/inventory/host_vars/openfoodnetwork.de/config.yml
+++ b/inventory/host_vars/openfoodnetwork.de/config.yml
@@ -4,3 +4,5 @@ rails_env: production
 admin_email: admin@openfoodfoundation.org
 
 mail_domain: openfoodnetwork.de
+
+unicorn_timeout: 120


kristinalim:
I checked `/var/log/nginx/error.log` (actually already rotated, so `error.log.1`) and saw many lines for product import like this:

```
2019/05/01 11:14:53 [error] 8758#8758: *12260 upstream prematurely closed connection while reading response header from upstream, client: 95.91.246.123, server: openfoodnetwork.de, request: "POST /admin/product_import/save_data HTTP/2.0", upstream: "http://unix:/home/openfoodnetwork/apps/openfoodnetwork/shared/sock/unicorn.openfoodnetwork.sock:/admin/product_import/save_data", host: "openfoodnetwork.de", referrer: "https://openfoodnetwork.de/admin/product_import"
```

This means that, for these requests, Unicorn terminates the connection with Nginx, implying that this is indeed a timeout issue.

Matt-Yorkley:
I would try setting something like `unicorn_workers: 3` and maybe `unicorn_timeout: 60`. We should add Germany's server stats to the wiki as well.

kristinalim:
@Matt-Yorkley A couple of months ago, we merged this PR which made sure that only one batch of rows is processed at a time for an open product import page: https://github.com/openfoodfoundation/openfoodnetwork/pull/3387 So an open product import page would use only 1 Unicorn worker at a time.

The current config for Germany is:

```
worker_processes 2
timeout 30
```

Considering that we have 2 Unicorn workers for Germany anyway, what do you think of a separate PR for increasing the number of Unicorn workers to 3? It doesn't seem to me essential to address the current issue and it might need more discussion.

kristinalim:
@Matt-Yorkley This has two approvals already - I'll provision meanwhile so Germany can try to import again. I'll try your recommendation if it still doesn't work. If it works, we can merge this PR and maybe you can create a follow-up PR with the changes you propose?

Thanks!

mkllnk:
A timeout beyond 60 seconds doesn't have the desired affect because nginx is cutting it off. I experimented with this config on the German instance:
```nginx
   keepalive_timeout 360;
   proxy_read_timeout 360;
   proxy_send_timeout 360;
```

Saving an order cycle with 7000 products still failed but it waits six minutes here. Six minutes is ridiculous but I was trying to make it somehow possible. If we want to allow experiments like this, we need to adjust our config: https://github.com/openfoodfoundation/ofn-install/blob/4d8013231eb65b0f77632685e86fe7e916d768fd/inventory/group_vars/all.yml#L218

mkllnk:
I did them directly on the server because there was very little time to solve the problem and ofn-install didn't support this kind of configuration easily. I now created https://github.com/openfoodfoundation/ofn-install/pull/436 to support high timeouts and https://github.com/openfoodfoundation/ofn-install/pull/437 use Germany's beefed up server. It's replacing this PR.

PR #431, opened 2019-05-02
Adjust ENV vars for db:seed 

Matt-Yorkley:
Needed to resolve issues with Semaphore deployments, and some manual deployments (depending on local dev setup).

Moves variables defined in `secrets.yml` out of `deploy.yml`. 

The db:seed task needs those two ENV vars to be available, but we shouldn't reference them directly in `deploy.yml`, as we have decoupled it from `secrets.yml`.


sauloperez:
Now I feel bad. I commented out this task when I bumped into this failure to get going 🙈

PR #432, opened 2019-05-02
Ensure certificates are present for www and naked url

Matt-Yorkley:
We had some reports of browsers intermittently failing to load the homepage when using the naked url `openfoodnetwork.org.uk`. At first it seemed to just be in Internet Explorer :nauseated_face: but it turned out to be intermittently affecting Firefox and Chrome as well. It turned out to be quite a critical issue, as multiple users were finding the site didn't load at all!

This seems to have resolved the problem on UK prod. It looks like France needs this as well, and probably others. We should double-check Katuma prod when provisioning this, as it has a setup with custom domains.

### DIFF
--- a/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
@@ -8,6 +8,10 @@ unicorn_timeout: 120
 
 swapfile_size: 1G
 
+certbot_domains:
+  - www.openfoodnetwork.org.uk
+  - openfoodnetwork.org.uk
+
 postgres_listen_addresses:
   - '*'
 
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -168,7 +168,7 @@ nginx_sites:
     - |
       listen 80;
       listen [::]:80;
-      server_name {{ domain }};
+      server_name {{ certbot_domains | default([domain]) | join(' ') }};
 
       add_header X-Content-Type-Options nosniff always;
       add_header X-Xss-Protection "1; mode=block" always;
@@ -186,7 +186,7 @@ nginx_sites:
     - |
       listen 443 ssl http2;
       listen [::]:443 ssl http2;
-      server_name {{ domain }};
+      server_name {{ certbot_domains | default([domain]) | join(' ') }};
       root {{ app_root }}/public;
 
       ssl_certificate      /etc/letsencrypt/live/{{ certbot_cert_name | default(domain) }}/fullchain.pem;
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
It's not looking good on UK Staging... needs further investigation.

Matt-Yorkley:
OK, it took a while but I think I've finally got this fixed. Tested on UK staging and prod.

Matt-Yorkley:
On Katuma it looks like you've added a manual nginx entry as a workaround for this:

katuma_nginx

This PR should remove the need for your workaround, and make it available to others.

sauloperez:
I remember now. Done in https://github.com/coopdevs/katuma-provisioning/pull/14/files. I desperately want to :fire: this repo :sob:

Matt-Yorkley:
I'm 95% sure this will be compatible with Katuma, but I think we can only test it by applying it to a live server...

lin-d-hop:
Interestingly the API on UK Prod only works on the bare URL and not with the www. No idea why that might be but thought I'd flag. Any thoughts @matt-yorkley

Matt-Yorkley:
> Interestingly the API on UK Prod only works on the bare URL and not with the www. No idea why that might be but thought I'd flag. Any thoughts @Matt-Yorkley

Maybe we provisioned again. This hasn't been merged yet.

lin-d-hop:
@matt-yorkley Any testing tips here? Is this something that @rachel and I can test or does this need a sysadmin to reprovision?

inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
```diff
@@ -8,6 +8,10 @@ unicorn_timeout: 120
 
 swapfile_size: 1G
 
+certbot_domains:
+  - www.openfoodnetwork.org.uk
+  - openfoodnetwork.org.uk
+
```
enricostano: I don't understand the need of a second certificate. Wouldn't a redirect at nginx level do the job?

Matt-Yorkley: It adds both domains to a single certificate. It's more useful for subdomains.

Matt-Yorkley:
@lin-d-hop it needs sysadmin testing.

@sauloperez shall we try this out on Katuma staging, and then production?

inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
enricostano: I don't like the `www` in the URL basically :joy: and I would not add a subdomain that would not be used to a certificate. But that's just me maybe. Don't bother, move on :muscle:

sauloperez: Mmm that's actually the standard way to deal with this common issue isn't it @enricostano? What do experts do?🤣

Matt-Yorkley: Confirmed Katuma had the same issue as UK:

FAIL

and the issue is now fixed. :tada:

PR #436, opened 2019-05-09
Adapt nginx timeouts to unicorn configuration

mkllnk:
We had several cases in which we wanted to increase the timeout to over a minute. While unicorn was running longer than a minute (e.g. `unicorn_timeout 120`), nginx was shutting down the connection after one minute. Now we have the same timeout for nginx and unicorn.

This is needed for Germany and Australia to keep the current configuration. France and UK also have a unicorn timeout of 120 configured and this change may fix some performance problems. When there are many requests, there can be a queue build-up. When the unicorn timeout is longer than the nginx timeout, unicorn will still be busy when nginx tries to pass on the next request.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -215,7 +215,9 @@ nginx_sites:
 
       error_page 500 502 503 504 /500.html;
       client_max_body_size 4G;
-      keepalive_timeout 60;
+      keepalive_timeout {{ unicorn_timeout }};
+      proxy_read_timeout {{ unicorn_timeout }};
+      proxy_send_timeout {{ unicorn_timeout }};
 
       include /etc/nginx/sites-available/ofn/*;
 


inventory/group_vars/all.yml
```diff
@@ -215,7 +215,9 @@ nginx_sites:
 
       error_page 500 502 503 504 /500.html;
       client_max_body_size 4G;
-      keepalive_timeout 60;
+      keepalive_timeout {{ unicorn_timeout }};
+      proxy_read_timeout {{ unicorn_timeout }};
+      proxy_send_timeout {{ unicorn_timeout }};
```
sauloperez: doesn't this mean that nginx will have `unicorn_timeout` to send data to unicorn and the same `unicorn_timeout` to read the response. Am I right? is that what we need?

mkllnk: I'm not sure.

I think, the send is usually quick and then we need a lot of time for the read. But we don't actually know. The product import could be very different.

For example, if the timeout is one minute then nginx could send data for one minute and then read data for one minute before it times out. But I guess that unicorn would time out after the first minute and we see an error message. I don't know though if unicorn starts counting after the read or before.

Anyway, this configuration ensures that nginx is always waiting at least as long as unicorn. We don't want nginx to timeout earlier and process the next request while the unicorn worker is still busy. In that case we would waste time on a worker and requests would pile up. So it's okay or even good if nginx has a higher timeout than unicorn (send plus read).

luisramos0:
Moving to test ready.

inventory/group_vars/all.yml
sauloperez: it's something worth finding out eventually.

sigmundpetersen:
Anything to test or straight-to-merge?

mkllnk:
No, all covered by automated tests now. And I tested the values already.

But we don't really have a process for this. In the main code we require someone else (not the author) to test the code. Does that make sense here? Can we have a staging process where we stage ofn master with an ofn-install pull request version? But that would need to include provisioning...

sauloperez:
sounds like a logical next step @mkllnk but after the v2-related improvements we did, I'd say this repo goes back to normal and the author testing manually might be enough. However, if we find ourselves back to that situation we could give it a chance.

Issue #339, opened 2019-03-19
Provision and deploy Canada's production with ofn-install


sauloperez:
We need to reprovision and then deploy Canada's production server with `ofn-install`. This will require adding its configuration settings to this repo and prevent people from using it while we provision and deploy.  

We need a new server for Canada that is fully set up with `ofn-install`. This will require:

- [x] Add configuration settings to this repo
- [x] Provision the new machine
- [x] Deploy to the new machine
- [x] Stop old server (at night when no one uses it)
- [x] Import database dump to the new server
- [x] Change DNS record to point to the new server
- [x] Ditch the old server

#### Community forum discussion
https://community.openfoodnetwork.org/t/upgrades-rollouts-and-potential-blockers/1620

sauloperez:
@mkllnk shared the CAN Digital ocean credentials in Bitwarden.

Matt-Yorkley:
It's done! :tada:

I'll let the dust settle a bit before deleting the old server.

kirstenalarsen:
ya hoo! happy @tschumilas :)

PR #449, opened 2019-06-04
Add new playbook and role for proxying

Matt-Yorkley:
Related to #339, but useful for all server migrations.

This new playbook and role will direct the server to proxy all requests to the specified target. This is useful for redirecting traffic during a server migration. Running `provision.yml` again on the server will remove the temporary configuration. 

Use `-e "proxy_target=<ip-address>"` to specify the target.

Tested on UK staging. :heavy_check_mark:

### DIFF
--- a/playbooks/temporary_proxy.yml
+++ b/playbooks/temporary_proxy.yml
@@ -1,0 +1,8 @@
+---
+
+- name: temporary proxy
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+
+  roles:
+    - role: temporary_proxy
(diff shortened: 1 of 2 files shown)


sauloperez:
Did we document how to use it and when somewhere @Matt-Yorkley if so, can we merge?

Matt-Yorkley:
Yep, I made a nice wiki page for the whole process of easy migrations with the new playbooks: https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server

PR #452, opened 2019-06-05
Asset migration playbook and role

Matt-Yorkley:
Part of the Canada migration #339.

Playbooks for migrating all the static assets from one server to another.

~~Copies contents of OFN asset folders from one server to another recursively, by executing `scp` on the control machine using the `-3` flag: the user's machine will connect to both the source and destination via SSH and act as a mid-way point for the transfer.~~

My initial version of this was really slow. I've switched to using rsync via the `synchronise` module, and now it's 100 times faster. :tada:

### DIFF
--- a/playbooks/pull_assets.yml
+++ b/playbooks/pull_assets.yml
@@ -1,0 +1,11 @@
+---
+
+- name: Pull assets
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+
+  tasks:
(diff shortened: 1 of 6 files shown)


Matt-Yorkley:
This needs some more testing...

Matt-Yorkley:
Tested and tweaked, used a UK Staging and Vagrant combo. :heavy_check_mark:

PR #453, opened 2019-06-05
Database migration roles

Matt-Yorkley:
Part of the Canada migration #339 (and future migrations).

I've kicked out some old code and added new roles for: 
- backing up a database and copying the db dump to the local machine
- copying a local db dump to a remote server and restoring it

Also DRYed some existing code, as it was re-used here.

Tested on Vagrant, Staging, and Canada production :heavy_check_mark:


roles/backup/tasks/main.yml
```diff
@@ -1,47 +0,0 @@
...
-    flat=yes
-
-# But we don't bother making a whole new copy of all the files every time, just rsync them.
-- name: sync public files to local
-  synchronize:
-    mode=pull
-    src={{ current_path }}/public
-    dest={{ local_rsync_path }}/{{ domain }}/files
-    copy_links=yes
-  tags: rsync
```
luisramos0: I thought the backup was backing up something more than the DB ☝️ 
is it still the case?

Matt-Yorkley: I was quite pleased to :fire: this code, tbh. I don't think it's in use, and it's pretty messy. We can perform the same tasks with the new playbooks I've added today, we're not losing anything.

luisramos0: ok :+1:

enricostano:
@Matt-Yorkley we're working on this https://github.com/coopdevs/backups_role

I hope you find it useful.

Matt-Yorkley:
@enricostano thanks, I'll take a look! This is just for doing a quick export and import during migrations from and old server to a new server.

sauloperez:
> Tested on Vagrant, Staging, and Canada production heavy_check_mark

I guess this is enough testing @luisramos0

PR #454, opened 2019-06-05
Temporary configs for the Canada migration

Matt-Yorkley:
We shouldn't merge this, but I thought it was worth posting up so it's visible.

We can use this as a reference for future migrations.

It turns out migrating a server where the domain name is the same is pretty complicated! I've made a few playbooks today to automate a lot of the migration steps, and this is the final bit of hacking I came up with so we can create the new TLS certificates before doing the switch.

- Use the awesome new support for multi-domain certificates and multi-domain redirects that got merged the other day :wink:
- Use some proxing magic for the Letsencrypt validation check so Certbot can validate both domains from the new host while the old one is still runing
- Create a multi-domain certificate on the new host so the cert is valid for both

### DIFF
--- a/inventory/host_vars/openfoodnetwork.ca/config.yml
+++ b/inventory/host_vars/openfoodnetwork.ca/config.yml
@@ -3,4 +3,40 @@
 domain: openfoodnetwork.ca
 rails_env: production
 
-admin_email:
+admin_email: admin@openfoodnetwork.ca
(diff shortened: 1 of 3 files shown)


sauloperez:
I guess you linked this from the new wiki page you created for this matter. In this case, we can close it as we won't lose it.

PR #456, opened 2019-06-06
Serve www and naked BE domains with the same cert

sauloperez:
This will fix the naked domain for BE, which now returns ERR_EMPTY_RESPONSE.


sauloperez:
Provisioned in production and working!

Issue #444, opened 2019-05-21
Replace custom initializers


mkllnk:
The Australian instance uses a custom initializer to activate the Connect And Learn feature toggle:

```rb
require 'open_food_network/feature_toggle'

module OpenFoodNetwork
  FeatureToggle.class_eval do
    private

    def self.features
      {order_cycles: true,
       connect_learn_homepage: true}
    end
  end
end
```

This initializer would be overridden by an ofn-install deployment. Now we have several options:

1. Enable ofn-install to store custom initializers.
2. Add a new environment variable `OFN_FEATURE_CONNECT_LEARN` to ofn-install and change the `FeatureToggle` class to use the variable.
3. Enable admins to set custom HTML on the homepage and remove the FeatureToggle code.
4. Replace the homepage with our custom Wordpress page.

I think that we don't want to wait for Wordpress. Enabling custom HTML would be obsolete when we switch to a Wordpress homepage. Using an environment variable needs changes in two repositories. It also changes the FeatureToggle class which will soon be obsolete. Option 1 - custom initializers seems to be the only option that could be beneficial in the future. Another use case is an initializer like this:

```rb
class LogBeforeTimeout
  def initialize(app)
    @app = app
  end

  def call(env)
    thr = Thread.new do
      sleep(120 - 1) # set this to Unicorn timeout - 1
      unless Thread.current[:done]
        path = env["PATH_INFO"]
        qs = env["QUERY_STRING"] and path = "#{path}?#{qs}"
        Rails.logger.warn "#{path} I’m about to timeout bringing down my unicorn worker too :("
      end
    end
    @app.call(env)
  ensure
    thr[:done] = true
    thr.run
  end
end
```

I recommend option 1.

sauloperez:
I'd try not to move on the opposite direction of https://github.com/openfoodfoundation/ofn-install/issues/188. Initializers are app code and such need to be with the rest of it.

> Using an environment variable needs changes in two repositories. It also changes the FeatureToggle class which will soon be obsolete.

I actually find this an upside. Whenever we deprecate FeatureToggle we'll easily know this needs to updated as well, whereas being in ofn-install is very likely that we forget about it.

mkllnk:
Will we still have Feature Toggles in the future? The Connect-And-Learn toggle is the only one at the moment and we would like to get rid of it as well.

Matt-Yorkley:
Closed by #462

PR #462, opened 2019-06-25
Add new switch for connect-and-learn feature

mkllnk:
In https://github.com/openfoodfoundation/openfoodnetwork/pull/3970 I'm changing the way feature toggles are switched on or off. It now uses environment variables instead of a hardcoded hash that was set in an initializer.

This is the second and last part of #444.

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -46,4 +46,4 @@ STRIPE_ENDPOINT_SECRET: {{ stripe_endpoint_secret }}
 BUGSNAG_API_KEY: {{ bugsnag_key }}
 {% endif %}
 
-OFN_FEATURE_CONNECT_AND_LEARN: "{{ feature_connect_and_learn | default('false') }}"
+{{ custom_application_env_vars | default('') }}
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -45,3 +45,5 @@ STRIPE_ENDPOINT_SECRET: {{ stripe_endpoint_secret }}
 {% if bugsnag_key is defined %}
 BUGSNAG_API_KEY: {{ bugsnag_key }}
 {% endif %}
+
+OFN_FEATURE_CONNECT_AND_LEARN: "{{ feature_connect_and_learn | default('false') }}"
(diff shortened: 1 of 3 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -45,3 +45,5 @@ STRIPE_ENDPOINT_SECRET: {{ stripe_endpoint_secret }}
 {% if bugsnag_key is defined %}
 BUGSNAG_API_KEY: {{ bugsnag_key }}
 {% endif %}
+
+OFN_FEATURE_CONNECT_AND_LEARN: "{{ feature_connect_and_learn | default('false') }}"
```
mkllnk: What do you think of this way of setting defaults instead of `{% if key is defined %}`?

enricostano: The template is cleaner in this way, but the resulting YAML is full of env vars set to `false`... The problem that I see is that I prefer to check if an env var does exist instead of parsing the content for a literal `false`.

Matt-Yorkley: We could have other env vars set like this, or additional feature toggles in the future, right? I think I'd prefer something more generic and reusable, like:

In Aus `config.yml`:
```
instance_env_vars:
   - "OFN_FEATURE_CONNECT_AND_LEARN: true"
```

In `application.yml.j2`:
```
{% for instance_env_var in instance_env_vars %}
{{ instance_env_var }}
{% endif %}
```

mkllnk: Uhh, interesting. I like that. So in many cases we could actually skip the Ansible variable. For example we don't define `feature_connect_and_learn` and just define the env var in the config.

We are basically storing arbitrary lines in an array and then printing them in a loop. We could also just define a text block in the config that is included.

In Aus `config.yml`:
```yml
application_env_vars: |
  OFN_FEATURE_CONNECT_AND_LEARN: true
```

In `application.yml.j2`:
```yml
{{ application_env_vars }}
```

Of course, this method can only be used for variables that are not used in other parts of the Ansible code. The downside is that we have two ways of setting variables. But I think it's okay.

Should we do that in a separate pull request?

Matt-Yorkley: There's lots of ways to set lists or blocks in variables, I think it's ok to mix and match. 

We may as well do it in this PR, I think.

mkllnk: Okay, please see my new commit, @Matt-Yorkley.

sauloperez:
@mkllnk 

waiting for

> One last thing: you might as well set the actual custom_application_env_vars value in inventory/host_vars/www.openfoodnetwork.org.au/config.yml at the same time, right?

to merge :warning:

mkllnk:
Ah, thank you for the reminder @sauloperez. I added that now.

PR #464, opened 2019-06-28
Add gzip compression to nginx configs

Matt-Yorkley:
Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3858 and https://github.com/openfoodfoundation/openfoodnetwork/issues/#3860

Enables gzip compression via nginx.

Tested on UK staging.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -200,6 +200,10 @@ nginx_sites:
 
       try_files $uri/index.html $uri @unicorn;
       location @unicorn {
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
> Does this work in combination with brotli? Do we need only one?

Not all browsers support brotli. The browser sends a header as part of the request that announces what compression types it supports. The brotli compression will only be used on those requests, so we should merge this one as well.

mkllnk:
Awesome!

PR #468, opened 2019-07-11
Data streaming

Matt-Yorkley:
This PR adds some replication settings and a postgres module to enable the use of data streaming...

With this change the replication is "ready" but not actually active.

### DIFF
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -19,3 +19,6 @@
 
 - src: geerlingguy.postgresql
   version: 1.4.5
+
+- src: libre_ops.wal2json
+  version: 1.0.1
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -21,4 +21,4 @@
   version: 1.4.5
 
 - src: libre_ops.wal2json
-  version: 1.0.1
+  version: 1.0.2
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -64,6 +64,8 @@ postgres_encoding: en_US.utf8
 # to the defaults if present. Make sure you use the "list" notation with the dashes!
 
 default_hba_entries:
+  - {type: local, database: replication, user: "{{ db_user }}", auth_method: trust}
+  - {type: host, database: replication, user: "{{ db_user }}", address: '127.0.0.1/32', auth_method: trust}
   - {type: local, database: all, user: postgres, auth_method: peer}
   - {type: local, database: all, user: all, auth_method: peer}
   - {type: host, database: all, user: all, address: '127.0.0.1/32', auth_method: md5}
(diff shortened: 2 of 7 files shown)


bin/requirements.yml
```diff
@@ -19,3 +19,6 @@
 
 - src: geerlingguy.postgresql
   version: 1.4.5
+
+- src: libre_ops.wal2json
```
sauloperez: Who is exactly does not understand WAL and needs JSON instead?

Matt-Yorkley: Everybody needs JSON! :trollface: 

Basically:
```
Postgres -> Wal2json ->  Kafka queue -> JDBC driver -> Postgres replica
```

But it also means that other things can read those database changes from the queue and parse them easily.

Matt-Yorkley:
Looks like the travis box has multiple postgres versions installed, and it's picked the wrong one...

roles/dbserver/tasks/main.yml
```diff
@@ -35,3 +35,8 @@
 - import_tasks: fix_template_encoding.yml
   tags: template_encoding
 
+- name: install wal2json module
+  include_role:
+    name: libre_ops.wal2json
+  vars:
+    wal2json_set_preload_libararies: false
```
sauloperez: :warning: isn't `libararies` a typo?

inventory/group_vars/all.yml
```diff
@@ -64,6 +64,8 @@ postgres_encoding: en_US.utf8
 # to the defaults if present. Make sure you use the "list" notation with the dashes!
 
 default_hba_entries:
+  - {type: local, database: replication, user: "{{ db_user }}", auth_method: trust}
+  - {type: host, database: replication, user: "{{ db_user }}", address: '127.0.0.1/32', auth_method: trust}
```
sauloperez: I don't understand these entries. There won't be any `replication` database in any of our servers. Aren't these the settings the new DB replica will have?

Matt-Yorkley:
> @Matt-Yorkley This won't be an issue in production and/or staging servers?

The issue I was having was related to building the wal2json module from source, and was definitely only in the Travis environment. I discovered that wal2json is now available as a prebuilt package directly from the official Potsgres repo, so I switched to that instead.

inventory/group_vars/all.yml
Matt-Yorkley: It looks weird, but this is how it needs to be written, based on what postgres expects and how the role writes entries to the file. It's not an extra database, it's basically permission to perform replication.

Docs:
> "The value `replication` specifies that the record matches if a replication connection is requested (note that replication connections do not specify any particular database)."

https://www.postgresql.org/docs/9.5/auth-pg-hba-conf.html

roles/dbserver/tasks/main.yml
Matt-Yorkley: That's how it's spelled in the role too. :see_no_evil: I'll fix it.

Matt-Yorkley: Done: https://github.com/openfoodfoundation/ofn-install/pull/468/commits/2a16f81028f58590833423c1c0f35651977380cc

Issue #385, opened 2019-04-09
Unicorn tuning


Matt-Yorkley:
Much like our default Postgres configuration, it looks like there are some Unicorn tuning options we could look at that are recommended for running Rails on production servers.

The main ones are:

- [x] Set `worker_processes` to around 3-4 (we're trialling this in UK and AUS). 
- [x] Use the `preload_app` feature. It seems to be universally recommended for speeding up production environments, and enables ruby's copy-on-write features.
- [x] Use the `unicorn-worker-killer` gem to automatically clean up worker processes between requests to reduce issues from memory leaks.

Notes:
https://devcenter.heroku.com/articles/rails-unicorn
https://blog.codeship.com/heroku-unicorn/
https://docs.gitlab.com/ee/administration/operations/unicorn.html
https://www.digitalocean.com/community/tutorials/how-to-optimize-unicorn-workers-in-a-ruby-on-rails-app

sauloperez:
I used `unicorn-worker-killer` in the past and that helped mem usage of unicorns to be consistent. I'd like however to check first whether this is a problem for us. #datadriven

mkllnk:
`preload_app` sounds like an excellent idea. That should save some memory.

I wouldn't go for the unicorn-worker-killer though because we don't have any memory leaks. We used to achieve a similar thing with monit which would kill workers when they ran over a memory limit but that was a pain. The configuration needs adjusting regularly.

A long-running unicorn worker occupies more memory than a freshly started one. But that is because it cached a few things which makes it faster to respond. Restarting wipes out that advantage and is additional load on the server.

The memory usage of our workers grows initially but then stays stable:
![Screenshot from 2019-07-18 11-20-18](https://user-images.githubusercontent.com/3524483/61422245-34ce7680-a94e-11e9-907e-326b58517e9e.png)

The dips are unicorn restarts. There is a clear upper limit to the memory usage (= no leak). The big drop is reducing our workers from 6 to 3 because of the concurrent checkout issue.

sauloperez:
so then we can focus or more cost-effective things than `unicorn-worker-killer`.

Matt-Yorkley:
Sounds good. I'll leave this issue open for implementing `preload_app` then.

sauloperez:
Wouldn't it be better to create a separate issue for it with the required details of what needs to be done?

Matt-Yorkley:
[...]

PR #471, opened 2019-07-13
Symlink issue

Matt-Yorkley:
We previously had a bug related to symlinks that meant files were not being backup up correctly in the /rollbacks directory. The change applied to fix it broke the way images are symlinked into the /current directory, but only affected instances that don't use S3, and probably only became apparent after 2 deploys.

I've made some adjustments and added some additional tests to ensure it doesn't happen again and both issues are resolved.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -2,10 +2,13 @@
 
 # Build codebase
 
-- name: remove the old build path
+- name: remove the old build paths
   file:
-    dest: "{{ build_path }}"
+    dest: "{{ item }}"
     state: absent
+  with_items:
+    - "{{ build_path }}"
+    - "{{ current_path }}"
   tags: clone
 
 - name: build codebase from git branch
@@ -62,8 +65,10 @@
 
 # Move new build into place
 
-- name: copy build to current
-  command: cp -TLr {{ build_path }} {{ current_path }}
+- name: move build to current
+  command: mv {{ build_path }} {{ current_path }}
+  args:
+    removes: "{{ build_path }}"
   changed_when: True
 
 # Update the database
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -1,7 +1,13 @@
 --- # Deploy
 
+- name: Check if app is deployed
+  stat:
+    path: "{{ current_path }}/"
+  register: current_path_check
+
 - name: Handle rollbacks
   import_tasks: rollbacks.yml
+  when: current_path_check.stat.exists
   tags: rollbacks
 
 - name: Deploy
--- a/roles/deploy/tasks/rollbacks.yml
+++ b/roles/deploy/tasks/rollbacks.yml
@@ -39,10 +39,8 @@
   args:
     creates: "{{ releases_path }}/{{ timestamp }}.sql.gz"
 
-- name: create a repo rollback version
-  command: mv {{ current_path }} {{ rollback_path }}
-  args:
-    removes: "{{ current_path }}"
+- name: create a repo rollback version # noqa 301
+  command: cp -TLr {{ current_path }} {{ rollback_path }}
   register: new_rollback
 
 - name: create a repo backup version
--- a/tests/deploy/setup.yml
+++ b/tests/deploy/setup.yml
@@ -1,0 +1,19 @@
+---
+- name: setup
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+
+  tasks:
+    - name: setup | deploy | add dummy product image directory
+      become: yes
+      become_user: "{{ unicorn_user }}"
+      file:
+        path: "{{ shared_path }}/spree/products/123/"
+        state: directory
+
+    - name: setup | deploy | add dummy product image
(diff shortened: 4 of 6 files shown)


roles/deploy/tasks/deploy.yml
```diff
@@ -2,10 +2,13 @@
 
 # Build codebase
 
-- name: remove the old build path
+- name: remove the old build paths
```
sauloperez: should we name it `previous release paths`? we use the term `build` already for one of them, so it might lead to confusion.

kristinalim:
@Matt-Yorkley Have you tried deploy to a staging server with these changes? I moved to Test Ready for now.

Matt-Yorkley:
I uploaded an image on Katuma staging and deployed this there. It didn't work the first time, but I tried again and it worked the second time... :thinking:

Matt-Yorkley:
Yeah, it's a quick fix. I wanted it working before any more servers are deployed.

> There can be several seconds between removing `current` and replacing it with `build`. That's downtime.

I think the unicorn workers keeps the whole app in memory, so changes in `current` don't take effect until unicorn is reloaded at the end.

> `cp -TLr` copies everything into `current` preserving all current assets. This reduces downtime because during copying and before unicorn reload all old assets are still available. But since it dereferences symlinks, it's probably the cause of the lost images as well.

We were doing this when copying `build` to `current` (which was incorrect) that's what caused the issue because symlinks in `current` were broken. It's now only doing `cp -TLr` when copying from `current` to `rollback`, so the rollback includes full copies of any symlinked files, but `current` should have it's symlinks preserved (which is what we want).

> I'm also wondering about image storage separately. If we copy or move folders around, they could get out of sync if someone is uploading or deleting images during deploy. So, uploaded assets should be in a shared/symlinked folder to avoid downtime and ensure consistency.

That's what we do already, the images are always going to a single shared/symlinked folder. It should be working now.

mkllnk:
> I think the unicorn workers keeps the whole app in memory, so changes in current don't take effect until unicorn is reloaded at the end.

Not the whole app. Views and assets are not kept in memory.

What do you think about my idea of rewriting unicorn.rb and using a releases folder?

Matt-Yorkley:
I agree there's definitely some optimisations we can make here with the switch from `build` to `current`, and the `releases-old` folder. Do we really need to write a new `unicorn.rb` for each release though? I didn't get that part. 

@sauloperez is keen to use `capistrano` for this, I think that might be a good move. It's basically what you suggested @mkllnk, but handled by a popular and well-written Ansible role that's designed for this purpose.

I think if we start looking at unicorn we should take a look at enabling the `preload_app` feature: https://github.com/openfoodfoundation/ofn-install/issues/385

sauloperez:
You read my mind @Matt-Yorkley. I have the feeling we are overengineering things for a problem that's been solved long ago.

My conclusion: let's adopt Ansistrano Deploy and focus on what makes OFN different.

Matt-Yorkley:
We should keep an eye on this: https://github.com/openfoodfoundation/ofn-install/pull/471#issuecomment-511540993

I'm also wondering if we'll need to run that command to pick up orphaned images from the releases folders one last time after deploying with this. The "new" current folder should have functioning symlinks, but the "old" one that gets displaced could have images that won't be in the shared folders...

sauloperez:
Screenshot from 2019-07-23 15-38-24

Do we need any extra reason to adopt Ansistrano?

sauloperez:
I gave it a try in katuma staging as well, to fully grasp it. I think it's ready so I'll provision and deploy v2.2.1 on Katuma tomorrow morning.

sauloperez:
I deployed it and as expected images where not available since they were still in `releases-old/` and not in `shared/`. Running:

```sh
cp -r releases-old/*/public/spree/products/* current/public/spree/products/
cp -r releases-old/*/public/images/enterprises/* current/public/images/enterprises/
```

fixed it. Scrolling in `/admin/products` I only see the lost images I saw last time (I haven't communicated with that user yet :see_no_evil: ).

So, we're missing CAN to fully fix this.

kristinalim:
Thanks for the additional pointers, @sauloperez.

I've already provisioned and deployed to Germany, which is also affected by this issue. For reference, what I did differently:

1. Used the `-n` option when copying files into `current/`, so that existing files are not overwritten by possibly outdated files in `releases-old/`.
2. Because the oldest release backup is lost after deployment, I already copied the files from the old releases into `current/` before deploying. I did this starting from the oldest backup, but because of the `-n` argument above I realized it would have been better if I started with the most recent backup...
3. After deployment, copied the files without `-n` from the new backup into the shared directory.

mkllnk:
Great strategy @kristinalim. :clap:

kristinalim:
This has already been done in Canada too. :heavy_check_mark: 

Noting that the new shared directories and their contents should be owned by the user and group `openfoodnetwork`. There was a permission error when I first tried to provision because some of the directories previously manually copied (as temporary fix) were owned by `ofn-admin`. The solution was to do a recursive `chown`, e.g.:

```sh
# As ofn-admin
sudo chown -R openfoodnetwork.openfoodnetwork shared/spree
```

PR #473, opened 2019-07-16
Remove old host_vars no one uses anymore

sauloperez:
Anyone using `local.ofn.org`?


enricostano:
> Was it used in the old LXC setup?

Yep.

Issue #477, opened 2019-07-23
[Spike] UK downtime 07/22/2019 03:57 PM UTC - 07/22/2019 04:18 PM UTC


sauloperez:
UK's Happy Apps reported the following yesterday

![Screenshot from 2019-07-23 10-03-23](https://user-images.githubusercontent.com/762088/61694121-278c0000-ad31-11e9-8c61-c17e8a9bae75.png)

but Today as well and twice 

```
An Incident of severity 'CRITICAL' was opened for App: UK Uptime on 07/22/2019 11:56 PM UTC
An Incident of severity 'CRITICAL' was closed for UK Uptime on 07/23/2019 12:22 AM UTC
```

```
An Incident of severity 'CRITICAL' was opened for App: UK Uptime on 07/23/2019 06:51 AM UTC
An Incident of severity 'CRITICAL' was closed for UK Uptime on 07/23/2019 07:12 AM UTC
```

sauloperez:
I'm digging through DJ's logs and as a side note, this might too much for system to process:

```
2019-07-22T03:03:05+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874752) COMPLETED after 6.1707
2019-07-22T03:03:05+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874753) RUNNING
2019-07-22T03:03:11+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874753) COMPLETED after 5.9719
2019-07-22T03:03:11+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874754) RUNNING
2019-07-22T03:03:11+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874754) COMPLETED after 0.7199
2019-07-22T03:03:11+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874755) RUNNING
2019-07-22T03:03:12+0000: [Worker(delayed_job host:production18 pid:26930)] Job ProductsCacheIntegrityCheckerJob (id=1874755) COMPLETED after 0.5178
[...]

PR #480, opened 2019-07-24
Enable Nginx logs in Datadog

sauloperez:
This will hopefully let us correlate and find the root cause of https://github.com/openfoodfoundation/ofn-install/issues/477.

This adds extra load to the servers and increases costs. That's why we will enable it just temporally.


sauloperez:
@Matt-Yorkley I'd like to merge this as it represents the current (or previous?) state of the UK's server. In any case, I shouldn't have waited so long; our infra must be an exact copy of this repo otherwise we miss the point of config management.

If someone happened to modify this manually, please open a PR after this one. This way, the git history will correctly represent the infra's state changes.

Matt-Yorkley:
The state represented here is slightly outdated, and this is superceded by #485.

sauloperez:
Well, my whole point was that this was the state the server was in, so if we ever needed to audit a git log was enough.

PR #484, opened 2019-07-26
Au updates

mkllnk:
This is the config used during the last deploy.

### DIFF
--- a/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
@@ -9,11 +9,10 @@ mail_domain: openfoodnetwork.org
 
 # Optional list of subdomains to register in letsencrypt certificate
 # Defaults to {{ domain }} if list is undefined
-#
-#certbot_domains:
-#  - openfoodnetwork.org.au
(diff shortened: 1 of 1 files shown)


inventory/host_vars/prod2.openfoodnetwork.org.au/config.yml
```diff
@@ -9,18 +9,17 @@ mail_domain: openfoodnetwork.org
...
 # Defaults to {{ domain }} if list is undefined
-#
-#certbot_domains:
-#  - openfoodnetwork.org.au
-#  - www.openfoodnetwork.org.au
-#  - prod2.openfoodnetwork.org.au
+certbot_domains:
+  - openfoodnetwork.org.au
+  - www.openfoodnetwork.org.au
+  - prod2.openfoodnetwork.org.au
```
sauloperez: has this ever existed? I mean, are there links out there in the wild that we might need to maintain?

mkllnk: It has existed during migration to the new server. But there shouldn't be any links to this out there. I can be handy for the next migration to a new server though. While domains are changing and nameservers deliver differing responses, we can use nginx's proxy to point to the right server.

PR #485, opened 2019-07-27
Enable use of nginx or postgres logging separately

Matt-Yorkley:
Separates nginx and postgres logging via datadog so they can be enabled individually.


sauloperez:
For those who wonder, during the UK downtime's last week, the current PostgreSQL logging generated a ton of useless logs. All we inferred from logs came from Nginx's ones.

Matt-Yorkley:
This is tested and live on UK prod.

PR #486, opened 2019-07-29
Automate the deployment of v2

sauloperez:
Automates the manual steps detailed in https://github.com/openfoodfoundation/ofn-install/wiki/Deployment-of-v2. While human need focus and commit mistakes, computers are fast and
reliable and can also be audited.

Better late than nothing :sweat_smile:

### DIFF
--- a/playbooks/v2_deploy.yml
+++ b/playbooks/v2_deploy.yml
@@ -1,0 +1,88 @@
+---
+
+- name: v2 deploy
+  hosts: ofn_servers
+  remote_user: "{{ user }}"
+  become: yes
+  become_user: "{{ unicorn_user }}"
+
+  roles:
+    # Include and enable only
+    - role: maintenance_mode
+      become: yes
+      become_user: root
+
+  handlers:
+    - include: ../roles/shared_handlers/handlers/main.yml
+
+  tasks:
+    - block:
+      - name: stop unicorn
+        service:
+          name: unicorn_{{ app }}
+          state: stopped
+        become: yes
+        become_user: root
+
+      - name: stop delayed job
+        service:
+          name: delayed_job_{{ app }}
+          state: stopped
+        become: yes
+        become_user: root
+
+      - include_role:
+          name: deploy
+
+      - meta: flush_handlers # Ensure handlers run successfully before reporting success
+
+      - name: refresh products cache
+        command: "{{ bundle_path }} exec rake ofn:cache:warm_products"
+        args:
+          chdir: "{{ current_path }}"
+
+      # Include and disable only
+      - name: disable maintenance mode
+        file:
+          dest: "/etc/nginx/maintenance.html"
+          state: absent
+        become: yes
+        become_user: root
+
+      - name: Notify Slack of successful deployment
+        slack:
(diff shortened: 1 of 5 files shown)


sauloperez:
It run successfully in staging.katuma.org. Going for production in US :construction_worker_woman:  :construction:

sauloperez:
Now just imagine if we had automated this before and had a way to schedule ansible jobs... the whole roll-out can be done automatically in a weekly basis :see_no_evil:

playbooks/v2_deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+      - meta: flush_handlers # Ensure handlers run successfully before reporting success
+      - name: refresh products cache
+        command: "{{ bundle_path }} exec rake ofn:cache:warm_products"
+        args:
+          chdir: "{{ current_path }}"
+
+      # Include and disable only
+      - name: disable maintenance mode
+        file:
+          dest: "/etc/nginx/maintenance.html"
```
Matt-Yorkley: You shouldn't delete this file, it stays there permanently. To disable maintenance mode, run the role again with the `disable_maintenance` variable defined/true.

playbooks/v2_deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+          state: absent
+        become: yes
+        become_user: root
+
+      - name: Notify Slack of successful deployment
```
sauloperez: @Matt-Yorkley should we turn these into a role?

Matt-Yorkley: Yeah, it might be nice to extract this stuff now. It's still quite specific to the deploy role, so maybe we could just put these tasks in `roles/deploy/tasks/success.yml` (and failure.yml) and use it where needed with: 
```
-  name: success messages
   include_role:
     name: deploy
     tasks_from: success.yml
```
What do you think?

Matt-Yorkley:
Yeah, it would have been great to do this before... :see_no_evil: 

#AutomateAllTheThings

playbooks/v2_deploy.yml
Matt-Yorkley: Or we could do something similar, but arranged like: `roles/notifications/tasks/deploy_success.yml`,  `roles/notifications/tasks/deploy_failure.yml`,  `roles/notifications/tasks/provision_success.yml` ...?

sauloperez:
Done in US :heavy_check_mark: It felt soooo good seeing the maintenance page, then Ansible executing the cache refresh and finally the site back again :heart:

sauloperez:
@Matt-Yorkley I read your comments after adding my second commit. What are your thoughts about it? I don't mind switching to your proposal.

playbooks/v2_deploy.yml
sauloperez: Considering this will die as soon as we deploy v2 to AU it might be better to wait until next time we end up duplicating it, agree?

sauloperez:
that's a very good idea @luisramos0 . Something to be aware of is that it also refreshes the cache, which might not be needed for the next big thing.

Matt-Yorkley:
I like it :+1:

sauloperez:
Thanks guys! addressed and merging.

PR #490, opened 2019-08-05
Fix missing assets for 10 seconds on deploy

Matt-Yorkley:
Fixes an issue with assets being missing briefly during deployment as originally noted here: https://github.com/openfoodfoundation/ofn-install/pull/471#pullrequestreview-262141281

Back to zero-downtime (tested and working) :+1:

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -2,13 +2,10 @@
 
 # Build codebase
 
-- name: remove the old build paths
+- name: remove the old build path
   file:
-    dest: "{{ item }}"
+    dest: "{{ build_path }}"
     state: absent
-  with_items:
-    - "{{ build_path }}"
-    - "{{ current_path }}"
   tags: clone
 
 - name: build codebase from git branch
@@ -64,6 +61,12 @@
     - restart unicorn
 
 # Move new build into place
+
+- name: remove the current build
+  file:
+    dest: "{{ current_path }}"
+    state: absent
+  tags: clone
 
 - name: move build to current
   command: mv {{ build_path }} {{ current_path }}


roles/deploy/tasks/deploy.yml
```diff
@@ -65,6 +62,12 @@
 
 # Move new build into place
 
+- name: remove the current build
```
sauloperez: do we really need to remove it? AFAIK `move build to current` will overwrite it, isn't it?

Matt-Yorkley: No, there's weird logic with some of the copy and move commands, where if the `dest` is a directory and it already exists, it'll move the `src` _inside_ the `dest` directory, so you get:

```
current/
   current/
      <files>
```

Maybe there's a better way to do it?

sauloperez: @danypr92 @raneq any idea about what @Matt-Yorkley mentions?

could be. I'm not that expert. In any case, all current files will still be in memory up until we reload the server when the unicorn will pick up the new ones, isn't it? so shouldn't be an issue.

kristinalim: Yeah, unfortunately overriding an existing directory is not straightforward to do with the `mv` command. There are gotchas when using `mv` options and when using glob. I don't know how to safely and properly do it either.

But also about using `mv`, it could be an issue when we need an existing file in the old release to be gone in the new release. :thinking: `mv` from build to current would not remove them, and initializers, decorators, CSS, and JS don't need to be referenced to have an effect.

Matt-Yorkley: It turns out that's not the case for assets, including images, precompiled js and css. They disappear and the site looks broken. Hence this PR.

raneq: > @danypr92 @raneq any idea about what @Matt-Yorkley mentions?

Sorry for the delay. Yes, we found [the same issue](https://github.com/coopdevs/odoo-role/pull/54/commits/82cf1861837f72908b9b2e532dba41c51c9e3b8b) while provisioning odoo. As  @kristinalim says, there's no direct way to overwrite a directory with `mv`. If target's empty: you can do

```
$ tree
.
├── source
│   └── etc
└── target

$ mv source/ -T target/
$ tree
.
└── target
    └── etc

```

However, if target's not empty, then mv complains and doesn't let you do the same overwrite.  At the end, it's easier to just remove the whole target dir and rename the source one.

Also it's worth to mention that doing things this way still has a danger. If something weird happens at the move stage, we can end up with neither old nor new content. At odoo role for instance we aren't managing properly this situation :grimacing:

PR #499, opened 2019-08-07
Enable Nginx log monitoring for FR and ES

### LINKED ISSUE #495: Enable Nginx logs monitoring for FR, UK and ES in Datadog
## Description

Nginx logs are required to correlate HTTP requests with server issues as well as to have an overarching view of the usage of the platform.

This was enabled only to the UK and it's has provided tremendously useful.

## Why?

This way we can troubleshoot issues much earlier than when they become a downtime or any other big problem.

## Context

Stems from the retro notes around UK's downtime: https://community.openfoodnetwork.org/t/uks-downtime-retro/1700

sauloperez:
Closes #495 

This will send the Nginx logs to Datadog where they can be inspected an analyzed.

### DIFF
--- a/inventory/host_vars/app.katuma.org/config.yml
+++ b/inventory/host_vars/app.katuma.org/config.yml
@@ -13,3 +13,5 @@ certbot_domains:
 certbot_cert_name: app.katuma.org
 
 swapfile_size: 2G
+
+enable_nginx_logging: true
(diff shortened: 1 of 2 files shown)


sauloperez:
Moving on with this. Sorry not to wait for the 2nd approval but I want to finish things and this one is particularly important.

PR #501, opened 2019-08-10
Add `host_id` variable

Matt-Yorkley:
Adds a simple `host_id` variable that can be used in Ansible tasks. We currently use `{{ ansible_limit }}` in some places to extract values like `es-prod`, but when using commands with groups like `--limit europe` this will fail.

It's really difficult (and messy) to accurately extract this value from the groups variables when a broader group is used, so we can just define it per-host.

It's really useful in things like Datadog where you want to apply a human-readable tag to metrics being sent from a host, as an identifier that can then be used for filtering.

### DIFF
--- a/inventory/host_vars/openfoodireland.eu/config.yml
+++ b/inventory/host_vars/openfoodireland.eu/config.yml
@@ -1,3 +1,4 @@
 ---
 
 domain: openfoodireland.eu
+host_id: ie-prod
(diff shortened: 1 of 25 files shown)


inventory/host_vars/openfoodireland.eu/config.yml
```diff
@@ -1,3 +1,4 @@
 ---
 
 domain: openfoodireland.eu
+host_id: ie-prod
```
sauloperez: shall we remove this file in a separate issue?

sauloperez: Created https://github.com/openfoodfoundation/ofn-install/issues/511

Issue #96, opened 2017-11-03
Implement log rotation


sauloperez:
Implement proper log rotation as discussed in https://github.com/openfoodfoundation/openfoodnetwork/pull/1927

mkllnk:
I added this to our Git post-receive hook which we use to deploy:
```sh
# Rotate log files
logrotate -s "$CURRENT_PATH/log/logrotate-status" "$CURRENT_PATH/log/logrotate.conf"
```

sauloperez:
I'd be great to know the contents of the logrotate.conf as well.

mkllnk:
Ah, sure. Sorry. Here it is:
```
/home/openfoodnetwork/apps/openfoodnetwork/current/log/*.log {
        missingok
        compress
        rotate 5
	weekly
}
```

So I decided to do it on deploy so that the log files are not cut in the middle. But it has the disadvantage that they can grow until the disk is full if you never deploy.

It rotates only if the file is older than a week so that we keep at least the last five weeks of log files.

We could set this up with cron and have size restrictions. But which values are useful depends on how much disk space is free for logs. So I just kept it minimal.

sauloperez:
Although #273 is not an explicit dependency for this, it'd be great to base the priority of this on actual disk usage data after closing #273.

sigmundpetersen:
Closed by #506

PR #506, opened 2019-08-18
Rotate Rails and (Datadog) Postgres Query logs

### LINKED ISSUE #503: Central Logrotate management
At some point we should probably create a role for `logrotate` and manage all log rotation in a clear and centralised way.

Logrotate can be configured to use the `conf.d` pattern :heart:, so we can just define a number of rotation rules in a role and write them to `/etc/logrotate.d`.

And there's a role in `ansible-galaxy` that handles it nicely: https://galaxy.ansible.com/arillso/logrotate

Matt-Yorkley:
Closes #503 and https://github.com/openfoodfoundation/ofn-install/issues/96

Adds central management of logrotate rules.

This should help a bit when investigating issues, and will help clean up bloated and outdated logs. 

We can extend the logrotate rules from this central point any time we need to regularly clean up any kind of files.

### DIFF
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -22,3 +22,6 @@
 
 - src: libre_ops.wal2json
   version: 1.0.2
+
+- src: arillso.logrotate
+  version: 1.4.5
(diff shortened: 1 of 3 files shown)


sauloperez:
Agree with @mkllnk . Having an entire month would be better and I don't think it'll fill up the disk but we need to check. Just implementing this protects from running out of space a whole lot. Thanks for doing it :clap:

Matt-Yorkley:
We could rotate weekly and keep a month's worth?

PR #507, opened 2019-08-30
Remove Skylight

sauloperez:
## What

Since we adopted Skylight to get response times across endpoints and
instances, we failed to get accurate numbers. Our Rails version is not
supported and thus Skylight fails to provide data for the slowest
endpoints, the ones we care about the most. Even with a supported one, we
could potentially hit any limits on tracing and have the same problem.

Recently, we started paying for Datadog's APM and the experience,
although it's still early, has been better. It makes it possible to
correlate between services and other metrics which helps to spot the
underlying issues.

Therefore, having two agents running on the server consumes system
resources so we better get rid of Skylight's one.

### DIFF
--- a/roles/skylight_uninstaller/tasks/main.yml
+++ b/roles/skylight_uninstaller/tasks/main.yml
@@ -1,0 +1,33 @@
+---
+- name: get agent PID
+  shell: "set -o pipefail && ps -ef | grep -v grep | grep -w skylightd | awk '{print $2}'"
+  args:
+    executable: /bin/bash
+  ignore_errors: yes
+  register: skylight_agent_pid
+  changed_when: false
+  become: yes
+
+- name: kill agent
+  command: "kill {{ skylight_agent_pid.stdout }}"
+  when: skylight_agent_pid.stdout | length > 0
+  become: yes
+
+- name: wait for process to die
+  wait_for:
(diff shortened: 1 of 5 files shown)


roles/skylight_uninstaller/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: wait for process to die
+  wait_for:
+    path: "/proc/{{ skylight_agent_pid.stdout }}/status"
+    state: absent
+  with_items: "{{ skylight_agent_pid.stdout }}"
+  become: yes
+  ignore_errors: yes
+  register: result
+
+- name: force kill stuck process
```
Matt-Yorkley: Was this really needed?

sauloperez:
Added WIP because there's something wrong when together with https://github.com/openfoodfoundation/openfoodnetwork/pull/4202. I guess the Skylight daemon stops itself.

sauloperez:
Confirmed the `skylightd` stops when restarting the unicorn and it doesn't come back because the gem is gone, I guess.

After a deploy

```sh
openfoodnetwork@staging:~/apps/openfoodnetwork/current$ ps aux | grep skylightd
openfoo+ 10568  0.0  0.0  14228   932 pts/0    S+   14:34   0:00 grep --color=auto skylightd
```

PR #508, opened 2019-09-03
Enable remote access from Metabase to Katuma's DB

sauloperez:
This makes it possible to execute queries from Metabase against Katuma's
production DB, as a first towards a global Business Intelligence tool
while allowing Katuma to bill its customers reliably.

This enables us to start collecting the pricing for each of the currently active hubs and producers and start charging them reliably.

Next step will be to use https://github.com/openfoodfoundation/ofn-data data streaming.

### DIFF
--- a/inventory/host_vars/app.katuma.org/config.yml
+++ b/inventory/host_vars/app.katuma.org/config.yml
@@ -16,3 +16,9 @@ certbot_cert_name: app.katuma.org
 swapfile_size: 2G
 
 enable_nginx_logging: true
+
+postgres_listen_addresses:
+  - '*'
+
+custom_hba_entries:
+  - { type: hostssl, database: "{{ db }}", user: metabase, address: '167.99.89.242/32', auth_method: md5 }


sauloperez:
Working on Katuma production already :tada:

enricostano:
In order to secure this, I suggest to enforce communication over SSL/TLS and use a whitelist of possible servers connecting to the exposed Postgres.

With @danypr92 we did this already, if you need help feel free to ping us.

Matt-Yorkley:
> In order to secure this, I suggest to enforce communication over SSL/TLS and use a whitelist of possible servers connecting to the exposed Postgres.

That's already the case with the current setup. We're adding a specific user, enforcing TLS, and only allowing the connection form a single IP address:

https://github.com/openfoodfoundation/ofn-install/blob/1ad70b40da52b06c454b89311440f8e70b6c7fdb/inventory/host_vars/app.katuma.org/config.yml#L24

The other default HBA entries only match connections from localhost addresses and internal system calls, all others will not be accepted.

The `*` character in `postgres_listen_addresses` actually applies to the IP(s) that the server's NICs are listening on, not the source IPs of inbound connections.

danypr92:
Sorry but I don see the message until now...

I wrote some notes when we apply this secure communication in a project: https://github.com/coopdevs/handbook/wiki/Secure-TCP-IP-Connections-with-SSL

If it can help...

PR #510, opened 2019-09-05
Update memcached config

### LINKED ISSUE #509: Configure memcached to use 128 MB max memory and 6 threads
## Description

For UK, we made the following changes to memcached config in `/etc/memcached.conf` manually:

- Increased the max memory from the default of 64 MB to 128 MB (existing line in config file)
- Increase the number of threads from the default of 4 to 6 (new line in config file)

Currently, memcached config is not managed through `ofn-install`. This should be done, and the other instances should use this configuration too.

## Why?

See Slack thread [here](https://app.slack.com/client/T02G54U79/CLSDUBV8A/thread/CLSDUBV8A-1567695797.125700):

1. We were using 64 MB max memory - This seems low, and we are probably exceeding this in some instances. (I did not check actual usage, actually.)
2. We were using default of 4 threads - At time of check in UK, we had 6 incoming memcached connections (`netstat -n | grep 11211`).

Also another cause for concern, but not related to this PR: There are some interesting results for "memcached cpu": https://www.google.com.ph/search?q=memcached+cpu

## Context

#497

Matt-Yorkley:
Closes #509

Increases `memcached` memory from 64 to 128 MB, and max threads from 4 to 6.

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -1,0 +1,7 @@
+---
+
+- name: restart memcached
+  service:
+    name: memcached
+    state: restarted
+  become: yes
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -43,3 +43,11 @@
       - memcached
   tags: packages
 
+- name: add memcached configuration
+  template:
+    src: memcached.conf.j2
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
Tested on UK Staging :heavy_check_mark:

PR #512, opened 2019-09-06
Add "all-prod" group for all production servers

Matt-Yorkley:
:rocket:


sauloperez:
I don't have a particular opinion on this. Even if that happened it's not critical and cannot go unnoticed either 🤷‍♂

Issue #516, opened 2019-09-10
brotli_nginx fails to find nginx source


mkllnk:
> TASK [brotli_nginx : download and extract nginx source] ***********************************************************************************************************************************************************
> Wednesday 11 September 2019  08:39:05 +1000 (0:00:06.848)       0:17:27.175 *** 
> fatal: [staging.openfoodnetwork.org.au]: FAILED! => {"changed": false, "msg": "Failure downloading http://nginx.org/download/nginx-1.10.3 (Ubuntu).tar.gz, HTTP Error 404: Not Found"}

That file isn't there. I would also strongly recommend to download via https. I found this link:
https://nginx.org/download/nginx-1.10.3.tar.gz

mkllnk:
Somehow the version output of nginx is different on this server.

```sh
ssh ofn-au-prod
nginx -v
# nginx version: nginx/1.16.1

ssh ofn-au-staging
nginx -v
# nginx version: nginx/1.10.3 (Ubuntu)
```

I think the role assumes that nginx has been installed from a custom ppa before.

mkllnk:
I solved the version problem and got a new one:

[...]

PR #518, opened 2019-09-13
Add "active-staging" group for active stagings

sauloperez:
This enables us to provision staging servers in one go and gives `master` a try while testing the app. This way we are a bit closer from the state production servers will be in.

### DIFF
--- a/inventory/hosts
+++ b/inventory/hosts
@@ -164,6 +164,14 @@ au-prod
 za-prod
 
 #------------------------------------------------------------------------------
+# Staging Servers
+
+[active-staging:children]
+uk-staging
+es-staging
+fr-staging
+
+#------------------------------------------------------------------------------
 # All
 
 [ofn_servers:children]


sauloperez:
https://github.com/openfoodfoundation/ofn-install/issues/519 could be a solution

luisramos0:
is this working Pau?

why is aus not included?

sauloperez:
> is this working Pau?

It does nothing per se but until we don't solve https://github.com/openfoodfoundation/ofn-install/issues/519 using it while provisioning can be harmful, so I'll close it to avoid any trouble.

> why is aus not included?

AFAIK we don't actively use it for testing and it's currently broken according to https://github.com/openfoodfoundation/ofn-install/issues/516

luisramos0:
:+1:

Matt-Yorkley:
Are we not adding this group? It seems reasonable. We just added the `all-prod` group for production servers, so maybe we can use one naming scheme for both groups?

sauloperez:
Yes, it scares me to have someone provisioning by mistake until ofn-install doesn't support multi-host secrets provisioning.

Issue #513, opened 2019-09-09
Zapier integration for USA Server


lauriewayne:
Set up USA server (and maybe ofn-install too?) so that users can connect to the database for reporting using Zapier. This issue is a cousin to #403  and maybe #400?

Matt-Yorkley:
I'll get this done today :+1:

sauloperez:
This is done but just needs to be provisioned against the production server to take effect. Could you handle it @Matt-Yorkley ? I would just provision all servers before waiting for the next release.

PR #524, opened 2019-09-19
Allow Zapier user access to Postgres for USA

Matt-Yorkley:
Part 2 of: https://github.com/openfoodfoundation/ofn-install/issues/513

Allows Zapier access to Postgres and sets Postgres to listen for external connections (for the Zapier IP only).


sauloperez:
This passed CI so there's not much more we can test here.

PR #527, opened 2019-09-21
Enable Datadog's APM in UK staging

sauloperez:
We need to be able to test changes in monitoring quickly and reliably without having to risk production. This will incur an extra cost but it'll save money in dev hours testing things in the long run.

### DIFF
--- a/inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
@@ -7,3 +7,5 @@ rails_env: staging
 admin_email: dev.ofnuk@gmail.com
 
 unicorn_timeout: 120
+
+enable_rails_apm: "true" # has to be explicitly defined as a string due to some datadog bug
--- a/roles/datadog/tasks/main.yml
+++ b/roles/datadog/tasks/main.yml
@@ -10,7 +10,8 @@
         - "env:{{ rails_env }}"
         - "host-id:{{ host_id | default(ansible_limit) }}"
       logs_enabled: true
-      apm_enabled: "{{ enable_rails_apm | default('false') }}"
+      apm_config:
+        enabled: "{{ enable_rails_apm | default('false') }}"
     datadog_config_ex:
       trace.config:
         env: "{{ rails_env }}"


sauloperez:
Still with this I don't see data coming through when using the app in UK staging. I must be missing pretty stupid. Any ideas @Matt-Yorkley ?

Screenshot from 2019-09-21 14-04-13

I thought it could be something related the `env:staging` but I don't see anything that might be skipping that in the role.

Matt-Yorkley:
I think you'll need a unicorn reload, because the APM is switched in via an initializer...

sauloperez:
🤦

sauloperez:
Still no luck. I reloaded the unicorn, the initializer seems properly configured

```
openfoodnetwork@staging:~/apps/openfoodnetwork/current$ bundle exec rails c
WARNING: Use strings for Figaro configuration. true was converted to "true".
The PGconn, PGresult, and PGError constants are deprecated, and will be
removed as of version 1.0.

You should use PG::Connection, PG::Result, and PG::Error instead, respectively.

Called from /home/openfoodnetwork/.gem/ruby/2.1.0/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:28:in `new'
Loading staging environment (Rails 3.2.22.5)
irb(main):001:0> ENV['DATADOG_RAILS_APM']
=> "true"
```

the agent's config has been fixed but raising the log level to DEBUG doesn't provide any details regarding APM. I'm currently reprovisioning and I'll deploy master right after to leave uk-staging in a known state.

Could you please @Matt-Yorkley take a look?

sauloperez:
ok, it seems that I just needed to wait a bit. Now data is coming through

Captura de pantalla 2019-09-23 a les 13 18 12

roles/datadog/tasks/main.yml
```diff
@@ -10,7 +10,8 @@
         - "env:{{ rails_env }}"
         - "host-id:{{ host_id | default(ansible_limit) }}"
       logs_enabled: true
-      apm_enabled: "{{ enable_rails_apm | default('false') }}"
```
Matt-Yorkley: I'm not sure about this change. `apm_enabled` is the key used in the datadog role.

inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
```diff
@@ -7,3 +7,5 @@ rails_env: staging
 admin_email: dev.ofnuk@gmail.com
 
 unicorn_timeout: 120
+
+enable_rails_apm: "true" # has to be explicitly defined as a string due to some datadog bug
```
Matt-Yorkley: Do we really want to permanently enable this in staging?

sauloperez: IMO it's the only reliable way to ensure we don't break production while adding things to Datadog. I think we'll spend much more time checking whether things could break production and possibly fixing it when it's too late than the extra money this costs. 

Besides, monitoring is a too important piece to assume any unnecessary risks.

roles/datadog/tasks/main.yml
kristinalim: @sauloperez Is it possible you're using a newer version of `Datadog.datadog`? The README for version `2.4.0` specified in `bin/requirements.yml` does not mention `apm_config`.

inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml
kristinalim: I think this is great. :+1:

PR #530, opened 2019-09-23
Install Ruby 2.2.10

Matt-Yorkley:
Part 2 of: https://github.com/openfoodfoundation/openfoodnetwork/issues/2891

Provisions Ruby `2.2.10` ready for use with PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/4301

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -51,7 +51,7 @@ swapfile_size: false
 
 #----------------------------------------------------------------------
 # Rails variables
-ruby_version: 2.1.9
+ruby_version: 2.2.10
 env:
   RAILS_ENV: "{{ rails_env }}"
   PATH: "{{ gem_home }}/bin:{{ ansible_env.PATH }}"


Matt-Yorkley:
Tested and working! :tada:

Matt-Yorkley:
Yep, there's n issue here that needs addressing: openfoodfoundation/openfoodnetwork#4301

luisramos0:
moved to in dev and marked as blocked by 4301

Matt-Yorkley:
As per the last Ruby bump:

> The build will not have both versions installed because it's starting from a fresh server, installing 2.2.10 and then deploying master (using 2.1.9).

The `ofn-install` CI build will be green when the `master` branch of the main repo uses `2.2.10`.

sauloperez:
I feel like this is still too manual. I'd be much easier to manage if we installed both the current and feature version so that the infra is ready to manage the app running in either version. As a result of the decoupling, we could freely merge the ofn-install PR and have it run either app version and so no need to sync merges and provisions.

sauloperez:
Looks like the role we use https://github.com/zzet/ansible-rbenv-role supports passing a list of ruby versions but also removes any existing ones. So it's should be pretty straightforward to do.

Matt-Yorkley:
It only removes old versions if you set a specific variable. So the previous version will still be available.

luisramos0:
ok, build is green.

luisramos0:
I merged 4310, checked green builds on both ofn and ofn-install. Provisioned both staging FR and staging UK. Both looking good. I am merging this now.
We need to provision all other servers afterwards:
- [ ] ka staging
- [ ] au staging
- [ ] uk prod
- [ ] au prod
- [ ] fr prod
- [ ] ka prod
- [ ] ca prod
- [ ] us prod
- [ ] de prod
- [ ] be prod

PR #533, opened 2019-09-24
Nginx version

Matt-Yorkley:
- Removes temporary fix for old (resolved) nginx ppa issue
- Adjusts the nginx and brotli tasks to fix issues with outdated nginx versions being present when brotli is built


Matt-Yorkley:
Aha... Travis says no. We have to build `brotli` first, because we add it to the nginx config with the `jdauphant.nginx` role, so the role's checks will fail loudly if it's not there. 

_Obviously_ :roll_eyes:

Matt-Yorkley:
I've adjusted the roles a bit, this should resolve the outstanding issues.

enricostano:
Can you selectively run "half" of `jdauphant.nginx`, install `brotli` and then run the rest of `jdauphant.nginx`'s tasks?

I've being doing this with the postgresql role until they solved the users/databases chicken/egg issue.

Matt-Yorkley:
I've basically done that, yes. That reminds me, we can remove that messy postgres role workaround now. :+1:

luisramos0:
@Matt-Yorkley if you clarify what needs to be tested I can take care of this.

Matt-Yorkley:
This is a tricky one to test, as the issue it resolves was only present on a fresh install of a new server. That's why I use Vagrant for testing ofn-install PRs. So you need to start with a fresh Ubuntu 16 image, run setup.yml, then provision.yml and nginx and brotli should be installed correctly.

luisramos0:
I dont think I am the best person to get this test done...

Matt-Yorkley:
So I've just carefully retested this myself, and I think it's good to go. The CI build is also green.

luisramos0:
ok, thanks Matt! I'd merge now.

PR #537, opened 2019-10-11
Only install required gems in production

### LINKED ISSUE #467: Install only required dependencies via bundler
A recent deployment failed because a gem dependency was deleted from the central repository. While it's very difficult to prevent this problem in general (setup our own gem repository), we can lower the risk and speed up deployments by only installing required dependencies.

The offending dependency in the last deployment, `fuubar`, is only for running specs and didn't need to be installed. Non-production gems are also more likely to be less well maintained and cause issues.

This idea isn't new but was hidden in a code comment:

https://github.com/openfoodfoundation/ofn-install/blob/17cfe0f22501e7f01dc6b661011f00211d96685c/roles/deploy/tasks/deploy.yml#L48-L51

Matt-Yorkley:
Closes #467 

Does not install `development` and `test` dependencies in production.

Tested in Vagrant. :heavy_check_mark:

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -59,9 +59,11 @@
   changed_when: bundler.stdout | length > 0
 
 - name: bundle install app dependencies
-  #TODO make the "--without development" part conditional on rails_env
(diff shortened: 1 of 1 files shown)


sigmundpetersen:
`no-test`?

PR #538, opened 2019-10-14
Make assets precompile tasks always clean assets before

luisramos0:
#### What? Why?
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/4328 

Cleaning assets before precompiling assets make translations always being updated, even if JS code is not updated.

Maybe there's some i18n-js config we could change to achieve this but I have not managed to find it.

#### What should we test?
I tested this PR and the problem is fixed because a new JS bundle is always generated.
Maybe another dev can verify the process again in staging. Or we can just deploy 2.5.2 again to FR live and see if the problem is fixed, it's harmless.

#### Release notes
Changelog Category: Fixed
Translations stored in the browser (Javascript) are now always updated even if no code changes on the javascript side.

### DIFF
--- a/roles/shared_handlers/handlers/main.yml
+++ b/roles/shared_handlers/handlers/main.yml
@@ -1,6 +1,6 @@
 # This role holds reusable handlers that can be included in multiple playbooks to keep things DRY
 - name: precompile assets
-  command: bash -lc "bundle exec rake assets:precompile:primary RAILS_GROUPS=assets RAILS_ENV={{ rails_env }}"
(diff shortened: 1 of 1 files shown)


luisramos0:
nice Kristina, thanks.

I am not testing this because I am the author. Another dev should test it.

The test is as simple as deploying using this branch to staging FR and then checking the server.
There should be a new js bundle file here
`ls -lart /home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/darkswarm/all-*`

Matt-Yorkley:
Tested on FR staging. Looks good!

luisramos0:
thanks Matt!

PR #539, opened 2019-10-14
Make Canada's server to use en_CA

luisramos0:
Canada's server is using en. This makes it use en_CA.

We may want to run the following SQL on the server after deploy so that all users will immeditely see this language:
update spree_users set locale = 'en_CA' where locale = 'en';

### DIFF
--- a/inventory/group_vars/ca.yml
+++ b/inventory/group_vars/ca.yml
@@ -4,8 +4,8 @@
 checkout_zone: Canada
 country_code: CA
 currency: CAD
-locale: en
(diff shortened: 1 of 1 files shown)


sauloperez:
This will need to be notified to the release manager in charge @luisramos0

luisramos0:
I think I will just provision/deploy Canada and get this fixed.

luisramos0:
ok, canada provisioned and deployed, SQL executed. Looking good. I'll ask instance manager to verify.

PR #540, opened 2019-10-22
Increase UK timeouts

Matt-Yorkley:
Increases UK timeouts to 6 minutes... 

:sob:


sauloperez:
We gotta stop doing this. Could you please explain what was the issue that led us here?

lin-d-hop:
The previous timeout increase did not persist and thus was reverted back to 2 mins on the last deploy.

This just isn't long enough for most reports for our users.

The issue was a regression of an old increase - not a new requirement for an increase.

Chat to me on slack if you want to know more.

kristinalim:
The previous build failed because of an error when fetching a dependency. I re-ran the build and it passed. :100: Moving to Ready to Go.

PR #543, opened 2019-11-05
Change en to en_DE on the german instance

luisramos0:
As requested by Kirsten.

This needs the new language en_DE to be added to the main ofn repo first.


luisramos0:
heads up @sauloperez, we need to make sure en_DE gets fetched from transifex before the next release.

sigmundpetersen:
en_DE is already merged to the codebase

PR #545, opened 2019-11-07
Increase timeouts for French production

Matt-Yorkley:
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/4439

Increases timeouts on French prod from 120 to 240. 

This is due to `Orders and Fulfilment` reports timing out. These should be fixed soon :crossed_fingers: so this is a temporary measure.


Matt-Yorkley:
Can whoever merges this also provision `fr-prod` with it as well? Thanks :kissing_heart:

Matt-Yorkley:
Rebased for Ruby `2.2.10`. I think we can merge this now.

PR #550, opened 2019-11-28
Increase uk timeout to 720secs

luisramos0:
Increase Uk timeout to make reports run.
This has been provisioned to uk and the reports now work after 7mins.

Fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/4521


sauloperez:
These reports desperately need to be run in a background job. This is insane... 🙈

sauloperez:
I merging this since this was already provisioned. It's a fact that needs to be reflected in the code.

PR #556, opened 2020-01-07
Allow setting an alternate region for S3 backups bucket

Matt-Yorkley:
Twinned with https://github.com/openfoodfoundation/openfoodnetwork/pull/4653

Allows setting an alternate region for S3 backups bucket.


sauloperez:
Tested in https://github.com/openfoodfoundation/openfoodnetwork/pull/4653#issuecomment-578111918 and working.

PR #560, opened 2020-02-05
Add non-www support for certbot in fr-prod

kristinalim:
This should resolve the certificate issue when accessing https://openfoodfrance.org. Right now, only www.openfoodfrance.org is supported.

Through curl:

```shell
$ curl https://openfoodfrance.org
curl: (51) SSL: no alternative certificate subject name matches target host name 'openfoodfrance.org'
```

From the browser, I get `NET::ERR_CERT_COMMON_NAME_INVALID`.


luisramos0:
I provisioned fr-prod, but the problem remained.

Then I logged in to the server and run certbot manually (I just had to select both domains and select "redirect from https to http". Problem solved.

I think provisioning should have fixed it, right?

PR #562, opened 2020-02-12
Disable Datadog in stagings to save money

sauloperez:
Last bill was $183.16 covering Datadog January 2020: 3 Commited Hosts
@ $54.00, 3 On-Demand Hosts @ $54.00, 2 On-Demand APM Hosts @ $72.00,
1.239 On-Demand Log Events (millions) @ $3.16.

This change should save us $90 every month, not a whole lot but we're
not checking staging's metrics. We can enable it temporally when
changing Datadog's config and disable it back again.


sauloperez:
I just provisioned this. All good.

sauloperez:
I had to reprovision UK. Now that one is gone too.

PR #563, opened 2020-02-12
Better deployments

Matt-Yorkley:
2-minute PR to improve deployments. Switches some tasks around to reduce downtime to a minimum.

Needs testing on staging.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -75,20 +75,6 @@
     - precompile assets
     - restart unicorn
 
-# Move new build into place
-
-- name: remove the current build
-  file:
-    dest: "{{ current_path }}"
-    state: absent
-  tags: clone
-
-- name: move build to current
-  command: mv {{ build_path }} {{ current_path }}
-  args:
-    removes: "{{ build_path }}"
-  changed_when: True
-
 # Update the database
 
 - name: check database status
@@ -143,3 +129,17 @@
   changed_when: True
   args:
     chdir: "{{ current_path }}"
+
+# Move new build into place
+
+- name: remove the current build
+  file:
+    dest: "{{ current_path }}"
+    state: absent
+  tags: clone
+
+- name: move build to current
+  command: mv {{ build_path }} {{ current_path }}
+  args:
+    removes: "{{ build_path }}"
+  changed_when: True
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -130,6 +130,23 @@
   args:
     chdir: "{{ build_path }}"
 
+# Precompile assets
+
+- name: precompile assets
+  command: bash -lc "bundle exec rake assets:clean assets:precompile:primary RAILS_GROUPS=assets RAILS_ENV={{ rails_env }}"
+  args:
+    chdir: "{{ current_path }}"
+  become: yes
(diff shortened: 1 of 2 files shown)


luisramos0:
we precompile assets after everything is working again?
what happens meanwhile if a user accesses the app while the assets are being precompiled? they will be waiting for the process to finish right?

Matt-Yorkley:
With this change the previous assets should still be in place whilst the new assets are being compiled, so they won't be "missing" during the process.

luisramos0:
ok, nice, we can test it in the next deploy by accessing the app while the assets are being precompiled.
if we merge this until next Tuesday I will be doing the deploy and I can try in prod.
maybe better if someone else tests it in staging.

Matt-Yorkley:
This needs testing before merging, and the build indicates it needs a bit of tweaking.

Matt-Yorkley:
I made some minor adjustments and now the build is green.

I deployed master to French staging and browsed / reloaded pages continuously whilst deploying and couldn't detect any changes or disruption at all. 

I think we're basically back to zero-downtime deployments with this PR. :tada:

roles/deploy/tasks/deploy.yml
```diff
@@ -111,14 +111,14 @@
 
 # Precompile assets
 
-- name: precompile assets
+- name: precompile assets # noqa 301
```
mkllnk: Since this is a newly built directory, we could test for `all.js` or similar before running this.

Matt-Yorkley: We could use the `creates:` directive on this task to achieve that. :+1:

Matt-Yorkley: I've changed my mind on this. I don't think we should couple this task to a specific file existing or not in a specific path, it means if the asset pipeline changes in the future deployments will break...

Matt-Yorkley:
> Can we compile the assets before migrating to minimise the time between database and code change?

I just tried it, but it broke the build. :disappointed: 

It looks like deployments on new servers would no longer work. Maybe we could split the deploy playbook into two different versions at some point (for new and existing servers), but it might get messy...

mkllnk:
Not in this PR but I like the idea of splitting that a bit. It would mean that we don't have the schema loading and seeding in the deploy step which should be safer.

Did you move the seeding further down as well or just the migration?

sauloperez:
I totally agree about what you said about migrations @mkllnk . I checked what Capistrano does and it seems we already do what's best. We basically match Capistrano's flow:

```
deploy
  deploy:starting
    [before]
      deploy:ensure_stage
      deploy:set_shared_assets
    deploy:check
  deploy:started
  deploy:updating
    git:create_release
    deploy:symlink:shared
  deploy:updated
    [before]
      deploy:bundle
    [after]
      deploy:migrate
      deploy:compile_assets
      deploy:normalize_assets
  deploy:publishing
    deploy:symlink:release
  deploy:published
  deploy:finishing
    deploy:cleanup
  deploy:finished
    deploy:log_revision
```

source: https://capistranorb.com/documentation/getting-started/flow/

luisramos0:
tested this in FR staging. looks good, I can reload the app while assets are precompiling.

the problem is really when the old code is not compatible with the new db, isnt it?

sauloperez:
Yep, we'll need to be aware of it and write backward-compatible migrations or the app will fail for the duration of the precompilation.

PR #569, opened 2020-02-19
Avoid mistakenly deploying master to production due to minor typos

Matt-Yorkley:
Now if you don't type the command correctly when deploying a release on a production server, you'll see this prompt _before_ the playbook runs, instead of realising after the playbook has finished that you've accidentally deployed `master` to production because of a typo... :see_no_evil: : 

deploying_master

:muscle: :muscle: :muscle:


sauloperez:
Working as expected :ok_hand:

Issue #571, opened 2020-02-20
Update French Production certificate


Matt-Yorkley:
The certificate on French Production is only valid for `www.openfoodfrance.org`, and not the bare domain `openfoodfrance.org`.

The process for updating certs is annoying and manual, but a recent patch in our certbot role cna help here.

We need to:

- [x] Update the `certbot_nginx` role: #572
- [x] Make sure the configs for `fr-prod` include both domains
- [x] Use the new option in the updated certbot role to force-regenerate the certificate with the new configs

Matt-Yorkley:
Done.

PR #572, opened 2020-02-20
Bump certbot_nginx role

Matt-Yorkley:
Related to #571 

Bumps the `certbot_nginx` role. 

This update includes a patch I submitted that allows certificates to be regenerated when the configs have changed.

This means if we update the certificate configs for a server, for example with France where it previously only had `www.openfoodfrance.org` in the list of domains but we have added the bare domain `openfoodfrance.org` to the list, we can use the following Ansible command to recreate the certificate with the updated configuration (the default behaviour is to ignore it if a certificate is already present):

`ansible-playbook playbooks/provision.yml --limit fr-prod --tags certbot -e "certbot_force_update=true"`

### DIFF
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -9,7 +9,7 @@
   version: v2.8.1
 
 - src: coopdevs.certbot_nginx
-  version: 0.0.4
+  version: v0.1.0
 
 - src: oefenweb.swapfile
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Tested and working :heavy_check_mark:

PR #573, opened 2020-02-21
Precompiling in Rails 4

Matt-Yorkley:
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/4819

Precompiling assets looks a bit different in Rails 4, it's one task instead of two, and the syntax is changed. 

This PR runs slightly different precompile tasks during deployments, based on a quick check in the Gemfile to see if Rails 3 is being used.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -111,19 +111,8 @@
 
 # Precompile assets
 
-- name: precompile assets # noqa 301
-  command: bash -lc "bundle exec rake assets:clean assets:precompile:primary RAILS_GROUPS=assets RAILS_ENV={{ rails_env }}"
-  args:
-    chdir: "{{ build_path }}"
-  become: yes
-  become_user: "{{ unicorn_user }}"
-
-- name: precompile nondigest assets # noqa 301
-  command: bash -lc "bundle exec rake assets:precompile:nondigest RAILS_GROUPS=assets RAILS_ENV={{ rails_env }}"
-  args:
-    chdir: "{{ build_path }}"
-  become: yes
-  become_user: "{{ unicorn_user }}"
+- name: precompile assets
+  import_tasks: precompile.yml
 
 # Move new build into place
 
--- a/roles/deploy/tasks/precompile.yml
+++ b/roles/deploy/tasks/precompile.yml
@@ -1,0 +1,15 @@
+---
+
+- name: precompile assets # noqa 301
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
I guess at some point it would be good to set a staging server aside for the upgrade, and test for potential deployment issues...

sauloperez:
whatever implementation we take @mkllnk , we do need @Matt-Yorkley 's idea of making ofn-install clever enough to perform the right action based on the Rails version. Only this way we'll be able to manage things without dying IMO.

Matt-Yorkley:
So the alternate proposal is to keep the script with the precompiling rake tasks in the main repo instead of in ofn-install, so we can change it per-branch, right?

Matt-Yorkley:
I tested this today with Rails 4. This playbook correctly checks the version in both Rails 3 and Rails 4, and runs the appropriate command.

Unfortunately the deployed app does not work yet on Staging servers. It works fine in development. Maybe there's some caching changes that require extra clearing, or some other configs for production/staging that need adapting...

mkllnk:
> the alternate proposal is to keep the script with the precompiling rake tasks in the main repo instead of in ofn-install, so we can change it per-branch, right?

Yes, that was my idea. It would be much easier to adapt to future Rails versions and is a simpler implementation. Plus we can re-use the code when testing as devs.

sauloperez:
so what @Matt-Yorkley . Do you agree on @mkllnk approach?

luisramos0:
@Matt-Yorkley this PR can go as is or we can decide to do it in the ofn repo. It's up to you :muscle:

luisramos0:
I have good reasons to try to push this forward :-) We are almost ready for deploy for real :-)

Matt-Yorkley:
> I have good reasons to try to push this forward

Agreed! :smile:

OK, let's go with this for now, and we can look at changing the precompiling strategy as a separate issue? I think it'll need some time, testing, and changes to both repos...

luisramos0:
ok, so this is ready for dev testing.
Someone should try to deploy with this somewhere. I can go for it at some point soon.

Time to lock a staging server for rails 4. I'll get that conversation going in slack.

Matt-Yorkley:
I did test this previously: https://github.com/openfoodfoundation/ofn-install/pull/573#issuecomment-592939843

Looks like we're going for UK staging. Maybe we can merge this and move on to: https://github.com/openfoodfoundation/openfoodnetwork/issues/4900

luisramos0:
:+1:

Matt-Yorkley:
We can probably remove this conditional soon anyway... :smile:

PR #576, opened 2020-02-25
Add new dependencies for brotli module

mkllnk:
This came up when provisioning the new Aus production server initially.

### DIFF
--- a/roles/brotli_nginx/defaults/main.yml
+++ b/roles/brotli_nginx/defaults/main.yml
@@ -10,6 +10,8 @@ nginx_brotli_build_deps:
   - dpkg-dev
   - build-essential
   - git
+  - libgd-dev
+  - libgeoip-dev
   - libpcre3
   - libpcre3-dev
   - zlib1g
--- a/roles/brotli_nginx/tasks/build.yml
+++ b/roles/brotli_nginx/tasks/build.yml
@@ -60,6 +60,12 @@
   register: nginx_modules_built
   become: yes
 
+- name: create modules path
+  file:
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
I did a fair bit of investigation with the nginx issues encountered during the recent provisioning.

I think it came down to the fact that the image had pre-installed packages that were out of date, which meant an older version of nginx was present before the nginx playbook ran, which meant the version on the server was not the expected version.

If an Ubuntu 18 image that doesn't have pre-installed packages is used, then the playbooks run fine. We should to look at how to avoid these issues in the future though.

One option is to define a fixed nginx version number and ensure that version is installed.

Another option is to add some tasks to the `setup.yml` playbook to clean up pre-installed packages. This playbook is usually only run once (when a new server is being set up). We could purge any possible unwanted packages like apache servers at this point, and also remove any pre-installed versions of nginx.

sauloperez:
we should aim for a plain simple ubuntu 18 image without extra things preinstalled. Otherwise, IMO the best manageable way to solve this is to specify the versions of the packages (we really should to avoid surprises).

Matt-Yorkley:
The extra dependencies and creation of the additional `/modules` directory in this PR are not needed if the latest version of nginx is installed (1.16.1). They are only needed for the outdated version 1.14.1.

mkllnk:
Interesting. Unfortunately, we can't choose the Ubuntu image.

Uninstalling unwanted packages sounds like a good idea to me.

Matt-Yorkley:
Maybe we can do both :+1:

PR #581, opened 2020-03-16
Add configs for using lexi (dev testing with LXD)

Matt-Yorkley:
Adds configs for using `lexi` for testing ofn-install. Uses Linux Containers via LXD.

See: https://github.com/libre-ops/lexi for installation and usage (it's really good...)


Matt-Yorkley:
I'll be using this for `ofn-install` testing from now on :muscle:

PR #582, opened 2020-03-26
Enable French and English in Katuma

sauloperez:
A new hub that trades with French customers asked Katuma for these languages.


sauloperez:
Already provisioned and working.

PR #583, opened 2020-03-27
Enable apm for UK and AU

sauloperez:
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/5085

Since UK was the first instance to enable this service, and disable it later on, its settings are still in `ofn-secrets` but AU needs openfoodfoundation/ofn-secrets#29.

I removed old disabling commands from staging servers, that I can confirm, don't have the DD agent running. I manually expired their API keys to be extra sure. 

I also disabled it for Katuma production since we don't check its monitoring and AU needs it much more than Katuma.


sauloperez:
Ok, things are working for AU. We already have metrics :tada:

Issue #442, opened 2019-05-21
Switch to Ansible 2.9


Matt-Yorkley:
Ansible 2.8 and 2.9 have been released, and they include a lot of new features and modules. It also changes and deprecates various bits of syntax, which apparently breaks some of our playbooks.

We should update the syntax and all switch to ~~2.8~~ 2.9 at some point.

- [x] Make a branch that runs the Travis build with Ansible 2.9
- [x] Make sure all playbooks are tested carefully (not just in the CI build)
- [x] Update/fix any syntax that needs it
- [x] Make sure everyone in the team updates their Ansible version to 2.9

Porting guide: https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.9.html

sauloperez:
In particular, https://docs.ansible.com/ansible/latest/modules/postgresql_query_module.html#postgresql-query-module would be a great addition.

PR #584, opened 2020-03-27
Do not become_user: postgres when including role

sauloperez:
This fixes the error:

```
ERROR! 'become_user' is not a valid attribute for a IncludeRole
The error appears to be in '/home/pau/dev/ofn-install/playbooks/db_integrations.yml': line 36, column 7, but may
be elsewhere in the file depending on the exact syntax problem.
```

This is consistent with the `include_role` calls we have in `roles/db_server/` and works I managed to run this playbook against AU production.

### DIFF
--- a/playbooks/db_integrations.yml
+++ b/playbooks/db_integrations.yml
@@ -34,8 +34,6 @@
       when: db_integrations is defined
 
     - name: Reconfigure postgres
-      become: yes
-      become_user: postgres
       include_role:
         name: geerlingguy.postgresql
         tasks_from: configure


playbooks/db_integrations.yml
```diff
@@ -34,8 +34,6 @@
       when: db_integrations is defined
 
     - name: Reconfigure postgres
-      become: yes
```
sauloperez: The playbook is already defined with `become: yes`.

Matt-Yorkley:
I've never seen that before... Have you edit updated your Ansible version to > 2.7 ?

Matt-Yorkley:
We test that playbook in the CI build as well...

sauloperez:
I'm running on Ansible 2.9.6, that must be why.

Matt-Yorkley:
Yep, there's an ofn-install issue for upgrading Ansible syntax, I think other things could be broken as well if using newer Ansible versions.

sauloperez:
well, so far so good. I've been using this version for quite some provisionings and deploys.

Matt-Yorkley:
Great. We'll need to check all playbooks before we close the issue though, and preferably we should all have the same version installed locally. It would be great to move to 2.9!

sauloperez:
> we should all have the same version installed locally

IMO we should use Pyenv as explained in https://gitlab.com/coopdevs/pymasmovil/#python-version

> We'll need to check all playbooks before we close the issue though

Do you want me to make CI run on Ansible 2.9.6? I'm quite sure it'll work. We might need to tweak similar things on playbooks we don't use regularly but https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.9.html should help on that.

Matt-Yorkley:
Sounds good. There's this issue already: #442

sauloperez:
Closing this in favour of #586

PR #585, opened 2020-03-31
Enable use of separate API key for javascript Bugsnag calls

Matt-Yorkley:
Twinned with: https://github.com/openfoodfoundation/openfoodnetwork/pull/5113

Allows configuration of a separate Bugsnag API key for javascript errors per-instance.


Matt-Yorkley:
I removed a superfluous blank line from the commit with that force-push :point_up:

PR #589, opened 2020-04-06
Kill Unicorns

Matt-Yorkley:
- Enables `unicorn-worker-killer` on all staging and production servers.
- Adds a missing lograte entry for exponential unicorn log
- Improves how ENV vars passed to unicorn service

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -141,7 +141,19 @@ logrotate_applications:
           - "{{ current_path }}/log/*.log"
         options:
           - weekly        # rotate weekly
-          - rotate 4     # delete logs older than 4 rotations
+          - rotate 4      # delete logs older than 4 rotations
+          - compress      # compress rotated logs
+          - missingok     # ignore missing logs
+          - notifempty    # don't rotate if empty
+          - copytruncate  # copy the log, then empty the active logfile
+
+  - name: unicorn
+    definitions:
+      - logs:
+          - "{{ shared_path }}/log/unicorn.log"
+        options:
+          - weekly        # rotate weekly
+          - rotate 4      # delete logs older than 4 rotations
           - compress      # compress rotated logs
           - missingok     # ignore missing logs
           - notifempty    # don't rotate if empty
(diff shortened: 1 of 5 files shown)


inventory/group_vars/all.yml
```diff
@@ -168,6 +180,8 @@ unicorn_pid: "{{ pid_path }}/unicorn.pid"
 unicorn_sock: "{{ sock_path }}/unicorn.{{ app }}.sock"
 unicorn_workers: 2
 unicorn_timeout: 30
+unicorn_env_vars: |
+  KILL_UNICORNS=true
```
Matt-Yorkley, line 184: Enable `unicorn-worker-killer` with the default settings on all servers (only applies to prod and staging)

inventory/host_vars/openfoodnetwork.org.au/config.yml
```diff
@@ -33,3 +33,8 @@ custom_application_env_vars: |
   OFN_FEATURE_CONNECT_AND_LEARN: "true"
 
 enable_rails_apm: "true" # has to be explicitly defined as a string due to some datadog bug
+
+unicorn_env_vars: |
+  KILL_UNICORNS=true
+  UWK_MEM_MIN=1900
+  UWK_MEM_MAX=2100
```
Matt-Yorkley, line 40: Apply looser settings to Aus production for now, as requested by Maikel here: https://github.com/openfoodfoundation/openfoodnetwork/pull/5162#pullrequestreview-387969686

roles/webserver/templates/unicorn.service.j2
```diff
@@ -7,7 +7,7 @@ Description=Open Food Network unicorn server
 Type=forking
 User=openfoodnetwork
 WorkingDirectory={{ current_path }}
-Environment=RAILS_ENV={{ rails_env }}
+EnvironmentFile={{ shared_path }}/unicorn_environment
```
sauloperez, line 10: :ok_hand:

Matt-Yorkley: As a side note here, we define some ENV vars for the openfoodnetwork user in `/etc/default/openfoodnetwork`, but as far as I can tell they don't get refreshed when they're re-provisioned. We should re-assess that soon and stop doing it / improve it. It's a separate issue though.

luisramos0: I'd add a comment to this line briefly explaining why RAILS_ENV doesnt work.

Matt-Yorkley:
Tested on `fr-staging` with this and the associated PR. It's working as intended :+1:

PR #591, opened 2020-04-06
Configurable DB pool size and set uk-prod value to 10

kristinalim:
### Description

- Relates to https://github.com/openfoodfoundation/openfoodnetwork/pull/5174

This allows setting the max size of the DB pool for the Rails application, and then increases the value for `uk-prod` from the default of 5 connections to 10.

> From what I understand, it can result to Rails processes waiting for each other to complete, while the DB server can take more simultaneous connections.

This is to remove a possible bottleneck in relation to recent performance issues.


sigmundpetersen:
@kristinalim I guess this need a rebase?
Also is there anything to test on this, a `dev test` perhaps?

Matt-Yorkley:
I resolved the conflict here. I've just done some more reading on this and I'm not sure we need to set it to 10 for UK:

> In Unicorn each process establishes its own connection pool, so you if your db pool setting is 5 and you have 5 Unicorn workers then you can have up to 25 connections. However, since each unicorn worker can handle only one connection at a time, then unless your app uses threading internally each worker will only actually use one db connection.

sauloperez:
well, then that's a confirmation that this won't change anything, right? shall we close this?

kristinalim:
Sorry I wasn't able to rebase this earlier!

Nice finding, @Matt-Yorkley! Yes, let's close this PR.

PR #592, opened 2020-04-06
Remove UK swap from configs

Matt-Yorkley:
The swap has already been removed by Pau, we're just making sure it doesn't come back during next provision.

### DIFF
--- a/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
@@ -7,8 +7,6 @@ rails_env: production
 unicorn_workers: 4
 unicorn_timeout: 720
 
-swapfile_size: 1G
-
 certbot_domains:
   - www.openfoodnetwork.org.uk
   - openfoodnetwork.org.uk


inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
```diff
@@ -7,8 +7,6 @@ rails_env: production
 unicorn_workers: 4
 unicorn_timeout: 720
 
-swapfile_size: 1G
-
```
sauloperez, line 11: but will that run `swapoff` and remove all the stuff?

Matt-Yorkley: You already removed the swap, right?

This Ansible role is not capable of removing or decreasing swap... :disappointed:

Matt-Yorkley:
Merging, so we don't provision and re-add the swap... :+1:

inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
sauloperez: Yes, I manually did a swapoff. Could you please add an issue for it?

PR #594, opened 2020-04-10
Reload unicorn on deploy instead of restart

Matt-Yorkley:
This change means that if a unicorn worker is part-way through serving a request at the point of deployment, it will be restarted as soon as it's finished instead of killing the request.

Also adds a quick check for cases where unicorn is not running yet (on first deploy for example) which would otherwise throw an error (you can't reload a service that hasn't been started yet).

### What should we test?

The Travis build does two deployments, but we should do a quick deploy to staging as well.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -71,7 +71,7 @@
   tags:
     - bundle_app
     - skip_ansible_lint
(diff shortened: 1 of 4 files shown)


Matt-Yorkley:
We added `reload` support to the unicorn service ages ago, but we haven't been using it in deployments... :see_no_evil:

Issue #575, opened 2020-02-24
App migration to new server takes 3 hours


mkllnk:
## Description

Matt moved the Australian OFN to a new server (#565). After preparing everything and transferring assets, the actual move caused a downtime of 3 hours. We can time it so that it's happening during the night but it's still very slow.

## Expected Behavior

The migration takes less than 20 minutes. I've done zero-downtime migrations before. Having a small downtime is much easier but doesn't need to take much longer than 5 minutes. It's only copying the database and restarting two services.

## Actual Behaviour

The migration took 3 hours.

## Steps to Reproduce

I can't find a description of @Matt-Yorkley's migration process. Can you point me to it?

I guess he did something like this:

1. `ansible-playbook playbooks/maintenance_mode.yml -l prod2.openfoodnetwork.org.au`
2. `ansible-playbook playbooks/db_backup.yml -l prod2.openfoodnetwork.org.au`
3. `ansible-playbook playbooks/db_restore.yml -l prod3.openfoodnetwork.org.au`
4. Edit nginx config on the old server to proxy to the new server.

And there were probably many other tasks done as well.

## Severity

bug-s3: a feature is broken but there is a workaround

Two ways at looking at it:
- The migration is a feature and I would say that 3 hours is too long (broken). The workaround is scheduling it very well and telling everybody beforehand.
- During those 3 hours everything is broken. The workaround is waiting a long time and scheduling your work so that it doesn't impact you too badly.

## Possible Fix

Transfer the database between servers directly which takes 2 minutes:
```sh
ssh offrance@openfoodfrance.org "pg_dump -h localhost -U ofn_user openfoodnetwork | gzip" > backup_20180809.sql.tgz
psql -h localhost -U ofn_user postgres -c "drop database openfoodnetwork"
psql -h localhost -U ofn_user postgres -c "create database openfoodnetwork"
zcat backup_20180809.sql.tgz | psql -h localhost -U ofn_user openfoodnetwork
```

The new server needs SSH access to the old server for this to work. A full description of my last migration:
https://community.openfoodnetwork.org/t/migrating-a-production-server/1271/2

Matt-Yorkley:
The migration was really slow (and stressful), and by the time it finished it was really late in my timezone. A couple of smaller factors:

- Ansible tasks seem to run a lot slower in the Rimu servers in general, I'm not sure why?
- The database was 400MB

[...]

PR #596, opened 2020-04-11
Ansible tuning

### LINKED ISSUE #588: Speed up Ansible
@kristinalim recently pointed out Ansible's pipelining feature that could help to speed it up, especially when deploying to a server on the other side of the world adds significant latency. 

There are several other things we could do to significantly increase the speed of some of our Ansible operations, so we can:

- reduce potential disruption during operations
- reduce downtime during operations that require the site to be down
- reduce the time it takes sysadmins to do basic operations like deploying

Further reading here, if anyone's interested: https://shadow-soft.com/turbo-charge-your-ansible/

These changes are really easy to implement.

Matt-Yorkley:
Closes #588 
Related to: #575 

Adds some enhancements to Ansible itself and to some playbooks:

**Enable Ansible's "Pipelining" feature**

Reduces the number of active connections and the amount of data transferred by Ansible when it runs tasks. This should reduce latency issues during trans-atlantic playbook execution

**Increase Ansible forking maximum**

Maximum number of forking processes Ansible will create locally when running playbooks against multiple servers. The default is 5. We regularly run playbooks on 8 servers at once (all-prod), so this should help speed things up.

**Use "free" strategy in provision and deploy playbooks**

When running playbooks on multiple servers, Ansible will normally wait until every server has completed a task before moving to the next one. Using the "free" strategy in a playbook disables this behaviour; tasks will proceed on each server as quickly as they can.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,6 +1,7 @@
 [defaults]
 roles_path = ./community:./roles
 inventory = ./inventory/hosts
+forks = 10
 
 callback_whitelist = profile_tasks
 
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -7,4 +7,5 @@ callback_whitelist = profile_tasks
 
 # Needed to resolve a problem with task output
 [ssh_connection]
+pipelining = True
 control_path = %(directory)s/%%h-%%r
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -11,13 +11,11 @@
 
   tasks:
     - name: Check we're not deploying master to production
-      pause:
-        prompt: |
-          WARNING: The current settings will deploy `master` branch to a production server!
-          Check for typos in the command if deploying a release, otherwise type YES to proceed
+      fail:
+        msg: |
+          WARNING: The current settings would deploy `master` branch to a production server!
+          Check for typos in the command if deploying a release.
       when: inventory_hostname in groups['all-prod'] and git_version == "master"
-      register: deployment_check
-      failed_when: deployment_check.user_input != "YES"
 
     - block:
       - include_role:
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -1,6 +1,7 @@
 ---
 - name: deploy
   hosts: ofn_servers
+  strategy: free
   remote_user: "{{ user }}"
   become: yes
   become_user: "{{ unicorn_user }}"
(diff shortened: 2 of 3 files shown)


ansible.cfg
```diff
@@ -1,10 +1,12 @@
 [defaults]
 roles_path = ./community:./roles
 inventory = ./inventory/hosts
+forks = 10
```
luisramos0, line 4: I read on the webpage you shared
"If you are orchestrating your playbook across many servers, then you may have noticed that the number of servers where the playbook executes simultaneously is 5."
but this is not true for us right? I think the deploy command runs on all our 8 servers at once, right?

playbooks/deploy.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - name: deploy
   hosts: ofn_servers
+  strategy: free
```
luisramos0, line 4: this one is going to be awesome!!!

ansible.cfg
Matt-Yorkley: I'm not 100% sure of the details. I think each "task" Ansible executes actually involves several smaller operations behind the scenes. I assume they're not performed in a totally linear way when interacting with multiple servers. :man_shrugging: 

I'm provisioning `all-prod` now with forks set to 10, and I can see a lot of Ansible activity in `htop`. As far as I can tell there's lot of little processes reading from the playbooks, doing thing with ssh, and interacting with python modules locally. 

I think the 10 forks will help :+1:

luisramos0: I see, cool!

Matt-Yorkley:
I did some testing as I went, but it would be good if someone else can try it out. 

I think there's some new configs in `ofn-install` that aren't on all staging servers yet, so maybe someone could merge this and then provision `all-staging` as a final test?

luisramos0:
I can do that. any expected behaviors to test or just a green provisioning?

sauloperez:
I'd say green is enough as we are not changing the tasks themselves.

Matt-Yorkley:
I did try this out whilst working on it. I think if we can provision and deploy multiple servers at the same time (like all-staging) and there's no error or failures, it's good to go.

luisramos0:
I provisioned all-staging, looks good. I could see staging AU taking an extra minute compared to the others (more changes to be done) :+1:
I deployed 2.9.5 to ES and FR staging: all good :+1:
Merging.

PR #597, opened 2020-04-13
Check unicorn client connections

Matt-Yorkley:
This setting means unicorn doesn't get hit if the queued connection has been cancelled before it gets to unicorn, i.e. if a user hits the refresh button 10 times in 1 second, don't process each of those requests because most of them are already technically been terminated on the client side.

Details in the original commit notes: https://github.com/defunkt/unicorn/commit/5c700fc2cf398848ddcf71a2aa3f0f2a6563e87b

The notes mention it worked really well on Black Friday, so this should help with our "Black Monday" situations on UK Prod :trollface:

### DIFF
--- a/roles/webserver/templates/unicorn.rb.j2
+++ b/roles/webserver/templates/unicorn.rb.j2
@@ -7,6 +7,12 @@ listen "{{ unicorn_sock }}"
 worker_processes {{ unicorn_workers }}
 timeout {{ unicorn_timeout }}
 
+# Enable this flag to have unicorn test client connections by writing the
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Did anyone test this in staging?

sauloperez:
not really, but well, I trust you and them :trollface:

Matt-Yorkley:
I'll provision this now and deploy to staging and test it a bit more...

Matt-Yorkley:
LGTM :+1:

PR #598, opened 2020-04-13
Split app and database

Matt-Yorkley:
WIP / rough draft


roles/db_restore/tasks/restore_database.yml
```diff
@@ -28,6 +30,7 @@
 - name: restore database from backup
   postgresql_db:
     name: "{{ db }}"
+    login_host: "{{ postgres_login_host }}"
```
sauloperez, line 33: why do we need this argument in this role? can't the playbook calling this role simply specify a different host depending on the type of infra (split or single server) is dealing with?

inventory/group_vars/all.yml
```diff
@@ -94,14 +95,27 @@ postgres_shared_preload_libraries:
...
+postgresql_global_config_options: "{{ postgresql_global_config_defaults + postgresql_global_config_custom }}"
+
+postgres_host_options: "{{ '-h {{ db_host }}' if db_host != 'localhost' else '' }}"
+postgres_login_host: "{{ db_host if db_host != 'localhost' else omit }}"
+postgres_delegate_host: "{{ db_host if db_host != 'localhost' else inventory_hostname }}"
+
+dbserver_certs:
+  - postgresql.crt
+  - postgresql.key
+  - root.crt
```
sauloperez, line 117: what are these? to accept SSL connections?

Matt-Yorkley: Indeed! It's using self-signed certificates with a custom CA root key.

roles/db_restore/tasks/restore_database.yml
Matt-Yorkley: I found if we supply the host, the hba authentication seems to go through TCP instead of direct peer-authentication of the system user, which messes it up in some cases. So if the host is remote, we supply it with `-h <host_ip>`, but if it's not we don't supply the `-h` flag at all, which is better.

Matt-Yorkley:
Some of the delegation bits apparently don't play nice with Travis...

roles/db_restore/tasks/restore_database.yml
Matt-Yorkley: I've simplified a lot of this now

sauloperez:
I share here the ideas that came up in the meeting:

* The two VPS need to be in a private network
* The DB server should only be accessible through SSH from the App server. In other words, you can only get into the DB server from within that network.
* We need to thoroughly check connections from the outside world: zapier, metabase and datadog. Is there any other?

playbooks/postgres_tuning.yml
```diff
@@ -1,13 +1,26 @@
 ---
 - name: PostgreSQL production tuning
-  hosts: ofn_servers
+  hosts:
```
sauloperez, line 3: outside of the scope but I think this should be merged into the provisioning playbook soon. We'll need to parametrize it so that it works for all instances but I think it makes sense not to treat these tasks as an exception anymore.

playbooks/provision_dbserver.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    - meta: flush_handlers # Ensure handlers run successfully before reporting success
+
+    - name: notify slack
+      slack:
+        token: "T02G54U79/BF25P9F7A/DJdtYaLLUpRJPiu72d8NqgGg"
+        msg: 'DB server {{ inventory_hostname }} provisioned'
+        channel: "#devops-notifications"
+        username: "ansible executed by {{ lookup('env','USER') }}"
+      when: inventory_hostname not in groups['local']
```
sauloperez, line 93: why not all this in provision.yml?

inventory/group_vars/all.yml
sauloperez: but weren't using SSL already?

Matt-Yorkley: Mostly we were not using SSL at all, just internal connections with peer authentication. The only case where we currently have connections from an external server is with Zapier, and we lock that down with a strong password and a single IP in the whitelist, for a single db user.

But that's using the generic certificates installed with postgres to handle the SSL, and they're probably not good enough for this use-case. Basically with encryption you have three parts: _Authentication_, _Authorisation_, and _Encryption in Transit_. The generic certs are only really good for the last one in that list. With custom certs we can do all three of those things for the connections between the app server and the db server, which will give vastly improved security.

playbooks/provision_dbserver.yml
Matt-Yorkley: Some of it is also in `provision.yml` (applying new config changes etc), but we have to set up and provision a totally different server here, and we probably want to be able to do that independently of the app server, right?

Matt-Yorkley:
So I've run into some issues with Travis again here. Basically the build runs in a Docker container with a messy environment that looks nothing like our production environment. In this case it doesn't respond well to some of the delegation I've used and I think it's not going to work (just in Travis, it works fine with linux containers for example). I've tried a few workarounds but hit nasty errors every time.

So I think if we split the db, we might need to switch CI provider for ofn-install. :disappointed:

sauloperez:
> So I think if we split the db, we might need to switch CI provider for ofn-install.

Switching from Travis seems like a good idea to me. I've seen it as a stale product for a long time... but I think it'll be more about reconsidering our CI strategy than the underlying runner IMO.

In [TimeOverflow](https://github.com/coopdevs/timeoverflow-provisioning/blob/master/.travis.yml) I ended up provisioning a small VPS after unsuccessfully trying your approach, for instance. I'm sure there are better ways.

PR #601, opened 2020-04-16
Add playbook for database migration with SSH forwarding

Matt-Yorkley:
Closes: #575 

Adds a playbook for migrating data from one server to another, transferring the database dump directly between the two using SSH agent forwarding. 

This should greatly reduce the downtime involved in migrating a server.

### DIFF
--- a/roles/db_transfer/tasks/main.yml
+++ b/roles/db_transfer/tasks/main.yml
@@ -1,0 +1,99 @@
+---
+
+- name: dump the database
+  postgresql_db:
+    name: "{{ db }}"
+    state: dump
+    target: "{{ postgres_dump_path }}"
+  become: yes
+  become_user: postgres
+
+- name: move to admin-owned path # noqa 301
+  command: mv {{ postgres_dump_path }} {{ admin_dump_path }}
+  become: yes
+  become_user: root
+
+- name: switch file ownership to admin
+  file:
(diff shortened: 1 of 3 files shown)


roles/db_transfer/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    owner: postgres
+    group: postgres
+  become: yes
+  become_user: root
+  delegate_to: "{{ groups[rsync_to][0] }}"
+
+- name: close active db connections # noqa 301
+  shell: |
+    psql -c "REVOKE CONNECT ON DATABASE {{ db }} FROM public;
+    SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='{{ db }}' AND pid <> pg_backend_pid();"
```
sauloperez, line 53: do the connections get restored from the Unicorn and DJ afterward? It's worth documenting whether this role requires restarting any services or not.

Matt-Yorkley: Yes, they start up again after. You can't restore a db dump whilst the connections are in use.

sauloperez: Yeah, I know. I just don't want any assumptions to be left undocumented that may lead to guesswork.

PR #609, opened 2020-05-12
Increase Canada timeouts to 360

Matt-Yorkley:
:snail:

### DIFF
--- a/inventory/host_vars/openfoodnetwork.ca/config.yml
+++ b/inventory/host_vars/openfoodnetwork.ca/config.yml
@@ -8,7 +8,7 @@ admin_email: admin@openfoodnetwork.ca
 mail_domain: openfoodnetwork.ca
 
 swapfile_size: 2G
-unicorn_timeout: 120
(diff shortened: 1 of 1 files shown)


luisramos0:
thanks Matt!

tschumilas:
I'm not bugging -  just trying to understand the process.  Can you tell me when this change will be effective?   I have the hub worried she won't be able to run the report tomorrow morning.  It is still timing out -  I just ran it (as hub manager).  Here is the screenshot. 
Bailey's Local Foods - OC Customer Totals report - May 12, 1pm

luisramos0:
That shows us the change went in, it's timing out at 6mins.

The second typical workaround here is to run the report for a shorter time window @tschumilas can you try that? You can even isolate problematic days with lots of data.

tschumilas:
Or I could run all days by groups of producers -  tried that and it worked.  But really we are saying,  even with this increased timeout,  this report won't run as one full report.

PR #610, opened 2020-05-13
Lets hack letsencrypt

Matt-Yorkley:
Prep for https://github.com/openfoodfoundation/openfoodnetwork/issues/5418 (and general improvement for future production server migrations).

Automates a tricky part of the process of migrating a live production server. 

This was quite a messy job before, and a bit hacky. It's critical for doing easy server migrations.

PR #611, opened 2020-05-14
Add config settings for Katuma's server migration

sauloperez:

I dropped the following from the existing config:

* `swapfile_size`: We migrate to have more RAM (8GB) thus, no need for swap
anymore
* `enable_nginx_logging`: Datadog is not enabled so this config is set
but not used.
* `disable_datadog`: The new server doesn't have it so a `true` won't
take in effect here.

I added the extra `certbot_domains` entry as documented in
https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server.

As for the entry `alpha.katuma.org`, the old subdomain we redirect to `app.katuma.org`, we need to decide whether we still maintain it or not. It's likely that we'll take the chance to drop it.

### DIFF
--- a/inventory/host_vars/migration.katuma.org/config.yml
+++ b/inventory/host_vars/migration.katuma.org/config.yml
@@ -1,0 +1,21 @@
+---
+domain: app.katuma.org
+host_id: es-prod
+rails_env: production
+
+admin_email: info@coopdevs.org
+
+mail_domain: katuma.org
+
+certbot_domains:
+  - app.katuma.org
+  - alpha.katuma.org
+  - migration.katuma.org
+
+certbot_cert_name: app.katuma.org
+
+postgres_listen_addresses:
+  - '*'
+
+custom_hba_entries:
+  - { type: hostssl, database: "{{ db }}", user: metabase, address: '167.99.89.242/32', auth_method: md5 }
--- a/inventory/hosts
+++ b/inventory/hosts
@@ -97,9 +97,13 @@ app.katuma.org ansible_host=51.15.136.157
 [es-staging]
 staging.katuma.org ansible_host=51.15.216.10
 
+[es-migration]
+migration.katuma.org ansible_host=116.203.60.113
+
 [es:children]
 es-prod
 es-staging
+es-migration
 
 #------------------------------------------------------------------------------
 # France


inventory/host_vars/migration.katuma.org/config.yml
```diff
@@ -0,0 +1,21 @@
...
+  - alpha.katuma.org
+  - migration.katuma.org
+
+certbot_cert_name: app.katuma.org
+
+postgres_listen_addresses:
+  - '*'
+
+custom_hba_entries:
+  - { type: hostssl, database: "{{ db }}", user: metabase, address: '167.99.89.242/32', auth_method: md5 }
```
sauloperez, line 21: @Matt-Yorkley totally unrelated but at some point, it'd be a good idea to replace this hardcoded IP with a reference to the inventory or a plain var, don't you think?

inventory/hosts
```diff
@@ -97,9 +97,13 @@ app.katuma.org ansible_host=51.15.136.157
 [es-staging]
 staging.katuma.org ansible_host=51.15.216.10
 
+[es-migration]
+migration.katuma.org ansible_host=116.203.60.113
```
sauloperez, line 101: I already added this DNS record :heavy_check_mark: 

```
❯ dig migration.katuma.org

; <<>> DiG 9.11.5-P4-5.1ubuntu2.1-Ubuntu <<>> migration.katuma.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45258
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;migration.katuma.org.          IN      A

;; ANSWER SECTION:
migration.katuma.org.   900     IN      A       116.203.60.113

;; Query time: 187 msec
;; SERVER: 127.0.0.53#53(127.0.0.53)
;; WHEN: Thu May 14 16:49:06 CEST 2020
;; MSG SIZE  rcvd: 65
```

inventory/host_vars/migration.katuma.org/config.yml
Matt-Yorkley: Sounds good. This specific metabase config is a unique case at the moment though, I'm not sure we need to fuss too much over it.

sauloperez:
Yep, this is just temporary. @Matt-Yorkley suggested not to merge this. I don't mind and that's what we'll do, but I'm of the opinion that any change to the servers should be reflected with a commit so we can audit it in the future. Besides, wasn't that the whole point of config management?

luisramos0:
yes, I dont see a problem with this being committed at all, I just want to know what are you planning afterwards :-)
Merge this and then have another PR to remove this and merge it with app.katuma.org. That's sounds ok to me as well :+1:

Matt-Yorkley:
Merging the temporary configs :+1:

Normally I do this on my own so I don't need to share the configs, but now we are both working on it together we need to be in sync. Also we'll be using some other PRs here, like https://github.com/openfoodfoundation/ofn-install/pull/610 and we don't want to make too much of a mess with stashing and switching branches etc.

We'll make a PR to remove this after the migrations :+1:

inventory/host_vars/migration.katuma.org/config.yml
```diff
@@ -0,0 +1,21 @@
...
+rails_env: production
+
+admin_email: info@coopdevs.org
+
+mail_domain: katuma.org
+
+certbot_domains:
+  - app.katuma.org
+  - alpha.katuma.org
+  - migration.katuma.org
```
Matt-Yorkley, line 13: Note for future migrations: this `certbot_domains` bit is the whole point of this temporary config file. We list both the current server's domain(s) and the new server's domain, with the current server at the top of the list. Also note the `domain` variable at the top of the file should match the current server and not the new server we're migrating to.

We can then use the `letsencrypt_proxy` playbook on the current server and then provision the new server, and it'll have a certificate that's valid for both domains even though we haven't done the migration yet :ok_hand:

PR #612, opened 2020-05-15
Rsync asset transfer

Matt-Yorkley:

Adds a playbook for transferring all custom instance assets directly between two servers using rsync and SSH forwarding.

Uses the same transfer method as the (relatively new) `db_transfer` playbook.

Katuma uses local storage for images, and we have to transfer ~1.3GB of assets between the two servers for the migration. So: lets do it in one fast transfer of 1.3GB using the high bandwidth of two web servers, instead of two slow transfers of 1.3GB download and then 1.3GB upload (with crappy bandwidth both ways)... 

:rocket:

### DIFF
--- a/playbooks/transfer_assets.yml
+++ b/playbooks/transfer_assets.yml
@@ -7,6 +7,12 @@
   hosts: ofn_servers
   remote_user: "{{ user }}"
 
+  pre_tasks:
+    - name: ensure rsync target is valid and singular
+      fail:
+        msg: "Rsync target must be a single valid host, e.g. `-e 'rsync_to=uk-staging'`"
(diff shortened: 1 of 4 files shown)


Matt-Yorkley:
Not tested yet! :warning:

Matt-Yorkley:
@sauloperez @luisramos0 I've just added some more commits here :point_up:.  

I've also been testing it in staging and it is now working well. :heavy_check_mark:

Matt-Yorkley:
I guess we can do a good live test for this today:

`ansible-playbooks playbooks/transfer_assets.yml --limit es-prod -e "rsync_target=es-migration"`

PR #613, opened 2020-05-16
Add "install yarn" to the provision playbook and "yarn install" to the deploy playbook

luisramos0:
This installs node and npm using https://github.com/geerlingguy/ansible-role-nodejs
and then uses npm to install yarn.

I tried a few other things but this was the only way to control the exact version of yarn we install. I didnt manage to control exact version of node though. So, with this we have node 14.x and yarn 1.22.4

node --version
v14.2.0
yarn --version
1.22.4

How to test:
- provisioning a server should leave node and yarn available on the command line
- deploy a version should create the node_modules folder - this folder will be empty for now because production mode will not deploy devDependencies. I tested this without the production flag and the modules like karma were deployed :+1:

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -77,7 +77,7 @@
 
 - name: run yarn install
   yarn:
-    path: "{{ current_path }}"
+    path: "{{ build_path }}"
     executable: "{{ npm_prefix }}/bin/yarn"
     production: yes
   become: true
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -78,7 +78,7 @@
 - name: run yarn install
   yarn:
     path: "{{ current_path }}"
-    executable: "/usr/local/lib/npm/bin/yarn"
+    executable: "{{ npm_prefix }}/bin/yarn"
     production: yes
   become: true
   become_user: "{{ unicorn_user }}"
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -73,6 +73,16 @@
     - skip_ansible_lint
   notify: restart unicorn
 
+# Install JS dependencies with yarn
+
+- name: run yarn install
+  yarn:
+    path: "{{ current_path }}"
+    executable: "/usr/local/lib/npm/bin/yarn"
+    production: yes
+  become: true
+  become_user: "{{ unicorn_user }}"
+
 # Update the database
 
 - name: check database status
--- a/roles/yarn/defaults/main.yml
+++ b/roles/yarn/defaults/main.yml
@@ -1,0 +1,4 @@
+npm_prefix: "/usr/local/lib/npm"
+
+node_version: 14.x
+yarn_version: 1.22.4
--- a/roles/yarn/defaults/main.yml
+++ b/roles/yarn/defaults/main.yml
@@ -1,4 +1,0 @@
-npm_prefix: "/usr/local/lib/npm"
-
-node_version: 14.x
-yarn_version: 1.22.4
--- a/roles/yarn/tasks/main.yml
+++ b/roles/yarn/tasks/main.yml
@@ -1,0 +1,17 @@
+---
+- name: install node.js
+  include_role:
+    name: geerlingguy.nodejs
+  vars:
+    nodejs_version: "{{ node_version }}"
+    npm_config_prefix: "{{ npm_prefix }}"
+
+- name: install yarn
+  npm:
+    name: yarn
+    version: "{{ yarn_version }}"
+    global: true
+    state: present
+  environment:
+    NPM_CONFIG_PREFIX: "{{ npm_prefix }}"
+    NODE_PATH: "{{ npm_prefix }}/lib/node_modules"
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -5,6 +5,9 @@
 - src: zzet.rbenv
   version: 3.4.2
 
+- src: geerlingguy.nodejs
+  version: 5.1.1
+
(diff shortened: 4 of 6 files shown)


roles/yarn/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+npm_prefix: "/usr/local/lib/npm"
+
+node_version: 14.x
+yarn_version: 1.22.4
```
luisramos0, line 4: I wonder where's the best place to have these versions. These are equivalent to the bundler version. How does ofn-install know what bundler version to install?

roles/deploy/tasks/deploy.yml
```diff
@@ -73,6 +73,16 @@
     - skip_ansible_lint
   notify: restart unicorn
 
+# Install JS dependencies with yarn
+
+- name: run yarn install
+  yarn:
+    path: "{{ current_path }}"
+    executable: "/usr/local/lib/npm/bin/yarn"
```
luisramos0, line 81: after provisioning, I connect to the server as openfoodnetwork and yarn is on the PATH and I can run "yarn install". I wonder why is this task not picking it up?
Without this executable config it errors out saying it cant find the yarn executable:
`Failed to find required executable yarn in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"`

roles/yarn/defaults/main.yml
Matt-Yorkley: For things that are very specific to a role, we would normally put them in the role's defaults. For global configs like this I think we would often put them in `inventory/group_vars/all.yml`. 

`all.yml` is a bit of a monster, and it would be nice to tidy it up at some point, but not in this PR.

The `Bundler` version is a special case, we dynamically pull it from the `Gemfile.lock` during deployment, in case it changes in the main codebase.

roles/deploy/tasks/deploy.yml
Matt-Yorkley: Ansible executes things in a slightly odd way with Python modules, and some of the environment bits like PATH of the specified user are often not as expected within the context of a task.

Specifying the executable looks fine here :+1:

luisramos0: ok, thanks!

roles/yarn/defaults/main.yml
luisramos0: ok, I rebased the PR and added one new commit to move these variable to all.yml.

luisramos0:
ok, ready for re-review.

roles/deploy/tasks/deploy.yml
```diff
@@ -73,6 +73,16 @@
     - skip_ansible_lint
   notify: restart unicorn
 
+# Install JS dependencies with yarn
```
sauloperez, line 76: why not naming the task below exactly this way? then there's no need for the comment

roles/yarn/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: install node.js
+  include_role:
+    name: geerlingguy.nodejs
+  vars:
+    nodejs_version: "{{ node_version }}"
+    npm_config_prefix: "{{ npm_prefix }}"
```
sauloperez, line 7: [nodenv](https://github.com/nodenv/nodenv) is the Node's equivalent of `rbenv` and we'll need it when upgrading node versions, as we do with Ruby. I don't mind starting with this, but at some point we'll need to manage multiple versions.

roles/deploy/tasks/deploy.yml
luisramos0: I'd always do that! But here I am keeping consistency, the file is organized in groups with comments and if I remove the comment this task would "appear" as part of the previous group.
This is also done below:
image

This is easier to see if you open the file in your IDE and scroll.

roles/deploy/tasks/deploy.yml
```diff
@@ -73,6 +73,16 @@
...
+
+- name: run yarn install
+  yarn:
+    path: "{{ current_path }}"
```
luisramos0: the build is broken with:
TASK [deploy : run yarn install] ***********************************************
fatal: [local_travis]: FAILED! => {"changed": false, "msg": "Package.json does not exist in provided path."}
Monday 25 May 2020  14:44:15 +0000 (0:00:00.261)       0:10:33.399 ************ 

I wonder if I need to run this in the build_path instead.
Trying it in a new commit.

luisramos0: yes, it fixes the build and it works.

luisramos0:
ok, the build is green.
I have tested this last version in staging FR and it's working correctly.

Provisioning will install node and yarn.
A normal deploy will create the node_modules folder (only after https://github.com/openfoodfoundation/openfoodnetwork/pull/5200 we will see something landing in node_modules).
To test that node_modules will be populated I removed the "production" flag from yarn install and I confirmed that the dev dependencies (like karma) do land on node_modules correctly during deploy.

@Matt-Yorkley could you please review this?
Thanks!

luisramos0:
thanks Matt :heart:
I think this must be tested by someone other than me.

sauloperez:
I tested it in katuma staging and all seems good:

```
ofn-admin@staging:~$ node --version
v14.10.0
ofn-admin@staging:~$ yarn --version
1.22.4
```

luisramos0:
The next deployment from ofn-install master will have to be preceeded by provision so that yarn is installed before an (for now) empty yarn deploy.
@mkllnk fyi as you will deploy the next release. Should go smoothly and the empty node_modules folder should be there.
Or to reduce risk we could provision/deploy all servers tomorrow just to get yarn out there.

luisramos0:
Thanks for testing Pau!

PR #619, opened 2020-05-21
Fix /rollback path symlink state

Matt-Yorkley:
With the new deployment changes, the `/rollback` path was being left in a state where it couldn't be overwritten by a new symlink in subsequent deploys (basically left as a non-empty directory).


luisramos0:
Now that the other PR was reverted we need to include it in this one and move this one through the pipe as usual, right?

PR #627, opened 2020-05-28
Restart unicorn when deploying with the old method for the last time

Matt-Yorkley:
Removing the current path for the last time apparently messes with assets a bit in this case. Subsequent deployments are fine, it's just the initial one that needs a restart.


Matt-Yorkley:
Tested in staging :heavy_check_mark:

PR #628, opened 2020-05-28
Fix flaky rollback playbook

### LINKED ISSUE #624: Improve rollback playbook
Possibly due to recent changes in the unicorn service, the `rollback` playbook is a bit flaky. This is a playbook we never really use in production, so we don't live-test it often.

The error is:
```
Database query failed: database "openfoodnetwork" is being accessed by other users
DETAIL:  There is 1 other session using the database
```

This occurs between a task for closing database connections and a subsequent task to drop the database (as part of restoring the backup). The connections sometimes restart before the second task has run, so postgres refuses to drop the database.

#### Possible solutions: 

- Stop `unicorn` and `delayed_job` briefly while running these tasks, then restart them afterwards
- Use a block with retries: if the task fails, try dropping the connections a second time and proceeding

Matt-Yorkley:
Closes #624 

Stops unicorn and delayed_job briefly whilst dropping and restoring a database dump.


roles/db_restore/tasks/restore_database.yml
```diff
@@ -1,5 +1,17 @@
...
+  service:
+    name: "{{ service_name }}"
+    state: stopped
+  loop:
+    - "unicorn_{{ app }}"
+    - "delayed_job_{{ app }}"
+  loop_control:
+    loop_var: service_name
+  become: yes
+  become_user: root
```
sauloperez, line 13: wouldn't it be safer to become_user: "{{ user }}" instead? and it's probably better to do that in the playbook so the role is not coupled to it.

Matt-Yorkley: It doesn't work. We get this error: 
```
failed: [local_travis] (item=delayed_job_openfoodnetwork) => {"ansible_loop_var": "service_name", "changed": false, "msg": "Unable to start service delayed_job_openfoodnetwork: Failed to start delayed_job_openfoodnetwork.service: Interactive authentication required.\nSee system logs and 'systemctl status delayed_job_openfoodnetwork.service' for details.\n", "service_name": "delayed_job_openfoodnetwork"}
```
I'm pretty sure that's why I added this in the first place.

PR #629, opened 2020-05-29
Unicorn assets fix

Matt-Yorkley:
Fixes an obscure bug introduced with Ruby 2.0 and `unicorn reload` via `bundle exec`. This is currently blocking deployments to staging servers. 

TLDR version: we need to use the `--keep-file-descriptors` flag when invoking `bundle exec unicorn` (via the unicorn service).

Further reading here:

https://yhbt.net/unicorn-public/20160317002736.GA26688@dcvr.yhbt.net/
rubygems/bundler#2628 (comment)
https://bundler.io/man/bundle-exec.1.html

:warning: Requires testing and then provisioning all servers.


Matt-Yorkley:
This is now on es-staging. We should be able to deploy there now without seeing the missing assets issue (outdated fingerprints on CSS and JS causing them to not be found).

Matt-Yorkley:
Ok it looks like if we alter the unicorn service file during provisioning we have to do a unicorn restart for the changes to take effect...

Matt-Yorkley:
> Why is this only happening in staging? We have deployed to prod after the zero downtime PR was merged, right?

The first deploy with the new method included a restart, which apparently avoided this bug.

We saw this issue come up multiple times when staging via Semaphore. It actually seemed to be a bit flaky, depending on whether or not the assets had changed between branches?

This fix has been applied to multiple staging servers, and the bug is clearly resolved (lots of deploys since then with no problems), so I think this ready for merging.

Matt-Yorkley:
:warning: We should merge this and provision `all-prod` before the next deployment.

luisramos0:
:+1:

PR #632, opened 2020-06-02
Whenever cleanup adjustment

Matt-Yorkley:
Catches a second case of possible duplicate entries in the crontab.

### DIFF
--- a/roles/deploy/handlers/main.yml
+++ b/roles/deploy/handlers/main.yml
@@ -20,7 +20,7 @@
   shell: "echo '{{ crontab_entries.stdout | regex_replace('# Begin Whenever[\\s\\S.]*# End Whenever.*') }}' | crontab -"
   become: yes
   become_user: "{{ unicorn_user }}"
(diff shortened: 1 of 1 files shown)


roles/deploy/handlers/main.yml
```diff
@@ -20,7 +20,7 @@
   shell: "echo '{{ crontab_entries.stdout | regex_replace('# Begin Whenever[\\s\\S.]*# End Whenever.*') }}' | crontab -"
   become: yes
   become_user: "{{ unicorn_user }}"
-  when: crontab_entries.stdout | regex_search(releases_path)
+  when: ( crontab_entries.stdout | regex_search(releases_path) ) or ( crontab_entries.stdout | regex_search('config/schedule.rb') )
```
sauloperez, line 23: is this due to the new -i (identifier) flag you added?

Matt-Yorkley: No, it's related to the mix of old entries. The -i flag is a definite improvement...

PR #636, opened 2020-06-04
Unicorn reload Gemfile fix

Matt-Yorkley:
Possibly fixes: #633 

Ensure the new unicorn process finds the new Gemfile when changing builds.

See: https://stackoverflow.com/a/16555086

PR #638, opened 2020-06-04
Switch back to restarting unicorn on deploy

Matt-Yorkley:
Closes #636 

Doing a restart seems to be enough to resolve the issue in #636. We can investigate issues with the new reload methods some other time.


sauloperez:
yes, good call :+1: it's a pitty though

PR #645, opened 2020-07-06
Add image settings to ofn-install

luisramos0:
#### What? Why?
To remove the page admin/image_settings/edit from OFN ([this PR](https://github.com/openfoodfoundation/openfoodnetwork/pull/5729)), we need to add an alternative way to change and track these settings.
Some are already in ofn-secrets:
s3_access_key
s3_secret
s3_images_bucket (which is converted to s3_bucket)

To keep consistency, in this PR, we add the 2 remaining S3 entries to ofn-secrets (to the example file) where the other ones are. It looks like this:
s3_headers: "{\"Cache-Control":"max-age=31557600\"}"
s3_protocol: "https"

In this PR we also set defaults for all image settings and copy custom settings from all servers to each server config.yml file.

#### What should we test?
I provisioned staging FR with this PR and I do get the images settings in application.yml as well as the custom ATTACHEMENT_PATH :+1: 

#### Release notes
Changelog Category: Changed
Image settings can now be set through ofn-install.

### DIFF
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -31,8 +31,9 @@ mail_domain: example.com
 #  OFN_FEATURE_CONNECT_AND_LEARN: "true"
 
 # Overridable Image Settings
-# attachment_url:
+# If you update attachment_styles, you will need to regenerate thumbnails. Use rake paperclip:refresh:thumbnails CLASS=Spree::Image
 # attachment_styles:
+# attachment_url:
 # attachment_path:
 # attachment_default_style:
 # attachment_default_url:
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -29,3 +29,10 @@ mail_domain: example.com
 #
 #custom_application_env_vars: |
 #  OFN_FEATURE_CONNECT_AND_LEARN: "true"
+
+# Overridable Image Settings
+# attachment_url:
+# attachment_styles:
+# attachment_path:
+# attachment_default_style:
+# attachment_default_url:
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -29,15 +29,15 @@ google_maps_api_key:
 #s3_headers:
 #s3_protocol:
 
-# Stripe Connect API keys
+# Stripe Connect API keys, set these if you are using Stripe
 # Find these under 'API keys' and 'Connect' in your Stripe account dashboard -> Account Settings
 # Under 'Connect', the Redirect URI should be set to https://YOUR_SERVER_URL/stripe/callbacks (e.g. https://openfoodnetwork.org.uk/stripe/callbacks)
 # Under 'Webhooks', you should set up a Connect endpoint pointing to https://YOUR_SERVER_URL/stripe/webhooks e.g. (https://openfoodnetwork.org.uk/stripe/webhooks)
 # Different keys can be made for testing or live environments
-stripe_client_id: # ca_xxxx
-stripe_instance_secret_key: # sk_xxxx
-stripe_instance_publishable_key: # pk_xxxx
-stripe_endpoint_secret: # whsec_xxxx
+#stripe_client_id: # ca_xxxx
+#stripe_instance_secret_key: # sk_xxxx
+#stripe_instance_publishable_key: # pk_xxxx
+#stripe_endpoint_secret: # whsec_xxxx
 
 # Skylight's API key. This enables performance instrumentation through Skylight. See https://github.com/openfoodfoundation/openfoodnetwork/pull/2070 for details.
 #skylight_authentication: ""
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -21,6 +21,8 @@ google_maps_api_key:
 #s3_access_key:
 #s3_secret:
 #s3_images_bucket:
+#s3_headers
+#s3_protocol
 #s3_region:
 #s3_backups_bucket:
 #s3_backups_region:
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -36,7 +36,6 @@ smtp_password:
 
 # Set these if using Amazon S3 for images
 #s3_images_bucket:
-#s3_region:
 
 # Optional settings for Amazon S3 for images
 #s3_headers:
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -29,17 +29,15 @@ smtp_password:
 # Set these if using Amazon S3 for images or backups
 #s3_access_key:
 #s3_secret:
-
-# Set these if using Amazon S3 for backups
-#s3_backups_bucket:
-#s3_backups_region:
-
-# Set these if using Amazon S3 for images
+# Set this if using Amazon S3 for images:
 #s3_images_bucket:
-
-# Optional settings for Amazon S3 for images
+# Set this if using Amazon S3 for backups:
+#s3_backups_bucket:
+# Optional settings for Amazon S3 for images:
 #s3_headers:
 #s3_protocol:
+# Optional settings for Amazon S3 for backups:
+#s3_backups_region:
 
 # Stripe Connect API keys, set these if you are using Stripe
 # Find these under 'API keys' and 'Connect' in your Stripe account dashboard -> Account Settings
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -13,6 +13,19 @@ secret_token: 'xxxxxx'
 # Create a Google Maps API key and enable Javascript and Places libraries
 google_maps_api_key:
 
+# Mail settings
+mail_host:
+mail_port:
+smtp_username:
+smtp_password:
+
+# Optional mail settings
+## Connection type 'TLS' means STARTTLS usually on port 587
+## Connection type 'SSL' means SSL/TLS usually on port 465
+# mail_secure_connection: 'TLS'
+# mails_from: from_address@example.com
+# mail_bcc: bcc_me@example.com
+
 # Set these if using Amazon S3 for images or backups
 #s3_access_key:
 #s3_secret:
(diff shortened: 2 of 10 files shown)


inventory/host_vars/www.openfoodnetwork.org.uk/config.yml
```diff
@@ -19,3 +19,10 @@ custom_hba_entries:
 
 enable_nginx_logging: true
 enable_rails_apm: "true" # has to be explicitly defined as a string due to some datadog bug
+
+# Images settings
+attachment_url: "/spree/products/:id/:style/:basename.:extension"
+attachment_styles: "{\"mini\":[\"48x48#\",\"\"],\"small\":[\"100x100#\",\"\"],\"product\":[\"240x240\u003E\",\"\"],\"large\":[\"600x600\u003E\",\"\"]}"
+attachment_path: "public/spree/products/:id/:style/:basename.:extension"
+attachment_default_style: "product"
+attachment_default_url: "/spree/products/:id/:style/:basename.:extension"
```
sauloperez, line 28: it's basically Katuma the one that doesn't use S3, right? In that case, I'd like to avoid that much duplication (we have quite a few already), which will ease adding new servers as well. We can just treat Katuma as the single exception and probably move it to S3 soon too. It's worth the simplification it will bring IMO.

tl;dr: I would make these settings the default value in roles/app/templates/application.yml.j2 (I believe all servers except Katuma share the same values) and use the var only if defined (just Katuma will).

luisramos0: I think we should stop using S3 and I dont think S3 should be the default. Relates to https://github.com/openfoodfoundation/openfoodnetwork/issues/5078

Anyway, it's a valid point, I'll fetch all the values from all the servers and see what's repeated. If all the servers have the same values, they should go to roles/app/templates/application.yml.j2 :+1:  I'd do this only to non-S3 specific settings though. I'll let you know when I have all the configs and then we decide, ok?

sauloperez: yes, treating Katuma's as an exception would indeed make things more manageable and easier to migrate from S3 to a better alternative.

luisramos0: ok, this is done in the new commits :+1: please have a look.

luisramos0:
I am testing this with OFN PR that removes image_settings and it is not working. I need to investigate.
Meanwhile, can you have a look @sauloperez ?

roles/app/defaults/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+s3_headers: "{\"Cache-Control":"max-age=31557600\"}"
+s3_protocol: "https"
+
+attachment_url: "/spree/products/:id/:style/:basename.:extension"
+attachment_styles: "{\"mini\":[\"48x48\u003E\",\"\"],\"small\":[\"100x100\u003E\",\"\"],\"product\":[\"240x240\u003E\",\"\"],\"large\":[\"600x600\u003E\",\"\"]}"
```
sauloperez, line 6: why do write `u003E` instead of `>`? I fail to see why that would be problematic

inventory/host_vars/www.openfoodfrance.org/config.yml
```diff
@@ -23,3 +23,6 @@ custom_hba_entries:
 
 enable_nginx_logging: true
 enable_rails_apm: "true" # has to be explicitly defined as a string due to some datadog bug
+
+# Images settings
+attachment_path: "home/openfoodnetwork/apps/openfoodnetwork/current/public/spree/products/:id/:style/:basename.:extension"
```
sauloperez, line 28: This overriden path seems a bit odd. It includes the default `public/spree/products/:id/:style/:basename.:extension` path but prepends it with a relative path from the home folder. I'm not sure it would work. Can't we simply remove it so the default one applies?

roles/app/templates/application.yml.j2
```diff
@@ -24,15 +24,13 @@ ADMIN_PASSWORD: {{ admin_password }}
...
 {% endif %}
+
 {% if s3_access_key is defined %}
 S3_ACCESS_KEY: {{ s3_access_key }}
 {% endif %}
 {% if s3_secret is defined %}
 S3_SECRET: {{ s3_secret }}
 {% endif %}
-{% if s3_region is defined %}
-S3_REGION: {{ s3_region }}
```
sauloperez, line 34: this is fine as long as we change Katuma's secrets to use `s3_backups_region` instead as that it's intended use (I'll do that now).

sauloperez: Done in https://github.com/openfoodfoundation/ofn-secrets/pull/42

roles/app/defaults/main.yml
luisramos0: yeah, I dont know, I just copied what was on the DB. Maybe I can change to > here and it will be converted when saving to db :+1:

inventory/host_vars/www.openfoodfrance.org/config.yml
luisramos0: this is copied from the live DB. I thought you have done this :-) maybe @Matt-Yorkley ?

roles/app/defaults/main.yml
sauloperez: yes, let's just try and see if we can avoid some of this obfuscation.

inventory/host_vars/www.openfoodfrance.org/config.yml
sauloperez: we should give it a try

Matt-Yorkley: Different production servers have different settings here, there's not a single value we can apply to all of them. It's about what setting was used when the server first started using S3, which wasn't always the default. All images for that server will now be stored with the given path. A couple of servers needed a different path prefix after the changes to symlinking in the deployment process...

roles/app/defaults/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+s3_headers: "{\"Cache-Control":"max-age=31557600\"}"
```
Matt-Yorkley, line 2: The build is failing due to this line. I think we either need to escape all of the double quotes, or wrap in single quotes:
```yaml
"{\"Cache-Control\":\"max-age=31557600\"}"
```
```yaml
'{"Cache-Control":"max-age=31557600"}'
```

Matt-Yorkley:
The one useful thing in that page that I guess we will now be missing was a button for a "regenerate thumbnails" action. If any changes are made to the scaling / sizing settings (with the weird > and # symbols for example) the images need to be regenerated. I guess it's not a big deal though.

sauloperez:
Let's copy that feature in a rake task then :ok_hand:

roles/app/defaults/main.yml
luisramos0: This does get to the DB, in DE we have > in AU we have the encoded version.
I removed this now. We can verify while testing the PR.

inventory/host_vars/www.openfoodfrance.org/config.yml
luisramos0: I am not sure why we need the full path here but I am going to keep it as is, it's not really the scope of this PR.

roles/app/defaults/main.yml
luisramos0: :+1: thanks, fixed.

luisramos0:
@Matt-Yorkley the regenerate thumbnails was a message to the user:
`image_settings_warning: "You will need to regenerate thumbnails if you update the paperclip styles. Use rake paperclip:refresh:thumbnails CLASS=Spree::Image to do this."
`I have added a comment to config example so users will know what they need to do if they change the attachment styles.

luisramos0:
This is ready for approval I think.
I have tested it with https://github.com/openfoodfoundation/openfoodnetwork/pull/5729 and it does work, the settings do get to the DB as expected :+1:

luisramos0:
Here we are keeping all settings as they were, all changes can be done at another time, issue for that created here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5911

sauloperez (approved):

luisramos0:
rebased to resolve conflicts.

This needs a second review.

inventory/host_vars/_example.com/secrets.example.yml
```diff
@@ -17,40 +13,45 @@ secret_token: 'xxxxxx'
...
+# Stripe Connect API keys, set these if you are using Stripe
 # Find these under 'API keys' and 'Connect' in your Stripe account dashboard -> Account Settings
 # Under 'Connect', the Redirect URI should be set to https://YOUR_SERVER_URL/stripe/callbacks (e.g. https://openfoodnetwork.org.uk/stripe/callbacks)
 # Under 'Webhooks', you should set up a Connect endpoint pointing to https://YOUR_SERVER_URL/stripe/webhooks e.g. (https://openfoodnetwork.org.uk/stripe/webhooks)
 # Different keys can be made for testing or live environments
-stripe_client_id: # ca_xxxx
-stripe_instance_secret_key: # sk_xxxx
-stripe_instance_publishable_key: # pk_xxxx
-stripe_endpoint_secret: # whsec_xxxx
+#stripe_client_id: # ca_xxxx
```
mkllnk, line 47: I think that this will fail because `application.yml.j2` doesn't have `if ... defined` guard clauses for these.

roles/app/defaults/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+s3_headers: "{\"Cache-Control\":\"max-age=31557600\"}"
+s3_protocol: "https"
+
+attachment_url: "/spree/products/:id/:style/:basename.:extension"
+attachment_styles: "{\"mini\":[\"48x48>\",\"\"],\"small\":[\"100x100>\",\"\"],\"product\":[\"240x240>\",\"\"],\"large\":[\"600x600>\",\"\"]}"
```
mkllnk, line 6: The small `100x100#` setting is actually a bug. It's the Spree default but OFN's design expects `112x112#` for the product listing. We are probably going to change this for the mobile design though.

mkllnk (changes requested):
I think we need to fix the Stripe config. Apart from that it's looking good. Well, I would change the default small image size to help new instances until we fix the problem in the mobile ux update.

inventory/host_vars/_example.com/config.yml
```diff
@@ -29,3 +29,11 @@ mail_domain: example.com
 #
 #custom_application_env_vars: |
 #  OFN_FEATURE_CONNECT_AND_LEARN: "true"
+
+# Overridable Image Settings
+# If you update attachment_styles, you will need to regenerate thumbnails. Use rake paperclip:refresh:thumbnails CLASS=Spree::Image
```
mkllnk, line 34: :bomb: I just read this is in the docs:

> WARNING: regeneration is broken on S3 and will truncate your original images.
https://github.com/thoughtbot/paperclip/wiki/Thumbnail-Generation#generatingregenerating-your-thumbnails

That is true for version 6.1.0 while we use version 3.4.2. I don't know when that bug was introduced but the issue was reported nine months after the last release. So we might be safe for a while.

inventory/host_vars/_example.com/secrets.example.yml
luisramos0: I think this is ok because they are defined as empty in roles/app/defaults/main.yml 
I just tested it in staging katuma by removing all stripe keys from the secrets file, this is how it ends up in the server:
image
:+1:

roles/app/defaults/main.yml
luisramos0: :+1:  agree, here I am just copying what's on the DB to the config file. Fixing this is out of scope for this PR. I created an issue for that: https://github.com/openfoodfoundation/openfoodnetwork/issues/5911

inventory/host_vars/_example.com/config.yml
luisramos0: it's good to know :+1:

luisramos0:
Hi @mkllnk I think the PR is ok as is, I tested in Katuma staging with no stripe secrets defined and it worked.

mkllnk (approved):

PR #648, opened 2020-08-18
Updated host vars for AU and Katuma

mkllnk:
I decreased the swap size for Australia because we don't need that much swap and we need more disk space.

When I last provisioned Katuma, I also encountered an error trying to remove Datadog even though it's gone already. So I took that one out as well.


sauloperez (approved):
I think we could remove swap altogether. There's plenty of room in the RAM. from the 1st of June until today :point_down: 

Screenshot from 2020-08-26 08-33-56
Screenshot from 2020-08-26 08-35-02

mkllnk:
I like to keep some swap as a warning buffer. It shouldn't be used but we know that something wrong when it's used without the system crashing straight away.

PR #650, opened 2020-08-26
Ofn install cleanup

Matt-Yorkley:
Cleans up temporary dual-purpose tasks that were designed to handle: 

- the transition from Rails 3 to Rails 4
- the transition from the old deployment style to the new style with symlinks

Both of these transitions are complete.

### What to test

I think the build should be enough in this case, we just need to test that deploying works. It no longer works with Rails 3.


sauloperez (approved):
:muscle: now the upgrade is really done :tada:

luisramos0 (approved):

luisramos0:
The build is broken.

I think we should add the ofn PR template to ofn-install so we follow the process. I think this needs What to Test section.
and probably also this is worth Release Notes.

luisramos0 (changes requested):
see comment above.

Matt-Yorkley:
I accidentally deleted a line with an unrelated conditional that was still needed, I've put it back now and the build is green.

luisramos0 (approved):

luisramos0:
I deployed with this branch to es-staging. all good :+1:

PR #651, opened 2020-09-02
Remove image config which is now handled by OFN code

luisramos0:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5729 will define these settings in the code and we will have the same Paperclip styles on all instances.

### DIFF
--- a/roles/app/defaults/main.yml
+++ b/roles/app/defaults/main.yml
@@ -11,7 +11,7 @@ attachment_default_url: "/spree/products/:id/:style/:basename.:extension"
 # - small: used in the FrontOffice: product list page
 # - product: used in the BackOffice: Product Image upload modal in the Bulk Product edit page and Product image edit page
 # - large: used in the FrontOffice: product modal
-attachment_styles: "{\"mini\":[\"48x48#\",\"\"],\"small\":[\"227x227#\",\"\"],\"product\":[\"240x240>\",\"\"],\"large\":[\"600x600>\",\"\"]}"
+attachment_styles: "{\"mini\":[\"48x48#\",\"jpg\"],\"small\":[\"227x227#\",\"jpg\"],\"product\":[\"240x240>\",\"jpg\"],\"large\":[\"600x600>\",\"jpg\"]}"
 
 stripe_client_id:
 stripe_instance_secret_key:
--- a/roles/app/defaults/main.yml
+++ b/roles/app/defaults/main.yml
@@ -4,14 +4,6 @@ s3_protocol: "https"
 
 attachment_url: "/spree/products/:id/:style/:basename.:extension"
 attachment_path: "public/spree/products/:id/:style/:basename.:extension"
-attachment_default_style: "product"
-attachment_default_url: "/spree/products/:id/:style/:basename.:extension"
-# This is where these styles are used in the app:
-# - mini: used in the BackOffice: Bulk Product edit page and Order Cycle edit page
-# - small: used in the FrontOffice: product list page
-# - product: used in the BackOffice: Product Image upload modal in the Bulk Product edit page and Product image edit page
-# - large: used in the FrontOffice: product modal
-attachment_styles: "{\"mini\":[\"48x48#\",\"jpg\"],\"small\":[\"227x227#\",\"jpg\"],\"product\":[\"240x240>\",\"jpg\"],\"large\":[\"600x600>\",\"jpg\"]}"
 
 stripe_client_id:
 stripe_instance_secret_key:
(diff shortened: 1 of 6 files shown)


roles/app/defaults/main.yml
```diff
@@ -3,7 +3,7 @@ s3_headers: "{\"Cache-Control\":\"max-age=31557600\"}"
 s3_protocol: "https"
 
 attachment_url: "/spree/products/:id/:style/:basename.:extension"
-attachment_styles: "{\"mini\":[\"48x48>\",\"\"],\"small\":[\"100x100>\",\"\"],\"product\":[\"240x240>\",\"\"],\"large\":[\"600x600>\",\"\"]}"
+attachment_styles: "{\"mini\":[\"48x48#\",\"\"],\"small\":[\"227x227#\",\"\"],\"product\":[\"240x240>\",\"\"],\"large\":[\"600x600>\",\"\"]}"
```
mkllnk, line 6: Great!

Can we add a comment to describe the usage of these styles?

- mini: used in the backend, bulk product edit and order cycle edit
- small: used in the shopfront
- product: product image upload modal and edit product image page
- large: product modal

I agree that `mini` should be cropped as well. It needs to fit in the table design and shows admins straight away how it will look like in the shop front. 

It's out of scope of this PR but the `product` style doesn't make much sense. It would be great to improve the admin interface with @Erioldoesdesign's help. But I guess that has to go to a wishlist item first.

mkllnk: Shall we convert everything to JPG in this PR as well? I think it would result in much smaller files and regenerating all images takes a whole day (last time I ran it for Australia). I think we have around 10GB in product image data and Australia uses JPG already.

luisramos0: yeah, I added the comments :+1:

I also changed to convert everything to jpg on all servers.

mkllnk (approved):

sauloperez (approved):
We'll need to run `rake paperclip:refresh` after deploying, right @luisramos0 ? or we plan on still use the sizes of the old thumbnails?

sauloperez:
Actually, I don't think there's much we can test here. I leave it up to you @luisramos0

mkllnk:
@sauloperez 
> We'll need to run rake paperclip:refresh after deploying, right

I'm still figuring that out. It's a bit complicated. Maybe I made it complicated. So here are some of the dependencies:

- https://github.com/openfoodfoundation/openfoodnetwork/pull/5968 needs to be tested before merging this pull request to confirm the image sizes.
- This pull request doesn't have any effect until #5729 is merged.
- The file type changes to `jpg` in this pull request breaks images on all servers. Running `paperclip:thumbnail:refresh` after deploying takes many hours, probably a whole day. We don't want broken images in that time. So we still need to find a way to run the refresh process with future styles to prepare for a deploy.
- I [commented on #5729](https://github.com/openfoodfoundation/openfoodnetwork/pull/5729#pullrequestreview-478290528) that we should move these style declarations to the openfoodnetwork source because they depend on the design code there.

In summary, we have three interdependent PRs in two repos at the moment. I'll try to find a way to simplify that because that's blocking mobile right now.

luisramos0:
as I commented on #6014, imo we should merge this PR and also #5729.
After that the task in #6014 can use Spree::Image to fetch the styles configured in the app and do the job based on those styles.

I think https://github.com/openfoodfoundation/openfoodnetwork/pull/5870 is a good candidate to, after all this:
- remove attachment styles from ofn-install
- remove attachment styles from app_configuration
- set attachement styles directly on Spree::Image.attachment here: https://github.com/openfoodfoundation/openfoodnetwork/blob/3a894a1cdb2f40098a025793a10aa44b7ab5ec4b/app/models/spree/image.rb#L9

luisramos0:
I just went for it now.
So, in this PR here I only leave attachment_url and attachment_path in ofn-install and then in https://github.com/openfoodfoundation/openfoodnetwork/pull/5729 I simply set the styles on the attachment property of Spree::Image (no ofn-install or even Spree::Config or even reformat_styles involved!).
Please have a look. I tested this PR with 5729 together and they look good :+1:

daniellemoorhead:
@sauloperez will you have time to test this and the other related PRs this week? The product listings design changes are dependent upon this work happening first, and we've been waiting months for those changes to go live 🙏

luisramos0:
This is a tricky space and there's a mix of PRs from me and Maikel so I think it's a good call to get one of the other devs to test this :+1:

mkllnk:
Since the history of some of these pull requests changed, it's a bit confusing. I looked at this one again and found that there isn't much to test. I deployed to fr-staging for a sanity check and images were still there. :heavy_check_mark: 

This is good to go. I'll merge it.

PR #653, opened 2020-09-09
use US-specific locale for translating strings

andrewpbrett:
This fixes issue 6003 in the [openfoodnetwork repo](https://github.com/openfoodfoundation/openfoodnetwork/issues/6003) from Laurie. 

It should be pretty straightforward; I don't know what the process for testing something like this would look like, if we should deploy it to a staging server first to make sure nothing bombs out (what if I'd mistyped en_US as en_USA for example and the en_USA.yml file doesn't exist, for example?)

The other consideration is whether transifex has been keeping en_US.yml up to date; I did `diff en.yml en_US.yml` and took a quick glance through and nothing jumped out as wrong, mostly swapping "organisation" for "organization" and things like that.

### DIFF
--- a/inventory/group_vars/us.yml
+++ b/inventory/group_vars/us.yml
@@ -3,7 +3,7 @@
 checkout_zone: USA
 country_code: US
 currency: USD
-locale: en
+locale: en_US
 language: en_US.UTF8
 language_packages:
   - language-pack-en-base
--- a/inventory/group_vars/us.yml
+++ b/inventory/group_vars/us.yml
@@ -4,6 +4,7 @@ checkout_zone: USA
 country_code: US
 currency: USD
 locale: en_US
+available_locales: en_US
 language: en_US.UTF8
 language_packages:
   - language-pack-en-base


sauloperez:
we also need to provide `available_locales` AFAIK. As we do in https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/group_vars/fr.yml#L8. This is where the app picks the list of locales from https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/open_food_network/i18n_config.rb#L8.

andrewpbrett:
Good catch, thanks @sauloperez - I think it'd still work without that line, since it looks like the default locale gets added to the list anyway on line 13 of that file. There are a few examples of other instances that don't have the line (AU for example) and I think it's only if there's more than one possibility, like in France, that we need the extra line. 

But in any case, I added the line, in case we ever want to support an additional locale on the USA instance.

sauloperez:
yeah, right, the source locale makes it work without that ENV var.

Matt-Yorkley (approved):
Thanks Andy :+1:

sauloperez (approved):

sauloperez:
Do you want to give to test it somehow @andrewpbrett ? otherwise, I think we can safely merge it.

andrewpbrett:
I think it looks safe enough to me. If you want I could spin up a server real quick and deploy to it (might be a little later today that I could get to it though).

sauloperez:
that sounds like quite a lot of work to me. I'm just merging

andrewpbrett:
It seems that this didn't have the intended effect. Strings are still not being translated using the en_US locale. I looked on the server and the `config/application.yml` file still has the line `LOCALE: en`. We would expect that to be `en_US` based on this change, right?

sauloperez:
It might be a silly question but was the server provisioned?

sauloperez:
> I thought that happened as part of the release process and so these changes would be reflected in the next release.

There's no written rule about that and we stopped doing because we rarely touch ofn-install so we only do it when required.

andrewpbrett:
Actually it seems like we only have some of the right strings showing up. The `config/application.yml` file now has `en_US`, and the `en_US` file has the right strings from Transifex, but ABN, for example, still is not translated correctly.

lauriewayne:
and it seems like sometimes invoices use the transifex strings and sometimes they don't (like for orders, "send invoice" does not seem to, but "print invoice" does?)

sauloperez:
it's all or nothing, however, if your language doesn't have a string translated the fallback (en) will be used. Keep in mind that some strings are redundant because are shown in many places so you might be looking at the wrong one in transifex.

lauriewayne:
We don't have any untranslated strings (that I know of) in Transifex, and there are several places where the translations aren't carrying over as expected. What's the best way for me to help with troubleshooting this? Would a spreadsheet of places where I do not see the translations carry over be more useful than screenshots, or something else?

abn
trans
gst

andrewpbrett:
Another bit of info that may help thinking through this - I set up a staging server and that has the correct strings for ABN, for example: 
Screen Shot 2020-10-15 at 8 22 14 AM

PR #657, opened 2020-09-23
Update US config to allow zapier access

andrewpbrett:
I did some digging through old emails about getting the USA server connected to Zapier. This thread with @Matt-Yorkley in Slack came up: https://openfoodnetwork.slack.com/archives/CJPMYTDJM/p1570784226004600

Which points to this answer on Stack Overflow: https://stackoverflow.com/a/32413170/374873

I looked at our pg_hba.conf file and saw that we're already adding a custom hba entry in the US [config.yml file](https://github.com/openfoodfoundation/ofn-install/blob/0d4924ca6eb2f7811023afc49c2b860c15ca281f/inventory/host_vars/openfoodnetwork.net/config.yml#L27). The answer on SO suggests that another line should be added:

`host    all             all             0.0.0.0/0               md5` 

I do see that there's already this line in the pg_hba.conf file:

`host all all ::1/128   md5`

which seems like it should do something very similar, but my networking knowledge isn't extensive enough to know for sure that it would do exactly the same thing, so I think it'd be worth trying this out.

### DIFF
--- a/inventory/host_vars/openfoodnetwork.net/config.yml
+++ b/inventory/host_vars/openfoodnetwork.net/config.yml
@@ -24,4 +24,5 @@ postgres_listen_addresses:
   - '*'
 
 custom_hba_entries:
+  - { type: host, database: all, user: all, address: '0.0.0.0/0', auth_method: md5 }
   - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }


inventory/host_vars/openfoodnetwork.net/config.yml
```diff
@@ -24,4 +24,5 @@ postgres_listen_addresses:
   - '*'
 
 custom_hba_entries:
+  - { type: host, database: all, user: all, address: '0.0.0.0/0', auth_method: md5 }
   - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
```
Matt-Yorkley, line 28: This existing line should be enough for Zapier access, at least from the Postgres side... I think AWS has some firewall rules enabled by default, maybe that's the culprit?

Matt-Yorkley: Alternatively it could be that we need to run the `db_integrations` playbook to create the relevant postgres user?

Matt-Yorkley: Have you tried tailing the postgres logs to see what it says about the connection attempts?

andrewpbrett: Yeah that was the conclusion I came to based on what you said in that Slack thread (😄), that it's a firewall/routing issue...I'm seeing the same thing you noted: `ping openfoodnetwork.net` can't find the host, `psql -h 52.3.171.208 -U postgres` just hangs and doesn't appear in any logs, `netstat -nlt` shows it listening on the standard port. 

There is a Postgres user already created, so that's not it. I also noticed that the OP in the Stack overflow question said that they got an error message when trying to connect, rather than just a hang. 

This makes me more skeptical that adding this line will work, unfortunately, so we might just have to wait on Zapier until we move off AWS or can get access to see/edit the firewall/security settings through an admin panel. Thanks for taking a look @Matt-Yorkley

Matt-Yorkley: I think there was something similar with the Italian instance on AWS. The firewall defaults were very restrictive...

andrewpbrett:
I'm going to close this until we migrate to AWS, thanks Matt 👍

PR #658, opened 2020-09-24
Add image settings to hosts missing them

mkllnk:
Testing openfoodfoundation/openfoodnetwork#5729 the images were broken and I noticed that the attachment path had changed. I checked all hosts managed by the global community and added the path were it was missing and different to the default.

This should be merged before openfoodfoundation/openfoodnetwork#5729 so that we can test with the right config on any staging server.

Outdated configs were removed to reduce the amount of files which may need adjustments.

### DIFF
--- a/inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
@@ -18,3 +18,6 @@ unicorn_timeout: 120
 # Size in bytes. You can also use units like 1G, 512MiB or 1000KB. See: `man fallocate`
 # The default is `false`, not installing a swapfile.
 swapfile_size: 1G
+
+# Images settings
+attachment_path: "public/images/spree/products/:id/:style/:basename.:extension"
--- a/inventory/host_vars/staging2.openfood.com.au/config.yml
+++ b/inventory/host_vars/staging2.openfood.com.au/config.yml
@@ -1,7 +1,0 @@
----
-
-domain: staging2.openfoodnetwork.com.au
-host_id: au-staging
-rails_env: staging
-
-admin_email:
--- a/inventory/host_vars/staging2.openfoodnetwork.org.uk/config.yml
+++ b/inventory/host_vars/staging2.openfoodnetwork.org.uk/config.yml
@@ -1,7 +1,0 @@
----
-
-domain: staging2.openfoodnetwork.org.uk
-host_id: uk-staging
-rails_env: staging
-
-admin_email: dev.ofnuk@gmail.com
--- a/inventory/host_vars/www.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/www.openfoodnetwork.org.au/config.yml
@@ -1,5 +1,0 @@
----
-
-domain: www.openfoodnetwork.org.au
-host_id: au-prod
-rails_env: production


inventory/host_vars/staging.katuma.org/config.yml
```diff
@@ -11,3 +11,6 @@ users_sysadmin:
   - dani
   - filipe
   - andy
+
+# Images settings
+attachment_path: ":rails_root/public/spree/products/:id/:style/:basename.:extension"
```
luisramos0, line 16: the only difference here is the :rails_root, do we really need this, I dont think so...

inventory/host_vars/staging.coopcircuits.fr/config.yml
```diff
@@ -14,3 +14,6 @@ users_sysadmin:
   - "{{ core_devs }}"
   - paco
   - filipe
+
+# Images settings
+attachment_path: "home/openfoodnetwork/apps/openfoodnetwork/current/public/spree/products/:id/:style/:basename.:extension"
```
luisramos0, line 19: this is the config in fr-prod, do we really need to use the same custom config in staging? I think we can use the default and configure the S3 FR accordingly.

inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
```diff
@@ -18,3 +18,6 @@ unicorn_timeout: 120
 # Size in bytes. You can also use units like 1G, 512MiB or 1000KB. See: `man fallocate`
 # The default is `false`, not installing a swapfile.
 swapfile_size: 1G
+
+# Images settings
+attachment_path: "public/images/spree/products/:id/:style/:basename.:extension"
```
luisramos0, line 23: same here, I'd just use the default (public/spree/products/:id/:style/:basename.:extension) and configure S3 accordingly.

inventory/host_vars/openfoodnetwork.net/config.yml
```diff
@@ -25,3 +25,6 @@ postgres_listen_addresses:
 
 custom_hba_entries:
   - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
+
+# Images settings
+attachment_path: ":rails_root/public/spree/products/:id/:style/:basename.:extension"
```
luisramos0, line 30: this is already the default here:
https://github.com/openfoodfoundation/ofn-install/blob/0d4924ca6eb2f7811023afc49c2b860c15ca281f/roles/app/defaults/main.yml#L7

inventory/host_vars/openfoodnetwork.de/config.yml
```diff
@@ -26,3 +26,4 @@ unicorn_timeout: 360
 
 # Images settings
 attachment_styles: "{\"mini\":[\"48x48#\",\"jpg\"],\"small\":[\"112x112#\",\"jpg\"],\"product\":[\"240x240>\",\"\"],\"large\":[\"600x600>\",\"jpg\"]}"
+attachment_path: ":rails_root/public/spree/products/:id/:style/:basename.:extension"
```
luisramos0, line 29: this is already the default here:
https://github.com/openfoodfoundation/ofn-install/blob/0d4924ca6eb2f7811023afc49c2b860c15ca281f/roles/app/defaults/main.yml#L7

luisramos0 (changes requested):
I did these configs recently, I double checked all the configs again now and I dont think these configs are needed. I explain why in each of them.

Re staging servers, I think we should use the default value and configure S3 according ly so that we have staging servers with vanilla config. I think that is much better.

inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
mkllnk: I like the idea of standardising but since we have non-standard production servers it would be good to also test with non-standard staging servers. When I tested your pull request removing the image settings, au-staging had broken images. That's exactly the code under test. I think that adding this config here is the easiest way to test that it works.

I would still be happy to standardise this later on. I think it will be easier to change this through the config without caching once we have the new code deployed. But if you feel very strongly about it then we can standardise everything now and find a different way to test this on staging.

luisramos0: Hello Maikel, it's not a strong opinion, just an opinion :-)
We can change these settings on staging servers to make the test yes, but I'd keep the configs in staging without all this messy stuff from production. My vote goes to change these things for the test as needed but afterwards keep the staging servers and GH as is without this PR.
Basically, I'd keep it simple in staging and work to standardize the settings in production instead of copying the mess from production.

Maybe another dev can decide @Matt-Yorkley or @sauloperez

mkllnk: Okay, you convinced me that we should standardise most of it. I'm still wondering if it could be good to keep one staging server different, au-staging for example. But as you said, we can change the config during dev tests. It's just a bit more complicated moving images in the bucket as well to see if the test works. My current thought is:

* Sort out all the non-standard configs except au-staging.
* Test openfoodfoundation/openfoodnetwork#5729 with a modified config.
* Then standardise au-staging as well.

In that case we won't need this pull request. I would still leave it open as long as we haven't sorted all the configs. It can be our todo list.

luisramos0: ok, let's do it :+1:

mkllnk: I adjusted all server configs except au-staging.

luisramos0: :+1:

luisramos0 (approved):
:+1:

luisramos0:
I am merging this with only my review to speed up things :+1:

PR #660, opened 2020-10-16
Minor tweak: ignore missing log files for pg queries

Matt-Yorkley:
I noticed some minor errors related to this whilst debugging something else... in most configurations the pg_log file doesn't exist so we should tell logrotate to ignore it in that case.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -171,6 +171,7 @@ logrotate_applications:
         options:
           - daily         # run daily
           - rotate 0      # don't rotate files
+          - missingok     # ignore missing logs
           - nocreate      # don't replace log files
         lastaction:       # delete log files older than 1 day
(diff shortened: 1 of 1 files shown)


luisramos0 (approved):

Matt-Yorkley:
Interesting... we have a failure in the build and don't think it's related to this PR. Apparently `states.yml` is unexpectedly missing? It could affect provisioning.

Matt-Yorkley:
Aha:
```
No such file or directory ... <path>/db/default/spree/states.yml
```

It's moved in master here: https://github.com/openfoodfoundation/openfoodnetwork/issues/6089. It'll need a tweak in ofn-install before we deploy next...

luisramos0:
ah, good finding, the build is doing it's job!!!

sauloperez (approved):

PR #661, opened 2020-10-16
Whitelist HTTP methods

Matt-Yorkley:
Adds HTTP method whitelisting to our nginx configs :muscle:

Any request from a browser or as part of an API call will reasonably declare one of the standard HTTP methods. We get a large volume of requests with no valid HTTP method declared, and they generally look pretty nasty. Examples of the kind of thing that gets sent with no valid HTTP method (and usually no user-agent header either):

Screenshot from 2020-10-16 22-40-07

For anyone that likes trawling log files, here's what all of our nginx traffic from the last 7 days for UK and France looks like when you filter out the requests with valid HTTP methods: [nasty logs](https://app.datadoghq.com/logs?cols=%40response_time%2C%40http.status_code%2C%40http.url_details.path%2C%40http.method%2C%40http.useragent&from_ts=1602279149870&index=&live=true&messageDisplay=inline&query=-%40http.method%3A%28GET%20OR%20POST%20OR%20PUT%20OR%20DELETE%20OR%20HEAD%20OR%20OPTIONS%29%20%40http.status_code%3A400&stream_sort=desc&to_ts=1602883949870)

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -231,10 +231,7 @@ nginx_acme_challenge_config: |+
     root /etc/letsencrypt/webrootauth;
   }
 
-nginx_valid_methods: |+
-  limit_except GET POST PUT PATCH DELETE OPTIONS {
-    deny all;
-  }
+nginx_valid_methods: limit_except GET POST PUT PATCH DELETE OPTIONS { deny all; }
 
 nginx_sites:
   default:
@@ -253,7 +250,6 @@ nginx_sites:
 
   ofn_80:
     - |
-      {{ nginx_valid_methods }}
       listen 80;
       listen [::]:80;
       server_name {{ certbot_domains | default([domain]) | join(' ') }};
@@ -264,12 +260,12 @@ nginx_sites:
       {{ nginx_acme_challenge_config }}
 
       location / {
+        {{ nginx_valid_methods }}
         return 301 https://$host$request_uri;
       }
 
   ofn_443:
     - |
-      {{ nginx_valid_methods }}
       listen 443 ssl http2;
       listen [::]:443 ssl http2;
       server_name {{ certbot_domains | default([domain]) | join(' ') }};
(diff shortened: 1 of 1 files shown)


luisramos0 (approved):
nice!

luisramos0:
the build is broken due to the states changes in the app repo, not related to this PR :+1:

inventory/group_vars/all.yml
```diff
@@ -231,6 +231,11 @@ nginx_acme_challenge_config: |+
     root /etc/letsencrypt/webrootauth;
   }
 
+nginx_valid_methods: |+
+  limit_except GET POST PUT DELETE OPTIONS {
```
sauloperez, line 235: What about PATCH and HEAD? they are also valid and one would expect them to work.

sauloperez (changes requested):

inventory/group_vars/all.yml
Matt-Yorkley: `HEAD` is included with `GET` in the nginx `limit_except` directive :+1: 

I forgot `PATCH` because we don't actually use it. I think it might get more use by Rails 4.2 though, I'll add it.

sauloperez (approved):

Matt-Yorkley:
Provisioned on all staging servers. Not sure how we want to test this one. I've done some basic testing by navigating around the app in Staging. I guess I'll leave it on all staging servers for now and it's be obvious if something breaks? The nginx logs for production show the app only uses these whitelisted methods, so I think we're good...

enricostano (approved):

Matt-Yorkley:
@enricostano! Come stai, amico? :smile:

sauloperez:
If Nginx logs only show these methods there's no reason not to provision. I'm fine waiting until the next deployment with this in staging.

Matt-Yorkley:
I'll merge it for now and we can provision maybe later in the week. There's a couple of other bits in the pipe that will also need a full provision on all servers when they're merged.

PR #665, opened 2020-10-27
Fix broken image paths

Matt-Yorkley:
We had some post-deployment fun this morning with broken image paths on `es-prod` and `fr-prod` due to two separate issues.

The updated values here are what I've already applied in production on both servers to fix them.

I've also removed a value here as we don't need it. See commit messages for details.


sauloperez (approved):

sauloperez:
This is already applied and working in production so I'm merging.

PR #666, opened 2020-10-29
Improve nginx stub_status config

Matt-Yorkley:
This status page is only accessed internally for stats collection. It shouldn't be accessible externally. The original config here meant that nginx was binding on 0.0.0.0 port 81, which is not necessary. External access to the port was producing a 403/Forbidden response (which is not too bad), but it shouldn't leave that port open at all.

### DIFF
--- a/roles/datadog/templates/status.conf.j2
+++ b/roles/datadog/templates/status.conf.j2
@@ -1,10 +1,7 @@
 server {
-  listen 81;
+  listen 127.0.0.1:81;
   server_name localhost;
-
   access_log off;
-  allow 127.0.0.1;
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Screenshot from 2020-10-29 12-58-17

 :fire: :imp: :fire:

luisramos0 (approved):

Matt-Yorkley:
Not sure how to test this one as it only affects instances that currently use Datadog, which is 3 production servers. I did some dev testing and verified that the port is not open after applying these settings.

andrewpbrett (approved):

PR #667, opened 2020-10-29
Add new attachment URLs for all servers using S3

luisramos0:
This PR updates the config for all servers using S3 to store images. The url is now the S3 URL including the bucket name and the region. See original issue https://github.com/openfoodfoundation/openfoodnetwork/issues/4177 and the main PR https://github.com/openfoodfoundation/openfoodnetwork/pull/6258

PR #673, opened 2020-11-20
Fix typo in fr-prod attachment_url

Matt-Yorkley:
Fixing a quick typo in the configs so the poor sucker** that deploys to production next Tuesday doesn't have any nasty surprises with all images suddenly being broken on `fr-prod`.

See https://github.com/openfoodfoundation/ofn-install/pull/667/files for reference.

\** it's me this week :sweat_smile:


andrewpbrett (approved):

Matt-Yorkley:
Note: there isn't an easy way to dev-test this other than the method I just tried:

1. go to French prod and find a product image. Right click and choose "view image".
2. Change the first part of the URL from `https://s3.amazonaws.com/ofn-prod` to `https://ofn-prod.s3.amazonaws.com`
3. Hit enter. You should see the same image loaded correctly.
4. Try it with the `.com` part removed and it definitely does not work.

PR #549, opened 2019-11-28
Provision Ruby 2.3.7

Matt-Yorkley:
Twinned with https://github.com/openfoodfoundation/openfoodnetwork/pull/4518

- Installs Ruby `2.3.7`
- Ensures previous version (`2.2.10`) is installed, for smoother testing process in CI builds

PR #677, opened 2020-12-01
Add ruby 2.4.4 to list of ruby versions so we can gracefully upgrade servers

luisramos0:
This was done for 2.3.7 here: https://github.com/openfoodfoundation/ofn-install/pull/549

This is needed for https://github.com/openfoodfoundation/openfoodnetwork/pull/6441

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -50,10 +50,11 @@ swapfile_size: false
 
 #----------------------------------------------------------------------
 # Rails variables
-ruby_version: 2.3.7
+ruby_version: 2.4.4
 
 ruby_versions:
   - version: 2.3.7
+  - version: 2.4.4
 
 env:
   RAILS_ENV: "{{ rails_env }}"


inventory/group_vars/all.yml
```diff
@@ -50,10 +50,11 @@ swapfile_size: false
 
 #----------------------------------------------------------------------
 # Rails variables
-ruby_version: 2.3.7
+ruby_version: 2.4.4
```
mkllnk, line 53: This doesn't actually have an impact on the server, does it? Shall we just select the last element in `ruby_versions` for this?

mkllnk (approved):

andrewpbrett (approved):

inventory/group_vars/all.yml
Matt-Yorkley: I think this might set the local rbenv version to 2.4.4 which is probably not what we want...

andrewpbrett: That was my first thought too, that this would **set** the ruby version to 2.4.4; but the impression I got from Luis's comment and https://github.com/openfoodfoundation/ofn-install/pull/549 is that it only installs it. But, uh, yeah, we should probably confirm this before merging 😂

luisramos0:
I will leave this one for you guys.
I thought we can install ruby 2.4.4 here. I thought the actually ruby
version is set in the Gemfile. I may be wrong.

inventory/group_vars/all.yml
Matt-Yorkley: Ah yes, maybe it's ok. Let give it a quick try before we merge though.

Matt-Yorkley: I have a feeling that when we merged #549, the two versions were both basically "working" without causing fatal errors (which is not the case here) and we were merging the update to master in the main repo around the same time (which is also not the case here).

Matt-Yorkley:
Alright, I've done various tests and I think this is good to go.

- Provisioned
- restarted unicorn manually
- Deployed the last release
- Deployed #6441
- Deployed the last release again

All looks good :heavy_check_mark:

PR #685, opened 2020-12-17
Add new /shared/files path for uploads

Matt-Yorkley:
Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/6530

Adds a new shared path for files and symlinks it into the build so it's available for uploads. This is needed for T&C PDF uploads, which currently disappear between deploys. This way we'll persist them outside of the build.

We'll need to separately handle migrating previously uploaded PDFs to the new path.


Matt-Yorkley:
Tested and working on UK Staging. Steps followed:

- Provision with this PR
- Deploy with this PR
- Upload a T&Cs PDF
- Deploy again
- The PDF is still there, it hasn't disappeared :heavy_check_mark: :heavy_check_mark:

sauloperez (approved):
:see_no_evil: we missed it while reviewing

sauloperez:
This is tested already so merging

PR #690, opened 2020-12-23
Update security groups to reflect current reality

luisramos0:
I wanted to remove myself from having access to servers.
Then I decided to update a few more users to keep the current global sysadmin group current. It's just a proposal. We can keep as is if anyone disagrees with the changes.


Matt-Yorkley (approved):
Cool. We currently don't have a playbook or automated process for removing old users' keys. Maybe we should make one...

sauloperez (approved):

PR #698, opened 2021-01-26
Decrease Unicorn and Nginx timeout to 60s

sauloperez:
This seems to be Nginx default and a much more sensible value. Having a single user running a request for 9min, putting the system on fire, blocking anyone else from using it, etc. it's far worse for the business than just killing the request. Better to have a pissed off user than the service down.

Metrics show that response times >1min happen around 8 times every couple of days.

Screenshot from 2021-01-26 11-17-41

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -186,7 +186,7 @@ unicorn_log: "{{ log_path }}/unicorn.log"
 unicorn_pid: "{{ pid_path }}/unicorn.pid"
 unicorn_sock: "{{ sock_path }}/unicorn.{{ app }}.sock"
 unicorn_workers: 2
-unicorn_timeout: 30
+unicorn_timeout: 60
 unicorn_env_vars: |
   KILL_UNICORNS=true
 
--- a/inventory/host_vars/openfoodbrasil.com.br/config.yml
+++ b/inventory/host_vars/openfoodbrasil.com.br/config.yml
@@ -7,7 +7,6 @@ admin_email: administrativo@openfoodbrasil.com.br
 mail_domain: openfoodbrasil.com.br
 
 swapfile_size: 2G
-unicorn_timeout: 120
 
 certbot_domains:
   - openfoodbrasil.com.br
--- a/inventory/host_vars/openfoodnetwork.ca/config.yml
+++ b/inventory/host_vars/openfoodnetwork.ca/config.yml
@@ -8,7 +8,6 @@ admin_email: admin@openfoodnetwork.ca
 mail_domain: openfoodnetwork.ca
 
 swapfile_size: 2G
-unicorn_timeout: 360
 
 certbot_domains:
   - openfoodnetwork.ca
--- a/inventory/host_vars/openfoodnetwork.de/config.yml
+++ b/inventory/host_vars/openfoodnetwork.de/config.yml
@@ -22,4 +22,3 @@ swapfile_size: 2G
 #  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 
 unicorn_workers: 2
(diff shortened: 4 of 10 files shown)


sauloperez:
This needs to be agreed with @openfoodfoundation/train-drivers-product-owners as well

RachL:
@sauloperez How many times does these request really end up with the system being down? 

I doubt we do have many pages that load under 1 min for a simple users currently.... I don't think we are ready for the 1 minute threshold, especially in reports. So I agree 9 minutes is too much, but 1 min will not provide solutions to users and support teams.
Can we meet halfway?

Matt-Yorkley:
> response times >1min happen around 8 times every couple of days

That's a lot of requests that might involve something really critical (like a hub getting their packing done or not on delivery day) that would negatively impacted by with these changes.

I totally understand the imperative to lower the timeout ceiling and agree with the reasoning behind it, but I'm not sure how viable this is right now on a practical level.

We could lower some of the larger values a bit (UK for sure)...

sauloperez:
I chose this number based on the metrics (see screenshot in the description), so that's evidence 

> I doubt we do have many pages that load under 1 min for a simple users currently

that this :point_up: we do. It's a few and well spot endpoints that suffer from this. Just so we're on the same page,

> How many times does these request really end up with the system being down?

having the system not serving any new requests by the order of minutes == system crashed, as in the server process literally died. They are the same from user's perspective. That matches [wikipedia's definition](https://en.wikipedia.org/wiki/Downtime) of downtime: _The unavailability is the proportion of a time-span that a system is unavailable or offline_.

When painfully slow requests happen, they drag all the available resources with them and the next requests pile up waiting to be processed or do so also taking minutes. I just wanted to make this clarification.

In any case, you both share a reasonable concern. I checked >2min and things look much better, while still improving the ability of our servers to deal with concurrent requests. See below the number of requests by endpoint that we would kill if we had the timeout at 2min.

Screenshot from 2021-01-27 10-12-07.

This :point_down: brings more insights

Screenshot from 2021-01-27 10-23-09

What's more, we could add a per-endpoint timeout setting at the nginx level but it's not straightforward to do that at Unicorn-level with our setup. We can choose to solve that issue with product import (I have a couple of ideas that would improve perf. without investing too much time on it) first and close this meanwhile. :+1: ?

We should investigate a bit further why is this mostly affecting the UK. Is it related to particularly large hubs/product catalogs unique to the UK or is it because it's still using swap or any other infra-level thing?

Lastly, I believe it's also worth sharing a consistent timeout threshold across instances. One less conditional to take into account while troubleshooting instance's perf. issues.

lin-d-hop:
Decreasing to 60 secs will cause an insane range of reports S1s. Agree with comments above that we can't just slam this limit on.

I would suggest creating a roll out plan with a goal of 60 secs across all instances.
Implement this iteratively.

As a proposal:
1. List the timeouts on all managed servers.
2. Decrease the timeout on the longest server timeout by 60 secs
3. Wait and see if there are issues or complaints for one week
4. Repeat steps 3 and 4 until all servers are on a 60 second timeout.

This reduces the changes of being inundated with S1s when people can't access reports.

Since the s1s that come up will likely be reports related, perhaps we should also wait until the reports refactoring work is in dev ready so that we can prioritise reports based on performance (and API access, but that's another story).

RachL:
Nice proposal @lin-d-hop 👏 I really like that approach.

sauloperez:
ok, fair enough @lin-d-hop . I still think that there's evidence that we can make it to 2min without much trouble.

I just double-checked against Nginx logs, because I believe Datadog just collects trace samples, meaning they are just of subset of all the requests, while all logs are processed.

Screenshot from 2021-01-27 10-55-19

Screenshot from 2021-01-27 10-55-36

numbers seem consistent.

PR #700, opened 2021-02-03
Don't use swap in CA production

sauloperez:
The new CA server has around 5.5Gb of free RAM so there's no point in having any swap, which when used, slows things down a few orders of magnitude.

This will hopefully mitigate https://github.com/openfoodfoundation/openfoodnetwork/issues/6802 but won't solve it.

### DIFF
--- a/inventory/host_vars/openfoodnetwork.ca/config.yml
+++ b/inventory/host_vars/openfoodnetwork.ca/config.yml
@@ -7,7 +7,6 @@ rails_env: production
 admin_email: admin@openfoodnetwork.ca
 mail_domain: openfoodnetwork.ca
 
-swapfile_size: 2G
 unicorn_timeout: 360
 
 certbot_domains:


Matt-Yorkley (approved):
Note: this won't actually _remove_ the swap, that will have to be done manually afterwards.

sauloperez:
right, I was thinking we'll need to restart all services or even deploy again to make sure the new processes don't map to swap but what you mention should happen before that. Do you remember the exact commands to do so?

Matt-Yorkley:
I think we can merge this now, and it'll mean we won't automatically re-add the swap when provisioning in the future :+1:

There's a few steps to actually disabling the swap after that, but I'm not sure they can be done all at once...

TBH, we might be better off just optimising the `swappiness` value instead of removing swap: https://www.howtoforge.com/tutorial/linux-swappiness/

mkllnk:
I don't see any harm in having the swap. If it's not used then it doesn't slow anything down. If, unexpectedly, we need more memory, it's better to slow down than to crash. And by looking at the swap usage, you can easily see if we need to increase the memory or not.

sauloperez:
Yes, sorry I'm not against swap per se. I'm against swap when: a) it's used while there's enough RAM available b) it's used regularly as a substitute for more RAM, as you said. In this case, it was related to a) but I'm not an expert on the topic and I have no idea how to make the system swap less.

PR #703, opened 2021-02-09
Add hosts for usa migration

andrewpbrett:


inventory/hosts
```diff
@@ -108,11 +108,12 @@ uk-staging2
 [us-prod]
 openfoodnetwork.net ansible_host=52.3.171.208
 
-[us-staging]
+[us-migrate]
+migrate.openfoodnetwork.net ansible_host=64.227.102.29
```
sauloperez, line 112: so we'll keep both domains during the migration and then do switch over?

inventory/hosts
```diff
@@ -108,11 +108,12 @@ uk-staging2
...
 
 [us:children]
 us-prod
-us-staging
```
sauloperez, line 115: it doesn't exist anymore? just curious.

inventory/hosts
andrewpbrett: Yeah, I was basing it on this: https://github.com/openfoodfoundation/ofn-install/pull/611/files

inventory/hosts
andrewpbrett: I don't think it ever existed? I made one very briefly but it wasn't needed.

sauloperez (approved):

PR #709, opened 2021-02-13
Update certbot package name

### LINKED ISSUE #702: Certbot installation fails
## Description
On a fresh Ubuntu 16.04 (x64) server on Digital Ocean, I get the following when Ansible tries to install Certbot:

```
'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'certbot=0.31.0-1+ubuntu16.04.1+certbot+1'' failed: E: Version '0.31.0-1+ubuntu16.04.1+certbot+1' for 'certbot' was not found
```

For fun, I also tried a fresh Ubuntu 20.04 (x64) server:
```
fatal: [migrate.openfoodnetwork.net]: FAILED! => {"changed": false, "msg": "No package matching 'python-psycopg2' is available"}
```

And 18:
```
'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'certbot=0.31.0-1+ubuntu18.04.1+certbot+1'' failed: E: Version '0.31.0-1+ubuntu18.04.1+certbot+1' for 'certbot' was not found
```
(same as 16)

I tried doing `sudo apt-get update` and `sudo apt upgrade` but it had no effect.

I found https://github.com/openfoodfoundation/ofn-install/issues/392 and also did `ansible-galaxy install -r bin/requirements.yml --force` locally, but also no effect. 

I also tried a totally different provider (Linode) just in case there was something weird with Digital Ocean or their distro; but got the same message. 

## Expected Behavior
Certbot should be installed

## Actual Behaviour
Installation fails

## Steps to Reproduce

1. Create new server
2. Run `ansible-playbook site.yml` with the host config etc set up
3. 
4.

## Severity
It seems like this would be an S1 equivalent since it prevents installation, but I'm not sure if it's just something with Digital Ocean or my own setup that's causing it. 

## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):

Matt-Yorkley:
Closes #702

Apparently the package has changed, and the old one is no longer present, so it can't be installed correctly on new servers. It's fine on exiting server that have already installed it.

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -83,6 +83,7 @@
         domain_name: "{{ certbot_domains | default([domain]) | join(',') }}"
         letsencrypt_email: "{{ developer_email }}"
         certbot_nginx_cert_name: "{{ certbot_cert_name | default(domain) }}"
+        certbot_version: "0.31.0-2~deb10u1+ubuntu{{ ansible_distribution_version }}.1+certbot+3"
       when: inventory_hostname not in groups['local']
       tags: certbot
 


andrewpbrett (approved):
I didn't think to do this! Instead of cloning the coopdevs role repo and pointing my Ansible to the local version. Good to know.

Matt-Yorkley:
Yeah, in the version we're using (which is slightly behind master) that `certbot_version` variable is defined in a `defaults.yml` file, which means it can be overridden :+1:

PR #714, opened 2021-02-19
Add core-devs to AU staging

andrewpbrett:


inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
```diff
@@ -22,3 +22,6 @@ swapfile_size: 1G
 # Images settings
 attachment_path: "public/images/spree/products/:id/:style/:basename.:extension"
 attachment_url: ofn-staging.s3.us-east-1.amazonaws.com
+
+users_sysadmin:
+  - "{{ core_devs }}"
```
mkllnk, line 27: This shouldn't be necessary. It's in the group already:
https://github.com/openfoodfoundation/ofn-install/blob/2afa38faa5ddebd8da667fd64db04428e389b36f/inventory/group_vars/au.yml#L19-L20

I suspect that the staging server hasn't been provisioned since you joined the team.

mkllnk (changes requested):

inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
andrewpbrett: ah, got it. I looked at https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/host_vars/staging.openfoodnetwork.org.uk/config.yml and noticed that it's defined there, so I put it in the analogous place for AU staging. It looks like that's redundant for UK staging then.

PR #715, opened 2021-02-19
Clear transient data during deployment

Matt-Yorkley:
This happens after the backup of the previous release has been made and before the migration step. Ensures the old data is cleared regularly and that larger migrations will be more efficient.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -95,6 +95,14 @@
     - rake
     - skip_ansible_lint
 
+- name: clear transient data
+  command: bash -lc "bundle exec rake ofn:data:remove_transient_data RAILS_ENV={{ rails_env }}"
+  args:
+    chdir: "{{ build_path }}"
+  tags:
+    - rake
+    - skip_ansible_lint
+
 - name: migrate database
   command: bash -lc "bundle exec rake db:migrate RAILS_ENV={{ rails_env }}"
   args:


andrewpbrett (approved):
👍

roles/deploy/tasks/deploy.yml
```diff
@@ -95,6 +95,14 @@
...
     - skip_ansible_lint
 
+- name: clear transient data
+  command: bash -lc "bundle exec rake ofn:data:remove_transient_data RAILS_ENV={{ rails_env }}"
+  args:
+    chdir: "{{ build_path }}"
+  tags:
+    - rake
+    - skip_ansible_lint
+
```
sauloperez, line 105: why don't we then run it once a week instead of coupling it to the deployment? This will make deployments slower and brings a 2nd way to trigger this, while we might not need it.

If what we want instead is to make sure it's run in all instances soon why don't we just do so manually once?

mkllnk: I see your point. Slowing down the deployment isn't great.

The intention here is to speed up migrations. The best timing for that is to run it just before the migration.

I often find event based triggers better than time based ones. Imagine you randomly deploy at the same time as the cleaning is run. The database should catch that but you could run into a deadlock depending on the migration.

mkllnk (approved):
Do we need a dev test or will the next deploy be a dev test? :stuck_out_tongue:

roles/deploy/tasks/deploy.yml
sauloperez: > Imagine you randomly deploy at the same time as the cleaning is run

but that still happens at 4am in the morning or similar. So it's very unlikely.

sauloperez:
Yep, it worked in staging

Screenshot from 2021-02-24 16-07-50


I still think we're coupling two things that don't need to but I see you all quite convinced, so I merge it.

PR #704, opened 2021-02-10
Enable feature toggles to entire FR instance

sauloperez:
This opens the gates to `:customer_balance` feature toggle for all users in France. Based on https://github.com/openfoodfoundation/openfoodnetwork/pull/6801.

PR #719, opened 2021-03-09
Toggle customer balances for UK

Matt-Yorkley:
Following #704, toggles customer balances for UK prod.

Related to: https://github.com/openfoodfoundation/openfoodnetwork/issues/7058


mkllnk (approved):
Okay, we can do this. It would be interesting to see how much impact this has.

The work is not complete yet, is it? So I think that there will be places (some reports) where user balances would be different.

PR #723, opened 2021-04-03
Add es to available locales in US config

andrewpbrett:
Pretty self-explanatory? :)

### DIFF
--- a/inventory/group_vars/us.yml
+++ b/inventory/group_vars/us.yml
@@ -4,7 +4,7 @@ checkout_zone: USA
 country_code: US
 currency: USD
 locale: en_US
-available_locales: en_US
+available_locales: en_US, es_US
 language: en_US.UTF8
 language_packages:
   - language-pack-en-base


inventory/group_vars/us.yml
```diff
@@ -4,7 +4,7 @@ checkout_zone: USA
 country_code: US
 currency: USD
 locale: en_US
-available_locales: en_US
+available_locales: en_US, es
```
sauloperez, line 7: I keep that `es` but in the future, we may want to use `es_US` instead.

sauloperez (approved):
Indeed, self-explanatory :joy:

andrewpbrett:
Yes! Laurie actually corrected me on that, I misunderstood something she said. Just pushed a fix.

On Sun, Apr 4, 2021, at 00:31, Pau Pérez Fabregat wrote:
> 

> ***@***.**** approved this pull request.

> Indeed, self-explanatory 😂

> 
> In inventory/group_vars/us.yml <https://github.com/openfoodfoundation/ofn-install/pull/723#discussion_r606761354>:

> > @@ -4,7 +4,7 @@ checkout_zone: USA
 country_code: US
 currency: USD
 locale: en_US
-available_locales: en_US
+available_locales: en_US, es

inventory/group_vars/us.yml
RachL: @andrewpbrett @sauloperez I would do right away an es_US : remember that the main menu is made with translations. So the about link in spanish could never go to the USwebsitee and would always link to Katuma's.
We have this in FR right now with the IT translation and it's pretty annoying.

lauriewayne:
Oh I can see where the "about" menu item would for sure be a pain! Hope we'll have a Spanish reviewer soon who can keep up with new terms and it will make sense for us to use es_US. Thank you all! <3

Matt-Yorkley (approved):

PR #728, opened 2021-04-23
Add environment variables for Discourse SSO and alternate geocoder services to servers if present

cillian:
If instances want a SSO integration with Discourse or to use [alternate geocoder services](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/geocoder.rb) we could add these environment variables to servers. Although maybe it's ok for those variables to be just added via `custom_application_env_vars`, feel free to close this if so.

### DIFF
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -83,3 +83,8 @@ smtp_password:
 #
 #db_integrations:
 #  - { user: datadog, state: present, password: "{{ datadog_db_password }}" }
+
+# For using a geocoding service other than Google you will need:
+#geocoder_api_key: pk.xxxx
+#geocoder_service: mapbox
+#geocoder_timeout: 7
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -59,6 +59,16 @@ BUGSNAG_API_KEY: {{ bugsnag_key }}
 BUGSNAG_JS_KEY: {{ bugsnag_js_key }}
 {% endif %}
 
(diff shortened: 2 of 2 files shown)


sauloperez (approved):
I'm not sure we really support Discourse SSO @mkllnk may know best but the GEOCODER env vars are definitely useful. Thanks one more time @cillian !

andrewpbrett (approved):

andrewpbrett:
Should we make a quick dev-test for this @cillian? E.g. provision a server and check that the ENV vars are set?

mkllnk:
> I'm not sure we really support Discourse SSO

Hm, we used that for a while. But it was a bit confusing for people and it's not used any more. Probably something we should remove. I doubt that anyone else is using it...

sauloperez:
That's what I thought. I would discard these Discourse vars before merging this.

cillian:
:wave: I have removed those Discourse variables now. The Ireland instance uses Discourse SSO but it's not critical if that feature is to be deprecated, `custom_application_env_vars` can be used in the meantime anyway.

@andrewpbrett I can do a dev-test but might need you to point me in the right direction. What servers is that normally done on? Or can it be done a personal server?

andrewpbrett:
I just did the following dev test:

Log in to a staging server (FR)
Open a console
`> ENV["GEOCODER_API_KEY"]`
`> nil`
Add `geocoder_api_key` to FR-secrets.yml
Checkout this branch locally in the own-install directory
Provision FR-staging
Open a new console on FR-staging
`> ENV["GEOCODER_API_KEY"]`
`> "test_key"`

👍 

Looks good to me!

PR #729, opened 2021-04-27
Adding the story template to ofn-install

RachL:


.github/ISSUE_TEMPLATE/story-template.md
```diff
@@ -0,0 +1,25 @@
...
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+## Description
+<!-- Describe the story in detail:
+**- As a:** (enterprise user, super admin, user...)
+**- On page:** (provide url of the page you want to modify. If not provide where will be created the new url and the name we want to give it)
```
sauloperez, line 13: there's something on that sentence that doesn't feel correct but I'm not sure. I summon you @Matt-Yorkley I hope you do know :joy:

sauloperez (approved):

.github/ISSUE_TEMPLATE/story-template.md
Matt-Yorkley: Hmm... I'm not sure this template really makes sense in ofn-install. Like `As a:** (enterprise user, super admin, user...)` doesn't really apply to ofn-install issues. Also the `provide url of the page you want to modify` doesn't really make sense either, we don't modify pages in ofn-install PRs. What do you think?

sauloperez: yeah, right. We can totally remove `page` probably replace it with `component`.

RachL: When I'm creating issues in OFN-install, they often have very visible UI aspect : translation, SSL problem etc, that's why I need the template that is on the OFN repo as well. We can replace by component, but sometimes people will link pages. So if it's ok for you that non dev forget to mention the URL they are working on, I'm fine with that :)

PR #733, opened 2021-05-05
Postgres upgrade cannon

Matt-Yorkley:
:warning: Experimental :warning:

Performs a fully automated in-place Postgres cluster upgrade in ~90 seconds.

Includes putting the server into "maintenance mode" for the duration :heavy_check_mark: 

Tested and working with 9.5 -> 10 upgrade and full production data :heavy_check_mark: 

upgrade-cannon


roles/postgres_upgrade/tasks/configure.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: disable old database auto-start
+  lineinfile:
+    dest: "/etc/postgresql/{{ pg_old }}/main/start.conf"
+    regexp: "^auto"
+    line: "manual"
+  become: yes
+
+- name: set new cluster port to 5432
+  lineinfile:
+    dest: "/etc/postgresql/{{ pg_new }}/main/postgresql.conf"
```
sauloperez, line 12: I'm not sure I'm following. When do we install the new PostgreSQL version and so this path gets created? I don't see it.

roles/postgres_upgrade/tasks/configure.yml
```diff
@@ -0,0 +1,36 @@
...
+    force: yes
+  become: yes
+
+- name: restart postgres
+  service:
+    name: postgresql
+    state: restarted
+  become: yes
+
+- name: rebuild stats and indexes # noqa 301
```
sauloperez, line 31: AFAIK this command will only vacuum but not reindex. That would require https://www.postgresql.org/docs/9.3/sql-reindex.html

roles/postgres_upgrade/tasks/configure.yml
Matt-Yorkley: The role tasks are split up a bit into sections. `roles/postgres_upgrade/tasks/install.yml ` is run before this.

roles/postgres_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+  environment:
+    PGDATAOLD: "/var/lib/postgresql/{{ pg_old }}/main"
+    PGDATANEW: "/var/lib/postgresql/{{ pg_new }}/main"
+    PGBINOLD: "/usr/lib/postgresql/{{ pg_old }}/bin"
+    PGBINNEW: "/usr/lib/postgresql/{{ pg_new }}/bin"
+  when: compatibility_check is succeeded
+  register: upgrade_result
+
+- debug:
+    var: upgrade_result.stdout_lines
```
sauloperez, line 53: I guess we want to keep this, right? we'll see `pg_upgrade`'s output, right?

sauloperez (approved):
This is awesome @Matt-Yorkley ! 

I checked https://www.postgresql.org/docs/current/pgupgrade.html and this brings me confidence :sweat_smile: 

> If a future major release ever changes the data storage format in a way that makes the old data format unreadable, pg_upgrade will not be usable for such upgrades. (The community will attempt to avoid such situations.)

alright! time to bring this in the next delivery meeting :100:

roles/postgres_upgrade/tasks/main.yml
Matt-Yorkley: I was thinking of dropping this and just recommending we only run the playbook with the `-vv` option.

roles/postgres_upgrade/tasks/configure.yml
Matt-Yorkley: Yeah maybe the wording is a bit inaccurate :+1:

It rebuilds the internal statistics used for query optimisation (which are wiped out in the upgrade) in several passes: https://www.postgresql.org/docs/10/app-vacuumdb.html

Basically after `pg_upgrade` finishes, it spits out a bash script called `analyze_new_cluster.sh` and recommends you run it, and the script just contains some notes and this line:
```
/usr/lib/postgresql/10/bin/vacuumdb --all --analyze-in-stages
```

So I thought this would be simpler and more explicit that calling he script directly.

Matt-Yorkley:
So we can do some more rigorous tests in staging with production data (I've been using linux containers with prod data), then potentially do the 7 upgrades we need from `9.5` to `10`.

If it's all good we can tweak it and re-test, then do 10 upgrades from `10` to `11` :heart_eyes:

roles/postgres_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
+---
+
+# This view for Datadog references system tables, which blows up if we try to upgrade the cluster
+# with it included. We can safely drop it here and re-add it later via the datadog role.
+- name: drop pg_stat_activity_dd view # noqa 301
+  command: "psql {{ db }} -c 'DROP VIEW IF EXISTS pg_stat_activity_dd CASCADE;'"
```
Matt-Yorkley, line 6: There was a pretty nasty issue related to this view, which wasn't caught by the compatibility checks, and meant the upgrade failed half-way through, leaving both clusters non-operational. :scream:

luisramos0 (approved):
This looks awesome!

Maybe you can add to the guide that you need to stop the webapp before you do this :-)

Matt-Yorkley:
> you need to stop the webapp before you do this

Ah right, it enables maintenance mode to stop incoming connections, but actually doesn't stop unicorn and delayed job. That definitely needs to be added! :100:

sauloperez:
Cool! what do you reckon we should do in terms of testing? Perhaps someone else should give it a try but not before we prioritize this up next?

Matt-Yorkley:
I tried this on `es-staging`. There was an issue with the system locale being `ES`, which mean the new version of Postgres had an unexpected collation by default, which stopped the upgrade. I've updated the playbook to update the new cluster's collation settings before doing the upgrade :+1:

Katuma Staging is now on Postgres 10 :tada:

Matt-Yorkley:
UK and Aus Staging servers are still on Postgres 9.5. If anyone wants to test the playbook on those two servers, the usage is:
```
ansible-playbook playbooks/postgres_upgrade.yml --limit uk-staging
```

mkllnk:
### Testing notes

#### au-staging

Failed:
```
TASK [postgres_upgrade : check upgrade compatibility] *************************************************************************************************************************************************************
Wednesday 09 June 2021  10:21:45 +1000 (0:00:01.753)       0:02:29.748 ******** 
fatal: [staging.openfoodnetwork.org.au]: FAILED! => {"changed": true, "cmd": ["/usr/lib/postgresql/10/bin/pg_upgrade", "--link", "--check", "--old-options=-cconfig_file=/etc/postgresql/9.5/main/postgresql.conf", "--new-options=-cconfig_file=/etc/postgresql/10/main/postgresql.conf"], "delta": "0:00:01.170520", "end": "2021-06-09 00:21:47.474736", "msg": "non-zero return code", "rc": 1, "start": "2021-06-09 00:21:46.304216", "stderr": "", "stderr_lines": [], "stdout": "\nThere seems to be a postmaster servicing the new cluster.\nPlease shutdown that postmaster and try again.\nFailure, exiting", "stdout_lines": ["", "There seems to be a postmaster servicing the new cluster.", "Please shutdown that postmaster and try again.", "Failure, exiting"]}
```

I logged in to see what's happening and it looks like staging-au has postgres versions 9, 10 and 11 installed and running. How do we clean that up?


#### uk-staging

Failed:

```
Performing Consistency Checks
-----------------------------
Checking cluster versions                                   ok
Checking database user is the install user                  ok
Checking database connection settings                       ok
Checking for prepared transactions                          ok
Checking for system-defined composite types in user tables  ok
Checking for reg* data types in user tables                 ok
Checking for contrib/isn with bigint-passing mismatch       ok
Checking for invalid "unknown" user columns                 ok
Checking for roles starting with "pg_"                      ok
Creating dump of global objects                             ok
Creating dump of database schemas
  openfoodnetwork
  postgres
  template1
                                                            ok
Checking for presence of required libraries                 ok
Checking database user is the install user                  ok
Checking for prepared transactions                          ok
Checking for new cluster tablespace directories             ok

If pg_upgrade fails after this point, you must re-initdb the
new cluster before continuing.

Performing Upgrade
------------------
Analyzing all rows in the new cluster                       ok
Freezing all rows in the new cluster                        ok
Deleting files from new pg_xact                             ok
Copying old pg_clog to new server                           ok
Setting next transaction ID and epoch for new cluster       ok
Deleting files from new pg_multixact/offsets                ok
Copying old pg_multixact/offsets to new server              ok
Deleting files from new pg_multixact/members                ok
Copying old pg_multixact/members to new server              ok
Setting next multixact ID and offset for new cluster        ok
Resetting WAL archives                                      ok
Setting frozenxid and minmxid counters in new cluster       ok
Restoring global objects in the new cluster                 ok
Restoring database schemas in the new cluster
  openfoodnetwork

*failure*

Consult the last few lines of "pg_upgrade_dump_33197.log" for
the probable cause of the failure.
Failure, exiting
```

I'm moving this back to dev.

Matt-Yorkley:
Hmmm looking at `pg_upgrade_dump_33197.log` on UK, the issue is with those custom Datadog table views again:

Screenshot from 2021-06-09 09-30-38

I thought that was fixed already, I'll have to take another look.

Matt-Yorkley:
I've resolved that issue with Datadog views and added some improved handling of databases intialized with different locales and re-tested with multiple production database dumps. I think this is ready for use.

Matt-Yorkley:
UK Staging and AU Staging were both upgraded this morning. Both looking good :ok_hand:

andrewpbrett (approved):
> UK Staging and AU Staging were both upgraded this morning. Both looking good

Seems like a sufficient dev-test to me... shall we merge @Matt-Yorkley?

Matt-Yorkley:
> shall we merge

Sure. I think it's ready to go :ok_hand:

PR #734, opened 2021-05-13
Split UK

magec:
### What? Why?

This:

- Adds `./bin/cli` command that will open a container with dependencies (alternative to setup)
- Changes the ansible code so we can have dbservers and webservers separated
- Adds a `terraform/uk` directory with UK infra definition that splits UK into a server and a hosted DB using a VPC
- There were issues while trying to compile brotly, I needed to make several changes in the role regarding to that :man_shrugging: 

### What should we test?

**Manual testing**

A very basic smoke test of the most critical paths: checkout, card payment, email sending, etc. We aim to catch any obvious misconfigured settings.

To test simply navigate to https://159.65.25.250/ (beware, the SSL cert is out of date) and do it from there. There's no DNS record pointing to it yet.

**Documentation**
https://github.com/openfoodfoundation/ofn-install/wiki/UK%27s-split-infrastructure

**Dev testing**

- [x] Ensure backward compatibility of the changed playbooks. Provisioning and deployment should work on staging.
- [ ] Check and test playbooks to migrate production DB to new infra: `db_backup`, `db_restore`, and `db_transfer`.

**Roll out**
- [ ] Stop the app
- [ ] Restore production DB dump
- [ ] Copy the SSL cert from current production
- [ ] Update DNS
- [ ] Provision
- [ ] Regenerate SSL cert to ensure it keeps working
- [ ] Deploy


docker/cli/Makefile
```diff
@@ -0,0 +1,21 @@
+UNAME=$(shell uname)
+mkfile_path := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+DEPENDENCY_FILES := $(mkfile_path)/Dockerfile $(mkfile_path)/../../requirements.txt
```
Matt-Yorkley, line 3: Could you explain a bit why we need to wrap everything up in this Docker container? Is it just for the Terraform CLI?

roles/brotli_nginx/tasks/build.yml
```diff
@@ -47,7 +47,7 @@
   changed_when: False
 
 - name: configure build # noqa 305
-  shell: "./configure {{ nginx_build_args }} --with-compat --add-dynamic-module={{ nginx_brotli_src_path }}"
+  shell: "./configure {{ nginx_build_args }} --add-dynamic-module={{ nginx_brotli_src_path }}"
```
Matt-Yorkley, line 50: We don't have any issues with this on Ubuntu 16 or 18... where you using Ubuntu 20 while testing? Was it a standard DO Droplet?

Matt-Yorkley:
We would need to fix those errors in the Travis build, I'm not sure how much work that would take...

docker/cli/Makefile
sauloperez: Yes, dependencies are listed in `requirements.txt`.

roles/brotli_nginx/tasks/build.yml
sauloperez: I don't remember if we agreed to go straight to Ubuntu 20 but it could as well be the standard.

sauloperez:
I provisioned both Katuma staging and FR staging and both seem to be ok. The most notable changes were:

```
TASK [webserver : provision service files] *********************************************************************************************************************************************************************************************************************
changed: [staging.coopcircuits.fr] => (item={'src': 'unicorn.service.j2', 'dest': '/etc/systemd/system/unicorn_openfoodnetwork.service'})                                                                                                                       
ok: [staging.coopcircuits.fr] => (item={'src': 'unicorn_environment.j2', 'dest': '/home/openfoodnetwork/apps/openfoodnetwork/shared/unicorn_environment'})                                                                                                      
changed: [staging.coopcircuits.fr] => (item={'src': 'delayed_job.service.j2', 'dest': '/etc/systemd/system/delayed_job_openfoodnetwork.service'})                                                                                                               
Friday 21 May 2021  12:39:22 +0200 (0:00:13.267)       0:05:56.741 ************     

TASK [brotli_nginx : install build dependencies] ***************************************************************************************************************************************************************************************************************
changed: [staging.coopcircuits.fr]  

TASK [app : template files] ************************************************************************************************************************************************************************************************************************************
changed: [staging.coopcircuits.fr] => (item={'src': 'application.yml.j2', 'dest': '/home/openfoodnetwork/apps/openfoodnetwork/shared/config/application.yml'})                                                                                                  
changed: [staging.coopcircuits.fr] => (item={'src': 'postgresql.yml.j2', 'dest': '/home/openfoodnetwork/apps/openfoodnetwork/shared/config/database.yml'})        

TASK [zzet.rbenv : install build depends] **********************************************************************************************************************************************************************************************************************
changed: [staging.coopcircuits.fr] => (item=['build-essential', 'curl', 'git', 'libcurl4-openssl-dev', 'libffi-dev', 'libreadline-dev', 'libssl-dev', 'libxml2-dev', 'libxslt1-dev', 'zlib1g-dev'])                                                             
Friday 21 May 2021  12:35:34 +0200 (0:00:05.426)       0:02:08.887 ************                                                                                                                                                                                 
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `pkg: "{{ item }}"`, please use `pkg: ['{{ rbenv_extra_depends }}']` and remove the loop.  
This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.                                                                                                                        
                                                                                                                                                                                                                                                                
TASK [zzet.rbenv : install extra build depends] ****************************************************************************************************************************************************************************************************************
changed: [staging.coopcircuits.fr] => (item=['libssl1.0-dev'])                                                                                                                                                                                                  
Friday 21 May 2021  12:35:40 +0200 (0:00:05.611)       0:02:14.499 ************  
```

playbooks/provision.yml
```diff
@@ -105,6 +103,63 @@
       become: yes
       tags: logrotate
 
+    - role: arillso.logrotate
+      become: yes
+      tags: logrotate
```
sauloperez, line 108: :warning: this is duplicated

roles/app/templates/postgresql.yml.j2
```diff
@@ -33,3 +36,4 @@ production:
   username: {{ db_user }}
   password: {{ db_password }}
   host: {{ db_host }}
+  port: {{ db_port }}
```
sauloperez, line 39: I wish someday we could rely on ENV vars alone and not have this duplicate as a template and as yml in the app :disappointed:

terraform/uk/main.tf
```diff
@@ -0,0 +1,94 @@
...
+resource "digitalocean_project" "uk" {
+  name        = "ofn-uk"
+  description = "Open Food Network UK"
+  purpose     = "Web Application"
+  environment = "Production"
+  resources = flatten([
+    digitalocean_droplet.appservers_production[*].urn,
+    digitalocean_database_cluster.production.urn
+  ])
+}
```
sauloperez, line 94: :heart:

sauloperez:
The build is failing on 

```
TASK [brotli_nginx : add brotli modules to nginx] ******************************
failed: [local_travis] (item=ngx_http_brotli_filter_module.so) => {"ansible_loop_var": "item", "changed": false, "item": "ngx_http_brotli_filter_module.so", "msg": "Destination directory /usr/share/nginx/modules does not exist"}
failed: [local_travis] (item=ngx_http_brotli_static_module.so) => {"ansible_loop_var": "item", "changed": false, "item": "ngx_http_brotli_static_module.so", "msg": "Destination directory /usr/share/nginx/modules does not exist"}
```

This module is becoming tricky :see_no_evil:

sauloperez:
Deployment works as well. I managed to place an order and send its confirmation :heavy_check_mark:

ansible.cfg
```diff
@@ -4,7 +4,7 @@ inventory = ./inventory/hosts
 forks = 10
 
 callback_whitelist = profile_tasks
-
+host_key_checking = False
```
Matt-Yorkley, line 7: Do we need this? :thinking:

Matt-Yorkley:
Do you think the droplet is big enough? The CPU hit 100% during that deployment yesterday, and that's without any real load...

Screenshot from 2021-06-01 09-10-24

sauloperez:
that definitely needs to be looked at. I doubt there was any request at the time of that spike :thinking:

roles/deploy/tasks/rollbacks.yml
```diff
@@ -9,13 +9,15 @@
     target: "{{ rollback_tmp_path }}"
   become: yes
   become_user: postgres
+  when: db_host == 'localhost'
```
Matt-Yorkley, line 12: So we don't make a backup during deployment?

.gitignore
```diff
@@ -7,3 +7,8 @@ inventory/host_vars/*/secrets.yml
 *.log
 **/*.retry
 .idea/
+
+secrets/*
```
Matt-Yorkley, line 11: Are there some additional secrets files here that are needed to use it, and if so where will we store them?

sauloperez: See the documentation :point_right: https://github.com/openfoodfoundation/ofn-install/wiki/UK%27s-split-infrastructure

roles/deploy/tasks/rollbacks.yml
sauloperez: This was totally coupled to dbservers running alongside the webserver so I thought skipping this was a good trade-off. Doing the backup against DO's hosted DB can be explored later on.

Matt-Yorkley:
So... I'm not really sure if we should go ahead with this change.

I feel like this decision was made as a knee-jerk reaction to performance issues on UK Production 3 months ago. At the time things were pretty bad. We had real problems! The database itself wasn't the bottleneck there though. 

If you look at the stats in Datadog for the last month or so you'll see that those problems we had 3 months ago are completely gone. Memory is stable at ~65%. CPU was the main issue previously, and was hitting 90-95%, but now it barely hits ~30% at peak. Response times on some of the worst endpoints are a fraction of what they were before. It's never looked so good!

The problems we had were a combination of some important missing indexes on newly-added database columns (which caused some nasty temporary issues for a couple of weeks, but were then fixed) and the general longstanding issues around callbacks triggering insane queries (which we've now largely fixed, and are continuing to improve). Displacing the database to a separate server wouldn't have actually solved either of these problems, right? Changing the code solved the problems, and it's already done and the results are pretty obvious.

So I'm not sure what problem we're actually solving with this now? I'm not sure we'll see any measurable improvement from this change. It would add a significant amount of additional complexity and infrastructure costs though. We would also be throwing out the sysadmin standardisation we worked hard to establish, by having one server with a wildly different setup to all the others. 

In the future, changes we would want to make in ofn-install would potentially need to be tested on both setups, right? So we would need to have a similar setup on at least one staging server. We also basically ignore this setup in the test suite, that is; it's not feasible to test it.

And... have we tested Zapier really carefully with this? I'm pretty sure the current playbooks we use for setting up additional database users for integrations (including Zapier) will be broken here, so if we wanted to adjust those integrations it would not work. How does Zapier connect to the database if it's on VPC? Are we proxying a connection via the app server? What does that look like and where is it configured?

I also see the DO Managed Databases have regularly scheduled downtime periods which seem to be weekly and last for "up to four hours". Do we want weekly downtime for our database?

This will also break the replication setup (which has been working fine since those last updates a month ago). So no Metabase for the UK, unless we implement a custom solution. With similar issues to the Zapier integration.

I'm just not sure about how much concrete value this will actually provide (if any) compared to the additional complexity and costs it will bring...

roles/datadog/tasks/pg_stats.yml
```diff
@@ -85,34 +111,52 @@
...
+    -d {{ db }} \
+    -c \"SELECT * FROM pg_catalog.pg_views WHERE viewname = 'pg_stat_activity_dd';\""
+  become: db_host == 'localhost'
   become_user: postgres
   register: pg_stat_activity_dd_exist
+  environment:
+    PGPASSWORD: "{{ db_password }}"
   changed_when: False
 
 - name: add pg_stat_activity_dd view
```
Matt-Yorkley, line 150: Has this playbook been re-tested on a server that _doesn't_ have a separate database to make sure it still works?

sauloperez: :point_right: https://github.com/openfoodfoundation/ofn-install/pull/734#issuecomment-845863490

Matt-Yorkley: Right, but this code only gets executed conditionally depending on what configurations the server has, and in the case of those 2 staging servers this specific part will be skipped. I have a suspicion it might not work now...

sauloperez: Possibly. This is still WIP.

Matt-Yorkley:
Previous comments aside, in regards to the **Terraform** part of this PR specifically:

Terraform is a nice tool and declarative infrastructure management is also _really cool_, but... I'm not clear what problem Terraform is solving for OFN here. Essentially it's just adding a complex additional toolchain that just enables the creation of DO droplets. But... that's a job we (realistically) do maybe once every 2 years? And it takes ~2 minutes and a few clicks to log in and create a droplet (or a db server). Also I think not all OFN web hosts will be compatible with Terraform? I doubt Aus production is, for example.

To put it another way: we could drop all of the Terraform parts of this PR, and everything that's being implemented here in regards to the split infrastructure would still work in exactly the same way, right?

Or to put it another way: if five months from now I want to add or update a DO Droplet and I decide to use the Terraform CLI, it would probably take me an hour or so of setting up the Docker container and reading all the Terraform docs, or... I could just log in and click a couple of buttons and it would be done in 2 minutes. So... I think I would choose the second option every time.

Or have I missed something?

mkllnk:
Those are good points. It's great to have the ofn-install part in case we need to split the database at some point but we don't need to do that now. The risks and costs are much larger than any potential performance gains right now.

sauloperez:
Yesterday @lin-d-hop @Matt-Yorkley and had a chat about this and concluded that given I'll be gone by the end of the month there are chances this will be delayed even further, putting yet another burden on the dev team. This decision has been communicated @magec who totally understands it.

We agreed then that this can be used as a basis for future infrastructure improvements when the time comes.

sauloperez:
I also removed the wiki page we created out of https://github.com/openfoodfoundation/ofn-install/pull/743.

PR #737, opened 2021-05-14
Provision Redis

Matt-Yorkley:
Part of https://github.com/openfoodfoundation/openfoodnetwork/issues/7541

- Installs and configures Redis
- Does _not_ remove Memcached (we can do that later, and we need to maintain some backwards-compatibility during the transition period until all servers are updated) :ok_hand: 
- Adds a failsafe check in `deploy.yml` to stop deployment if Redis is in the Gemfile but has not been installed yet (for instances not managed by the core team, so we don't break anything in production) :heart:

### DIFF
--- a/roles/common/templates/redis.conf.j2
+++ b/roles/common/templates/redis.conf.j2
@@ -1,0 +1,4 @@
+# {{ ansible_managed }}
+
+maxmemory 124mb
+maxmemory-policy volatile-lru
--- a/roles/deploy/tasks/check_redis.yml
+++ b/roles/deploy/tasks/check_redis.yml
@@ -14,5 +14,11 @@
   check_mode: yes # Don't actually modify the file
   become: yes
   register: redis_requirement
-  when: "'redis-service' not in services"
-  failed_when: redis_requirement is failed
+  ignore_errors: true
+
+- name: stop deployment if redis is both required and absent
+  fail:
+    msg: |
+      The deployed OFN version requires Redis to be installed.
+      Please provision with the latest copy of ofn-install before deploying.
+  when: "redis_requirement is changed and 'redis-server.service' not in services"
--- a/roles/deploy/tasks/check_redis.yml
+++ b/roles/deploy/tasks/check_redis.yml
@@ -1,0 +1,18 @@
+---
+
+# This check fails if Redis is present in the new build's Gemfile but Redis is not installed yet,
+# which will block the deployment from proceeding (until the server is correctly provisioned).
+
+- name: gather host services facts
+  service_facts:
+
+- name: check redis presence in gemfile
+  lineinfile:
+    dest: "{{ build_path }}/Gemfile"
+    regex: "^gem .*redis"
+    state: absent
+  check_mode: yes # Don't actually modify the file
+  become: yes
+  register: redis_requirement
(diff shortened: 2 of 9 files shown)


roles/deploy/tasks/check_redis.yml
```diff
@@ -0,0 +1,18 @@
+---
+
+# This check fails if Redis is present in the new build's Gemfile but Redis is not installed yet,
+# which will block the deployment from proceeding (until the server is correctly provisioned).
+
+- name: gather host services facts
+  service_facts:
+
+- name: check redis presence in gemfile
```
andrewpbrett, line 9: 💪

andrewpbrett (approved):
Looks good to me - do you want to propose a dev-test on it @Matt-Yorkley? Maybe provision and make sure `which redis` gives a path? And then do more of a dev-test on https://github.com/openfoodfoundation/openfoodnetwork/pull/7648?

Matt-Yorkley:
> Looks good to me - do you want to propose a dev-test on it @Matt-Yorkley? Maybe provision and make sure `which redis` gives a path? And then do more of a dev-test on [openfoodfoundation/openfoodnetwork#7648](https://github.com/openfoodfoundation/openfoodnetwork/pull/7648)?

I've done that already (on UK Staging) and manually set the log level to :debug temporarily to check the output, but yes it would be good if another dev can do the same on a different staging server to double-check it :+1:

roles/common/templates/redis.conf.j2
```diff
@@ -0,0 +1,4 @@
+# {{ ansible_managed }}
+
+maxmemory 124mb
+maxmemory-policy volatile-lru
```
sauloperez, line 4: for more details: https://redis.io/topics/lru-cache. LGTM :+1:

sauloperez (approved):

sauloperez:
I'll test this now on a staging server.

roles/common/templates/redis.conf.j2
Matt-Yorkley: I kept the same default max memory setting that we currently use for Memcached (124MB).

The **Volatile LRU** eviction strategy is there for the next step in this process, where we switch session storage from `ActiveRecord` to `Redis`. At that point this specific setting will be more important :+1:

Matt-Yorkley: Basically it means if the cache store ever gets dangerously full, it will look for the most stale records first and evict them one by one as needed, so for example if you have a bunch of stale sessions that haven't been used in months, it'll drop the oldest ones first.

Matt-Yorkley:
Alright, since the PR in the main repo was merged to master we have to merge this one and provision Redis on the rest of the staging servers or deployments for testing will start failing. 

I'll do that now :+1:

andrewpbrett:
Whoops, sorry! I thought this had already been merged. :)

sauloperez:
@Matt-Yorkley something we forgot on #7648 is to add Redis to our docker-compose.yml. https://github.com/coopdevs/lazona_connector/pull/55/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3 may come in handy as a reference.

PR #740, opened 2021-05-22
Pass configs needed for mail to the app

Matt-Yorkley:
Sets up: https://github.com/openfoodfoundation/openfoodnetwork/pull/7673

Updates the way mail-related configs are passed to the app. Previously we were setting them in the environment of the unicorn user, but now we pass them to the app itself.

When testing this I also noticed there are some cases where unicorn is restarted where the ENV vars for the unicorn user are not actually passed to Rails, possibly due to the service being restarted as root (sudo)? :warning: (that's fixed now)

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -61,6 +61,24 @@ BUGSNAG_JS_KEY: {{ bugsnag_js_key }}
 
 SITE_URL: {{ domain }}
 
+MAIL_DOMAIN: {{ mail_domain }}
+MAIL_HOST: {{ mail_host }}
+MAIL_PORT: {{ mail_port }}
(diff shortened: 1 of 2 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -59,6 +59,8 @@ BUGSNAG_API_KEY: {{ bugsnag_key }}
 BUGSNAG_JS_KEY: {{ bugsnag_js_key }}
 {% endif %}
 
+SITE_URL: {{ domain }}
```
andrewpbrett, line 62: 💪

andrewpbrett (approved):

luisramos0 (approved):

sauloperez (approved):

PR #741, opened 2021-05-24
Add Ireland to all-prod 🇮🇪

andrewpbrett:
I think this is everything we need to do - @Matt-Yorkley want to double check?


sauloperez (approved):

inventory/hosts
```diff
@@ -182,6 +191,7 @@ es-prod
 fr-prod
 uk-prod
 de-prod
+ie-prod
```
Matt-Yorkley, line 194: :+1:

inventory/hosts
```diff
@@ -111,6 +111,15 @@ uk-prod
...
 uk-staging2
 
+#------------------------------------------------------------------------------
+# Ireland
+
+[ie-prod]
+openfoodnetwork.ie ansible_host=167.172.44.121
+
+[ie:children]
+ie-prod
```
Matt-Yorkley, line 121:
```suggestion
```

sauloperez:
I went ahed commit your suggestion @Matt-Yorkley so we merge it.

sauloperez:
I think it'd be about time we split this long hosts file into a file per instance, don't you agree?

PR #742, opened 2021-05-25
Remove libssl1.0-dev dependency

Matt-Yorkley:
This outdated SSL package is no longer required to build Ruby on modern systems since Ruby 2.4.x

:tada:

Also frees up the rebenv role's `rbenv_extra_depends` variable to be used for more useful things, like installing `jemalloc`...


sauloperez (changes requested):
Could you please remove it from the Dockerfile as well?

PR #745, opened 2021-05-29
Redis + Sidekiq

Matt-Yorkley:
- Adds a Sidekiq systemd service
- Configures an additional Redis instance for jobs processing
- Automates the switch from DelayedJob to Sidekiq so jobs can finish and the old DJ worker is shut down
- Maintains backwards compatibility for (non-core) instances that haven't updated yet

I created a multi-redis-instance provisioning role here, but eventually extracted it to a reusable role (in the last commit) and published it on Ansible Galaxy [here](https://galaxy.ansible.com/libre_ops/multi_redis).


Matt-Yorkley:
The Travis build is currently not happy, pending: #746

roles/deploy/handlers/main.yml
```diff
@@ -1,27 +1,63 @@
...
-  changed_when: false
-  failed_when: crontab_entries.rc > 1
-  listen: update whenever
+  become_user: root
+  when:
+   - "'delayed_job_openfoodnetwork.service' in services and services['delayed_job_openfoodnetwork.service']['status'] == 'enabled'"
+   - not sidekiq_requirement.changed
+  listen: update jobs queue
+
+- name: pause to allow final jobs to complete
```
sauloperez, line 40: so stopping the service in the task below doesn't wait for the action to finish? manually I think it does.

roles/sidekiq/templates/sidekiq.service.j2
```diff
@@ -0,0 +1,23 @@
...
+After=syslog.target network.target
+
+[Service]
+Type=notify
+WatchdogSec=10
+
+WorkingDirectory={{ current_path }}
+User={{ unicorn_user }}
+
+ExecStart=/bin/bash -lc 'exec {{ unicorn_home_path }}/.rbenv/shims/bundle exec sidekiq \
```
sauloperez, line 14: I assume sidekiq reads from `config/sidekiq.yml` by default. Is that it?

roles/deploy/handlers/main.yml
Matt-Yorkley: The first time we switch from DelayedJob to Sidekiq it should wait 30 seconds before stopping and disabling DelayedJob. I'll give it another test...

Matt-Yorkley: I missed a bit here. I just updated and re-tested and it's working correctly with the 30-seond wait :+1:

It should only do this once. After delayed job is disabled it'll ignore those steps.

bin/requirements.yml
```diff
@@ -26,5 +26,8 @@
 - src: libre_ops.wal2json
   version: 1.0.5
 
+- src: libre_ops.multi_redis
```
Matt-Yorkley, line 29: This got almost 50 downloads in the first 24 hours of going up on Ansible galaxy, so I guess people think it's useful :smile:

roles/sidekiq/templates/sidekiq.service.j2
Matt-Yorkley: Yeah, if `config/sidekiq.yml` exists, it'll read it.

roles/deploy/handlers/main.yml
Matt-Yorkley: Triple-checked :heavy_check_mark:

sauloperez (approved):

roles/deploy/handlers/main.yml
Matt-Yorkley: Screenshot from 2021-05-31 12-04-13

inventory/group_vars/all.yml
```diff
@@ -67,6 +67,20 @@ env:
...
+multiredis_instances:
+  - name: cache
+    port: 6380
+    configs: |
+      maxmemory 124mb
+      maxmemory-policy allkeys-lru
+  - name: jobs
+    port: 6381
+    configs: |
+      maxmemory 50mb
```
Matt-Yorkley, line 79: Actual memory usage of this second Redis instance (from the Sidekiq dashboard):

Screenshot from 2021-05-31 12-08-32

roles/deploy/handlers/main.yml
Matt-Yorkley: That's some silky-smooth ops right there :sunglasses:

Matt-Yorkley:
We'll need to provision `au-staging` and `es-staging` now I think :+1:

PR #746, opened 2021-05-30
Don't use jemalloc in Travis

Matt-Yorkley:
The Travis build environment is really flaky, and it apparently doesn't like `jemalloc`.

This avoids using it in Travis when testing our playbooks.


andrewpbrett (approved):

Matt-Yorkley:
A green build is enough here :+1:

PR #749, opened 2021-06-04
Upgrade to ruby 2.7.3

luisramos0:
Pairs with https://github.com/openfoodfoundation/openfoodnetwork/pull/7737

I left 2.5.8 in the list just because the upgrade is very recent, I am not sure this is necessary.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -50,7 +50,7 @@ swapfile_size: false
 
 #----------------------------------------------------------------------
 # Rails variables
-ruby_version: 2.5.9
+ruby_version: 2.7.3
 
 rbenv_extra_depends:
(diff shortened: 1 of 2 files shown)


andrewpbrett (approved):

sauloperez:
> I left 2.5.8 in the list just because the upgrade is very recent, I am not sure this is necessary.

I think we only need to keep the most recent version to allow for rollbacks but we can clean up later.

sauloperez (approved):

Matt-Yorkley:
I'll do some tests now and hopefully merge it today :ok_hand:

Matt-Yorkley:
Deployment failed. I had to add a small fix here in the deployment playbook. It's working now :ok_hand:

PR #751, opened 2021-06-05
Dotenv ofn-install

Matt-Yorkley:
Provisions configs to a **Dotenv** file per environment

To test this in Staging we'll need to provision with this PR then deploy https://github.com/openfoodfoundation/openfoodnetwork/pull/7745 and see if the ENV vars are okay.

### DIFF
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -27,8 +27,8 @@ mail_domain: example.com
 #custom_hba_entries:
 #  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 #
-#custom_application_env_vars: |
-#  OFN_FEATURE_CONNECT_AND_LEARN: "true"
+#custom_application_dotenv_vars: |
+#  OFN_FEATURE_CONNECT_AND_LEARN="true"
 
 # Overridable Image Settings
 # attachment_url:
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -27,7 +27,7 @@ mail_domain: example.com
 #custom_hba_entries:
 #  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 #
-#custom_application_dotenv_vars: |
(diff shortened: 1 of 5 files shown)


sauloperez:
Why is dotenv needed?

Matt-Yorkley:
It's just a proposal, but switching from Figaro to Dotenv in the main repo has various advantages: https://github.com/openfoodfoundation/openfoodnetwork/pull/7745#issue-662454878

I've started hitting cases where having all ENV vars for all environments in the same file at the same time with no flexibility is a real pain. I was looking at making it easier to run Sidekiq in dev, and looking at running `webpacker-dev-server` locally, and handling local Redis setups, etc. Having per-environment ENV vars that are easy to manage and apply local overrides for is really useful...

inventory/host_vars/_example.com/config.yml
```diff
@@ -27,8 +27,8 @@ mail_domain: example.com
 #custom_hba_entries:
 #  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 #
-#custom_application_env_vars: |
-#  OFN_FEATURE_CONNECT_AND_LEARN: "true"
+#custom_application_dotenv_vars: |
+#  OFN_FEATURE_CONNECT_AND_LEARN="true"
```
mkllnk, line 31: I prefer to keep the old name. It's accurately describing the content without mentioning the implementation. We may go with another gem at some point. :shrug:

mkllnk (approved):

inventory/host_vars/_example.com/config.yml
Matt-Yorkley: Hmmm... I think it's still safest to use separate variables during the transition (we can delete the old one later). How about renaming `custom_application_dotenv_vars` to just `custom_env_vars`?

mkllnk: Yes, I like that. We don't need a reference to `application.yml` any more. :+1:

mkllnk (approved):

andrewpbrett (approved):

Matt-Yorkley:
Note; there's a build error here that's not related to these changes. Should be fixed by: https://github.com/openfoodfoundation/ofn-install/pull/755

PR #752, opened 2021-06-05
Puma ofn-install

Matt-Yorkley:
Twinned with: https://github.com/openfoodfoundation/openfoodnetwork/pull/7746

Setup for switching from Unicorn to Puma.


Matt-Yorkley:
This is an interesting one in terms of Ops. We need to introduce some substantial breaking changes in order to switch to Puma, but we have non-core instances that are lagging behind on OFN versions.

Rough-sketch-on-the-back-of-a-knapkin for rollout:
- bump the main repo to `v4.0.0` when we release the Puma change
- make a copy of the ofn-install `master` branch (without these changes) and call it `ofn-v3`
- add a big note in the ofn-install `master` branch readme that it can ony be used with OFN `v4.0.0` and above, and direct users to use the `ofn-v3` branch for provisioning/deploying `v3.x.x`

Matt-Yorkley:
Note: the build currently fails as it deploys `master` from the main repo, which doesn't have the puma PR in it yet. It'll be green when https://github.com/openfoodfoundation/openfoodnetwork/pull/7746 is merged :ok_hand:

andrewpbrett (approved):
That rollout plan sounds smart. There's a small merge conflict to clean up but otherwise this sounds great.

jibees (approved):
I'm not the one who is most familiar with this kind of thing... Should we wait for a third approval, by @mkllnk?

PR #761, opened 2021-06-18
Remove unused feature toggle configuration

mkllnk:
Removing unused functionality. Partnered with https://github.com/openfoodfoundation/openfoodnetwork/pull/7806.

This will create a merge conflict with https://github.com/openfoodfoundation/ofn-install/pull/751.

### DIFF
--- a/roles/app/templates/application.yml.j2
+++ b/roles/app/templates/application.yml.j2
@@ -90,7 +90,3 @@ GEOCODER_TIMEOUT: {{ geocoder_timeout }}
 {% endif %}
 
 {{ custom_application_env_vars | default('') }}
-
-{% if beta_testers is defined %}
-BETA_TESTERS: {{ beta_testers | join(',') }}
-{% endif %}
--- a/inventory/host_vars/app.katuma.org/config.yml
+++ b/inventory/host_vars/app.katuma.org/config.yml
@@ -22,8 +22,3 @@ postgres_listen_addresses:
 
 custom_hba_entries:
   - { type: hostssl, database: "{{ db }}", user: metabase, address: '167.99.89.242/32', auth_method: md5 }
-
-beta_testers:
-  - saulopefa+ridorta@gmail.com
-  - eugeni@chafer.cat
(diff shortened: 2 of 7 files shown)


roles/app/templates/application.yml.j2
```diff
@@ -88,9 +88,3 @@ GEOCODER_SERVICE: {{ geocoder_service }}
 {% if geocoder_timeout is defined %}
 GEOCODER_TIMEOUT: {{ geocoder_timeout }}
 {% endif %}
-
-{{ custom_application_env_vars | default('') }}
```
Matt-Yorkley, line 92: This is still used in some cases.

inventory/host_vars/_example.com/config.yml
```diff
@@ -26,9 +26,6 @@ mail_domain: example.com
 #
 #custom_hba_entries:
 #  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
-#
-#custom_application_env_vars: |
-#  OFN_FEATURE_CONNECT_AND_LEARN: "true"
```
Matt-Yorkley, line 31: And we could keep an example here.

roles/app/templates/application.yml.j2
mkllnk: Is that in secrets? I couldn't find any use in ofn-install.

Matt-Yorkley: Yep :+1:

sigmundpetersen:
What to do on this one @mkllnk ?
Close or continue?

jibees (approved):
I didn't know it exists in `ofn-install` ;)
I re triggered a launch, since the CI failed. I suspect this occurs sometimes, and it's not related to this PR in particular.

mkllnk:
> I re triggered a launch, since the CI failed. I suspect this occurs sometimes, and it's not related to this PR in particular.

No, the build has been completely broken for a while. No need to re-trigger. It always fails unless we fix it.

PR #763, opened 2021-06-22
Remove --production flag from yarn install step

Matt-Yorkley:
WIP

Apparently this flag can cause issues and is not recommended.


sauloperez (approved):
no idea what that changes but if we now manage to deploy :+1:

PR #771, opened 2021-07-27
Update Semaphore deployment script

Matt-Yorkley:
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/7986

When deploying to staging via Semaphore, specific versions of master (including releases) will use the exact commit, but pull requests will use the latest version of the branch merged with master. This means we'll get accurate release versions staged, but we'll also include recent changes when testing PRs that haven't been rebased recently.

### DIFF
--- a/roles/semaphore_deployment/templates/deploy.j2
+++ b/roles/semaphore_deployment/templates/deploy.j2
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 set -e
-branch="$1"
+IFS=' ' read -ra git_args <<< "$1"
+branch=${git_args[0]}
+commit=${git_args[1]}
 allowed=^[a-zA-Z0-9._/\-]+$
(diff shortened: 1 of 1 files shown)


roles/semaphore_deployment/templates/deploy.j2
```diff
@@ -1,22 +1,36 @@
 #!/bin/bash
 
 set -e
-branch="$1"
+IFS=' ' read -ra git_args <<< "$1"
+branch=${git_args[0]}
+commit=${git_args[1]}
```
mkllnk, line 6: We don't check the commit var for invalid characters.

Matt-Yorkley: Updated

andrewpbrett (approved):

PR #777, opened 2021-09-20
www redirect

### LINKED ISSUE #643: OFN works on multiple domains, e.g. www.openfoodnetwork.org.uk and openfoodnetwork.org.uk
## Description
OFN works on multiple domains like https://www.openfoodnetwork.org.uk and https://openfoodnetwork.org.uk
and these domains have different user sessions/cookies which creates problem like https://github.com/openfoodfoundation/openfoodnetwork/issues/5179 (stripe).

We need to ensure each OFN instance only works on one domain: better use standard practice here and always use www.

## Expected Behavior
Any access to URLs like https://openfoodnetwork.tld/* redirect the user to https://www.openfoodnetwork.tld/*

## Actual Behaviour
User is able to browser OFN on https://openfoodnetwork.tld/* and https://www.openfoodnetwork.tld/* and login to each independently :face_with_head_bandage: 

## Steps to Reproduce
Login on  https://openfoodnetwork.org.uk
then go to https://www.openfoodnetwork.org.uk and see you are not logged in.

## Severity
bug-s3: a feature is broken but there is a workaround

Matt-Yorkley:
Updates configs to redirect from `www` to bare domain.

Fixes #643.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -310,7 +310,7 @@ nginx_sites:
         return 301 https://{{ domain }}$request_uri;
       }
 
-  ofn_www:
+  000_redirect_www:
     - |
       listen 443 ssl http2;
       listen [::]:443 ssl http2;
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -315,6 +315,12 @@ nginx_sites:
       listen 443 ssl http2;
       listen [::]:443 ssl http2;
       server_name www.{{ domain }};
+
+      ssl_certificate      /etc/letsencrypt/live/{{ certbot_cert_name | default(domain) }}/fullchain.pem;
+      ssl_certificate_key  /etc/letsencrypt/live/{{ certbot_cert_name | default(domain) }}/privkey.pem;
+
+      {{ nginx_tls_cipher_suites }}
+
       return 301 https://{{ domain }}$request_uri;
 
   ofn_443:
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -307,8 +307,15 @@ nginx_sites:
 
       location / {
         {{ nginx_valid_methods }}
-        return 301 https://$host$request_uri;
+        return 301 https://{{ domain }}$request_uri;
       }
+
+  ofn_www:
+    - |
+      listen 443 ssl http2;
+      listen [::]:443 ssl http2;
+      server_name www.{{ domain }};
+      return 301 https://{{ domain }}$request_uri;
 
   ofn_443:
     - |
--- a/inventory/host_vars/coopcircuits.fr/config.yml
+++ b/inventory/host_vars/coopcircuits.fr/config.yml
@@ -1,6 +1,6 @@
 ---
 
-domain: www.coopcircuits.fr
+domain: coopcircuits.fr
 host_id: fr-prod
 rails_env: production
 
@@ -12,8 +12,8 @@ unicorn_timeout: 240
 unicorn_workers: 4
 
 certbot_domains:
-  - www.coopcircuits.fr
   - coopcircuits.fr
+  - www.coopcircuits.fr
 
 postgres_listen_addresses:
   - '*'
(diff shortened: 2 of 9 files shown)


jibees (approved):
LGTM ;) Thanks!

mkllnk (approved):
:smile:

inventory/group_vars/all.yml
```diff
@@ -300,9 +300,16 @@ nginx_sites:
...
-        return 301 https://$host$request_uri;
+        return 301 https://{{ domain }}$request_uri;
       }
 
+  ofn_www:
+    - |
+      listen 443 ssl http2;
+      listen [::]:443 ssl http2;
+      server_name www.{{ domain }}
+      return 301 https://{{ domain }}$request_uri;
```
mkllnk, line 311: How can people access this if we don't have the `www.` domain in the certbot list? :thinking: 
Do we need to add that? Or is it optional? Does that mean that new instances won't have `www` by default?

mkllnk (commented):
I was just about to provision a staging server with this and test the www but then I had the thought below.

RachL:
Will this one close https://github.com/openfoodfoundation/ofn-install/issues/643 ?

mkllnk:
Yes, it will.

Matt-Yorkley:
> I was just about to provision a staging server with this and test the www but then I had the thought below.

It turns out this is a really tricky one to test because it won't actually work in staging, as staging servers never receive incoming requests for either `www.` or the bare domain. I tested this locally by setting up a production-like nginx environment with reverse-proxying and multiple virtual domains routed via custom `/etc/hosts` entries...

inventory/group_vars/all.yml
Matt-Yorkley: > Does that mean that new instances won't have www by default?

`www` will redirect to the bare domain on all instances (new or old) :+1:

> How can people access this if we don't have the `www.` domain in the certbot list?

The `domain` variable in ofn-install will always be the bare domain from now on. Any incoming requests to `www.{{ domain }}` will be processed by this server block (by nginx) and be redirected to the bare domain (which will then be processed by the regular `ofn_443` server block).

If you're doing a simple 301 response you don't need to handle TLS termination, as it'll be handled in the subsequent URL, so any additional certificates for the www subdomain are effectively obsolete/ignored after this.

There would have been a bit less complexity if it weren't for Katuma, which uses `app.katuma.org` for their primary doamin in production. In that case, requests to `app.katuma.org` will be processed the same as before. Presumably the server won't ever receive requests on `www.app.katuma.org`, but if it does they should be redirected to `app.katuma.org` correctly.

Matt-Yorkley:
Otherwise we could provision a single small instance from an opposite time zone and do some quick checks, and it would be easy to undo. 

I'm feeling lucky with this pull request number...

inventory/group_vars/all.yml
mkllnk: I'm still concerned about this one. Here's the scenario I'm thinking of:

- We deploy to au-prod.
- We have certificates for www and bare. All good.
- We don't renew the cert for www and it will expire.
- Someone enters www.openfoodnetwork.org.au in the browser. The default is http, the redirect works, all good.
- Someone links to https://www.openfoodnetwork.org.au and the browser can't establish a TLS connection to receive a redirect. :bomb: 
- Someone types the www URL in the browser and has something like a HttpsEverywhere plugin. :bomb: 
- Browser decide to make HTTPS the default. :bomb: 

You said that the redirect can be handled without TLS. But the above is in a TLS block and isn't the point of the encryption to hide and verify the request and response? So how can the server respond with a redirect if the TLS connection isn't established? Is that a fancy HTTP2 feature?

Matt-Yorkley: Aha... yeah you're right, the TLS termination has to happen before the redirect can be made. In that case, we can keep the `www.` domain in the list of `certbot` domains, but it has to be lower down the list than the bare domain. I'll re-add that now and do some tests.

Matt-Yorkley: I've updated it :+1:

Matt-Yorkley:
I did a prod-test on BE and CA. All went well. This is now live on all production servers.

PR #781, opened 2021-10-04
Add updated TLS cipher suite configuration

Matt-Yorkley:
Adds some TLS hardening and ensures ciphers blacklisted under the HTTPS2 spec don't get mistakenly used when connections are upgraded from HTTPS1 on the fly.

Should resolve some recent reports on certain Mac OSX / Safari versions resulting in `ERR_HTTP2_INADEQUATE_TRANSPORT_SECURITY` errors with HTTPS.

Also boosts our TLS analysis rating on www.ssllabs.com from a **B** to an **A+** :ok_hand:


mkllnk (approved):
:tada:

andrewpbrett (approved):

PR #789, opened 2022-01-13
Adding configuration of OFNv3 for Costa Rica (laferia.cr)

rngkll:


mkllnk (approved):

RachL:
@mkllnk do you think we can move this one forward?

mkllnk (approved):
This looks good. I'll merge. But just flagging a few issues that would prevent us from helping out:

- I can't reach https://staging.laferia.cr/.
- `ssh ofn-admin@staging.laferia.cr` is asking for a password.
- `ssh ofn-admin@laferia.cr` cannot be reached.
- If the global team was to manage this instance, we need the secrets file.

rngkll:
Hi, we have staging turned off because of cost. we turn it on when required.

Thanks for the merge.

PR #794, opened 2022-01-25
Avoid truncation of sidekiq log

mkllnk:
Each log file has to be rotated explicitly now. That makes sure that we apply rotation rules suited for each file.

But I would also be happy to go another path and log sidekiq via systemd. I guess that we would just need to remove the redirection from the service:
https://github.com/openfoodfoundation/ofn-install/blob/59888d0078b4a4a28b4983ec54355e6238d3d1e7/roles/sidekiq/templates/sidekiq.service.j2#L14-L16

**WIP:** the first commit belongs to another PR but I wanted to avoid conflicts.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -181,7 +181,8 @@ logrotate_applications:
   - name: rails
     definitions:
       - logs:
-          - "{{ current_path }}/log/*.log"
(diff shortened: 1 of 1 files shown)


Matt-Yorkley (approved):
:+1:

PR #802, opened 2022-04-11
Create adam.pub

lin-d-hop:
add adam to uk server


jibees:
Should it not be added to `inventory/group_vars/uk.yml` as well?

lin-d-hop:
yeah
I'm being lazy trying to do via the UI but I can't because two changes needed.

jibees:
@lin-d-hop https://github.com/openfoodfoundation/ofn-install/pull/803

PR #804, opened 2022-05-02
Add storage directory to shared files

mkllnk:
We want to persist the files in that directory like previously uploaded files under public/images.

I provisioned this to uk-staging and then I copied all storage files from previous deployments to the shared folder. Images seem to work again.


mkllnk:
I need this to deploy the release today. So I'll merge without review.

PR #810, opened 2022-06-10
Adding CH instance

mcmpp:


inventory/group_vars/ch.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+# See https://github.com/openfoodfoundation/ofn-install/wiki/Setup for more info
+checkout_zone: Europe
+country_code: CH
+currency: CHF
+locale: en                  # Default locale for OFN instance
+available_locales: de_CH,fr_CH,it_CH,en    # Available locales for OFN instance
```
sigmundpetersen, line 8: Should `en` here rather be `en_CH`?

mcmpp: good idea. Done!

inventory/group_vars/ch.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+# See https://github.com/openfoodfoundation/ofn-install/wiki/Setup for more info
+checkout_zone: Europe
+country_code: CH
+currency: CHF
+locale: en                  # Default locale for OFN instance
```
mkllnk, line 7:
```suggestion
locale: en_CH               # Default locale for OFN instance
```

inventory/group_vars/ch.yml
```diff
@@ -0,0 +1,23 @@
...
+timezone: Vienna
+
+
+# Used for the LetsEncrypt email setting
+developer_email: mikel.cordovilla@tegonal.com
+
+# Users listed here will have their key from /files/keys added to the app_user
+users_sysadmin:
+  - mikel
+  - matt
```
mkllnk, line 22: Why only Matt? He's actually not working on OFN at the moment. If you want us to be able to deploy:
```suggestion
  - mikel
  - "{{ core_devs }}"
```

inventory/hosts
```diff
@@ -243,3 +244,24 @@ br
...
+#------------------------------------------------------------------------------
+# Switzerland
+
+[ch-prod]
+app.openfoodswitzerland.ch ansible_host=195.15.219.220
+
+[ch:children]
+ch-prod
+
+>>>>>>> Stashed changes
```
mkllnk, line 267: This looks like a merge-conflict accident.

mkllnk (changes requested):
Great to see another instance joining! :muscle: 

This looks like you are joining the globally managed pool. Can you point me to a reference where that was discussed with the global team?

You would also need to share the secrets.yml file with me (e.g. place it on the server) so that I can add it to our encrypted repository for future provisioning.

mcmpp:
Hi @mkllnk ,

Thanks so much for your help. Answering to your points: 

Why do you assume we are joining the globally managed pool?? any configuration that should not be there?? this was not discussed as we plan to manage it by ourselves at least for the moment. I'm sorry, but we haven't been very present in the international group since we are devoting all our efforts to the local instance. When we free up some time, I hope we can contribute more to the global group.

I suppose that in case we are the ones who manage the instance, it is not necessary for me to share the secrets file with you, am I right ?

inventory/hosts
```diff
@@ -198,6 +198,7 @@ uk-prod
 de-prod
 ie-prod
 hu-prod
+ch-prod
```
mkllnk, line 201: This `europe` list is included in `all-prod` which we use to deploy all managed instances. Since we have a mix of managed and non-managed instances now, we should probably rename these groups. But for now, you need to remove this line if you don't want us to deploy to your instance.

inventory/hosts
```diff
@@ -243,3 +244,22 @@ br
...
+
+#------------------------------------------------------------------------------
+# Costa Rica
+
+[cr-prod]
+laferia.cr ansible_host=95.217.177.232
+
+[cr:children]
+cr-prod
+
```
mkllnk, line 257: Why are you adding Costa Rica here, too? We have that already:
https://github.com/openfoodfoundation/ofn-install/blob/92258df9e070c695836a2ca49f721652218bffae/inventory/hosts#L66-L74

mkllnk (changes requested):
Thank you for clarifying. A few instances joined the global pool lately and therefore it was easy for me to assume that you may be following the stream. But it's great if you want to manage your instance yourself. This PR just needs one line removing then.

inventory/hosts
mcmpp: done

mkllnk (approved):
Great, thanks!

PR #832, opened 2022-09-29
Create the n8n task

### LINKED ISSUE #822: Add N8N Postgres user to all managed instances
## Description

In order to use N8N to get metrics, giving limited, read-only access to N8N is very helpful and much lighter on instance resources than doing this via the reports API.

The user should be read only and can be limited to requests from the following IP: 128.199.44.140
I know Matt set this up for the UK Zapier user so it would be good to do the same here.

## Acceptance Criteria & Tests

1. All managed instances have an additional Postgres user called: n8n (or similar if numbers aren't allowed in usernames
2. In each case the user is limited to Read-Only access
3. Limit the user to only be able to read if accessed via 128.199.44.140
4. If it's not difficult it would also be great to add a timeout of 30secs for queries by this user.
5. After the password for the user is created can we store them in Bitwarden? I can do the legwork on that if passwords are shared securely with me :)

jibees:
close #822 
works with https://github.com/openfoodfoundation/ofn-secrets/pull/77

Still need todo:

- [ ] Limit the user to only be able to read if accessed via 128.199.44.140
- [ ] Add a request timeout to 30 sec
- [ ] Test...

 - create a n8n_user
 - Grant SELECT privs. on n8n_user for default db

### DIFF
--- a/roles/n8n/tasks/main.yaml
+++ b/roles/n8n/tasks/main.yaml
@@ -1,0 +1,22 @@
+- name: Create n8n user with
+  when: n8n_db_user is defined and n8n_db_password is defined
+  become: true
+  become_user: postgres
+  postgresql_user:
+    state: present
+    db: "{{ db }}"
+    user: "{{ n8n_db_user }}"
+    password: "{{ n8n_db_password }}"
+    role_attr_flags: NOLOGIN
+
+- name: Grant SELECT privilege on n8n user
+  when: n8n_db_user is defined and n8n_db_password is defined
+  become: true
+  become_user: postgres
+  postgresql_privs:
+    state: present
+    db: "{{ db }}"
+    privs: SELECT
+    objs: ALL_IN_SCHEMA
+    role: "{{ n8n_db_user }}"
+    grant_option: yes
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -89,6 +89,7 @@ db: openfoodnetwork
 test_db: ofn_test
 db_user: ofn_user
 db_host: localhost
+n8n_db_user: n8n_user
 postgres_encoding: en_US.utf8
 
 #----------------------------------------------------------------------
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -88,3 +88,6 @@ smtp_password:
 #geocoder_api_key: pk.xxxx
 #geocoder_service: mapbox
 #geocoder_timeout: 7
+
+# N8N
+n8n_db_password:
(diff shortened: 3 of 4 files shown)


jibees:
Executed on fr-staging:

```
ansible-playbook playbooks/provision.yml --limit=fr-staging -e "@../ofn-secrets/fr-staging/secrets.yml" --ask-vault-pass --tags n8n
```

checked on `fr-staging`:

```
openfoodnetwork=# \du
                                   List of roles
 Role name |                         Attributes                         | Member of
-----------+------------------------------------------------------------+-----------
 n8n_user  | Cannot login                                               | {}
 ofn_user  | Create DB, Replication                                     | {}
 postgres  | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
```

Unfortunately seems like priviliege doesn't work: 

```
openfoodnetwork=# SELECT * FROM information_schema.table_privileges where grantee = 'n8n_user';
 grantor | grantee | table_catalog | table_schema | table_name | privilege_type | is_grantable | with_hierarchy
---------+---------+---------------+--------------+------------+----------------+--------------+----------------
(0 rows)
```

roles/n8n/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
+- name: Create n8n user with
+  when: n8n_db_user is defined and n8n_db_password is defined
+  become: true
+  become_user: postgres
+  postgresql_user:
+    state: present
+    db: "{{ db }}"
+    user: "{{ n8n_db_user }}"
+    password: "{{ n8n_db_password }}"
+    role_attr_flags: NOLOGIN
```
mkllnk, line 10: Why? Doesn't the user need to log in? Or is that only temporary until you did the IP restriction?

mkllnk (commented):
Previous integrations didn't use a separate role. We used this for IP restriction:
https://github.com/openfoodfoundation/ofn-install/blob/4ecc7ac92dde7244b8325bfd31583b5dcc25042f/inventory/host_vars/openfoodnetwork.ca/config.yml#L19-L20

And in secrets:
```
db_integrations:
  - { user: zapier, state: present, password: "{{ zapier_db_password }}" }
```

That doesn't give you the same flexibility though. Would it work here?

I have code for the 30 second restriction. I hope to share that later.

mkllnk:
I opened my approach here: https://github.com/openfoodfoundation/ofn-install/pull/834

But I didn't have enough time to test it thoroughly.

roles/n8n/tasks/main.yaml
jibees: Yep, I wanted to be sure no one can login through this one. But it needs to be deleted. 
I guess my whole PR should be deleted as well since yours is better.

PR #834, opened 2022-10-03
N8n support

### LINKED ISSUE #822: Add N8N Postgres user to all managed instances

mkllnk:
This is an alternative to https://github.com/openfoodfoundation/ofn-install/pull/832.

closes #822 

I think that this works but I couldn't test it yet. I accidentally upgraded postgresql on au-staging and the new version doesn't seem to be supported by our role. We need to fix the db_integrations playbook.

### DIFF
--- a/roles/db_integrations/tasks/main.yml
+++ b/roles/db_integrations/tasks/main.yml
@@ -26,6 +26,7 @@
     - "GRANT SELECT ON ALL SEQUENCES IN SCHEMA public TO {{ db_integration.user }};"
     - "ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO {{ db_integration.user }};"
     - "ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON SEQUENCES TO {{ db_integration.user }};"
+    - "ALTER ROLE {{ db_integration.user }} SET statement_timeout = $$30s$$;"
   loop_control:
     loop_var: permission_query
   when: db_integration.state == 'present' and integration_config.changed
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -116,6 +116,8 @@ default_hba_entries:
   - {type: host, database: all, user: all, address: '127.0.0.1/32', auth_method: md5}
   - {type: host, database: all, user: all, address: '::1/128', auth_method: md5}
 
+custom_hba_n8n: {type: hostssl, database: "{{ db }}", user: "n8n", address: "128.199.44.140/32", auth_method: md5}
+
 custom_hba_entries: []
 
 postgresql_hba_entries: "{{ default_hba_entries + custom_hba_entries }}"
--- a/inventory/host_vars/_example.com/config.yml
+++ b/inventory/host_vars/_example.com/config.yml
@@ -21,12 +21,13 @@ mail_domain: example.com
 #swapfile_size: 1G
 
 # Enable external database access for third party integrations
-#postgres_listen_addresses:
-#  - '*'
-#
-#custom_hba_entries:
+postgres_listen_addresses:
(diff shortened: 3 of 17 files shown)


roles/db_integrations/tasks/main.yml
```diff
@@ -26,6 +26,7 @@
     - "GRANT SELECT ON ALL SEQUENCES IN SCHEMA public TO {{ db_integration.user }};"
     - "ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO {{ db_integration.user }};"
     - "ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON SEQUENCES TO {{ db_integration.user }};"
+    - "ALTER ROLE {{ db_integration.user }} SET statement_timeout = $$30s$$;"
```
jibees, line 29: > This new option is only executed when a user is added and therefore
existing users are not affected. New users like n8n and metabase will be
restricted.

I'm wondering what happens if we provision a new instance? All user will have this 30s timeout? You don't think it can be a problem?

jibees (commented):
Thanks! 
This PR doesn't create the `n8n` user in db. You don't think it's necessary?

roles/db_integrations/tasks/main.yml
mkllnk: This only applies to the db_integrations. That's zapier, metabase and n8n. I think that all of them should be restricted to avoid downtime of the app.

mkllnk:
> This PR doesn't create the n8n user in db. You don't think it's necessary?

That happens in the ofn-secrets PR.

jibees:
So, we tried to connect with N8N, still without any success. 
We suspect any issue with the IP configured as address: 

```
{type: hostssl, database: "{{ db }}", user: "n8n", address: "128.199.44.140/32", auth_method: md5}
```

That's why, in order to test, we changed it to :
```
{type: hostssl, database: "{{ db }}", user: "n8n", address: "0.0.0.0/0", auth_method: md5}
```
to accept connections from any IP addresses. 

I tested through `psql` command, locally:

```
psql -h staging.openfoodnetwork.org.uk -p 5432 -U n8n openfoodnetwork -W
```

I've seen it worked, but now, it seems to not work anymore, don't know why!

```
psql: error: connection to server at "staging.openfoodnetwork.org.uk" (188.166.136.240), port 5432 failed: Operation timed out
	Is the server running on that host and accepting TCP/IP connections?
```


Of course, integration with N8N is not working either.

lin-d-hop:
Ok so we need to add to the ip-whitelist for UK staging by updating the OFN security ansible scripts:
https://github.com/openfoodfoundation/ofn-security/pull/36/commits/fcdfb22b9832fd159b064b95a32ba6be924e5089

jibees:
Then, I'v created this PR https://github.com/openfoodfoundation/ofn-security/pull/40 to make the n8n ip white ;)

jibees:
_**Conclusion**_

This PR can be merged!

mkllnk:
I merged the security PR and tried to provision but uk-staging is either down or blocking my IP....

PR #835, opened 2022-10-10
Update postgres role and config for db integration

mkllnk:
The db_integrations playbook was broken. I guess that the internal workings of the postgresql role changed and it didn't expect a subtask to be included on its own. I'm now including the whole role as in the provisioning playbook to make it work again and reduce maintenance. It will be a little bit slower though.

### DIFF
--- a/bin/requirements.yml
+++ b/bin/requirements.yml
@@ -21,7 +21,7 @@
   version: 4.16.0
 
 - src: geerlingguy.postgresql
-  version: 3.3.0
+  version: 3.4.0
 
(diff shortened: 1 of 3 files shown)


jibees (approved):

jibees:
Once this is merged, don't know how the process continue: should we provision all the servers?

mkllnk:
> should we provision all the servers?

No, this is just a fix to make sure that we can run the playbooks when needed. But it doesn't need to run now. It will unblock the n8n user provisioning.

PR #837, opened 2022-10-13
Unblock the installation of yarn on the CI (#1)

heroinedor:
unblock the installation of yarn on the CI: an issue exist about this on the NPM module : ansible/ansible#29240 (comment) The workaround is to use enviroment to define the PATH in ansible's yaml

### DIFF
--- a/roles/yarn/tasks/main.yml
+++ b/roles/yarn/tasks/main.yml
@@ -15,3 +15,4 @@
   environment:
     NPM_CONFIG_PREFIX: "{{ npm_prefix }}"
     NODE_PATH: "{{ npm_prefix }}/lib/node_modules"
+    PATH: "/usr/bin:{{ ansible_env.PATH }}"


jibees (approved):
Thanks! 🙏 for the fix, and the reference

Specs failed, but I guess this one is not enough to make specs green.

jibees:
Maybe you could create only one PR with atomic commit? I guess this will be clear enough.

heroinedor:
> Maybe you could create only one PR with atomic commit? I guess this will be clear enough.

I would rather like to make one PR per correction because the errors are very long to find, reproduce and analyse.
And I fear that the whole  work (make the CI green again) would take too much time for a single PR.

mkllnk (approved):
Great. I'm happy with smaller PRs, each addressing one problem.

mkllnk:
### Testing notes

I provisioned with this change and it doesn't change already provisioned servers because yarn is installed already. I then uninstalled yarn and provisioned again. Everything works as before. A deployment afterwards went well. All good. Merging.

PR #840, opened 2022-11-02
Increase Puma workers and threads for FR config

pacodelaluna:
This is related to the peformance issue we faced in France.

https://github.com/openfoodfoundation/openfoodnetwork/issues/9859

Increasing the workers and threads manually seems to give us better performance. Note that we have 8 workers on Nginx side. The idea was that Nginx requests were flooding Puma requests pool.

Oct 23, 5:10 pm – Oct 23, 11:19 pm
Screenshot from 2022-11-02 12-02-43

Oct 30, 5:10 pm – Oct 30, 11:19 pm
Screenshot from 2022-11-02 12-04-14

Even if it is not easy to compare as there is a bit less traffic on the second date, it seems to have a positive impact to have increase these workers and threads.


jibees (approved):
Looks good to me, as it reflect the configuration we currently have!
Thanks! 🙏

PR #846, opened 2022-11-30
Add custom HBA entry to connect Metabase to au-prod database

dacook:
I'm using the zapier user, because it was quicker than creating a new one ;)


jibees (approved):
LGTM.
Merging!

PR #850, opened 2022-12-15
Replace malloc with jemalloc

abdellani:
Hi 

The objective of this PR is to make the rails server use `jemalloc` instead of `malloc`. This modification may solve the memory leak issue as `jemalloc` frees the allocated memory space automatically when it's no longer in use.

In this PR, I installed the library on the sytem, then updated the command that runs the puma server on `roles/webserver/templates/puma.service.j2`

If this solution works, we can discuss compiling the ruby with jeamlloc as explained [here](https://scalingo.com/blog/improve-ruby-application-memory-jemalloc)

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -44,6 +44,7 @@
       - libxslt1-dev
       - libxrender1
       - webp
+      - libjemalloc2
   tags: packages
 
 - name: install base packages for old distros
--- a/roles/webserver/templates/puma.service.j2
+++ b/roles/webserver/templates/puma.service.j2
@@ -13,7 +13,7 @@ Environment=BUNDLE_GEMFILE={{ current_path }}/Gemfile
 # Helpful for debugging socket activation, etc.
 # Environment=PUMA_DEBUG=1
 
-ExecStart=/bin/bash -lc "bundle exec --keep-file-descriptors puma -C {{ shared_path }}/config/puma.rb"
+ExecStart=/bin/bash -lc "LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.2 bundle exec --keep-file-descriptors puma -C {{ shared_path }}/config/puma.rb"
 ExecReload=/bin/bash -lc "bundle exec pumactl phased-restart -S {{ shared_path }}/pids/puma.state"
 
 SyslogIdentifier=puma


jibees:
I can see:

```
TASK [common : install base packages] ******************************************
fatal: [local_test]: FAILED! => {"changed": false, "msg": "No package matching 'libjemalloc2' is available"}
```

in the spec.

abdellani:
@jibees  Thank you.
The package repository used on the staging server includes the `libjemalloc1` only.

```bash
~$ apt search jemalloc
Sorting... Done
Full Text Search... Done
libjemalloc-dev/bionic,now 3.6.0-11 amd64 [installed]
  development files and documentation for jemalloc

libjemalloc1/bionic,now 3.6.0-11 amd64 [installed,automatic]
  general-purpose scalable concurrent malloc(3) implementation
```

dacook:
I just found this commit from a while ago: https://github.com/openfoodfoundation/ofn-install/commit/cffc49a398d93ac4beb8ee4e2657b908b18e37b9 . This is to compile Ruby with jemalloc. 
Does that mean it's already in use by Puma, or do we still need to add the puma parameter to use it? 

If so, then we can still test this to see if jemalloc helps with our memory issues. 
This is the testing I did before, which you could  try replicating: https://github.com/openfoodfoundation/openfoodnetwork/issues/10016#issuecomment-1354052177 (running the Order Cycle Sales report for a full year of data).

I think we can prove the concept by manually editing the script without needing to take time updating the playbook. So:
1. Maybe we can test on a staging environment using the puma parameter for jemalloc1
2. Also we could test on vagrant (with some prod data imported) using the puma parameter for jemalloc2

I'm going to have a look at the websockets issue next, but if I have time today I will try one of these tests to see if it makes a difference. Otherwise hopefully you can try Mohamad.
If it doesn't show any memory improvement, then I guess we have to go back to the drawing board...

## More regarding jemalloc package version
So `libjemalloc2` installs successfully on my Vagrant VM (Ubuntu 20) using this branch:
```
gh pr checkout 850
vagrant up
ansible-playbook provision.yml --limit=vagrant
...
TASK [common : install base packages] **********************************************************************************
changed: [local_vagrant]
...
```

But the GitHub Actions are configured to use Ubuntu 18, so I guess that version isn't available.
I can see that the FR, UK and AU staging servers are also using older Ubuntu.
FR Production, however is using Ubuntu 20 with jemalloc2 installed.

BTW I noticed on the ubuntu servers:
* jemalloc1 is actually v3.6.0
* jemalloc2 is actually v5.2.1 (2019)

The latest version was released earlier this year: v5.3.0
I don't think it will matter too much which version we test.

PR #852, opened 2022-12-19
Add websockets connection to `/cable`

### LINKED ISSUE #851: Allow `/cable` path as a websocket connection
## Description
Since https://github.com/openfoodfoundation/openfoodnetwork/pull/9804 and https://github.com/openfoodfoundation/openfoodnetwork/pull/9630, we now use reflexes and then websocket connection. 
We need to modify `nginx` configuration of each server to accept such connection.
This should be done via:

```
location /cable {
 proxy_pass http://rails;
 proxy_http_version 1.1;
 proxy_set_header X-Forwarded-Proto https;
 proxy_set_header X-Forwarded-Ssl on;
 proxy_set_header Upgrade $http_upgrade;
 proxy_set_header Connection "upgrade";
}
```

in the `/etc/nginx/sites-available/ofn_443.conf` file.

jibees:
Closes #851

#### How to test

Probably the best way is to provision a staging server (not the FR one), which is well configured + deploy https://github.com/openfoodfoundation/openfoodnetwork/pull/9804 on that server and check that on `/admin/orders` page you don't have any websocket errors

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -388,6 +388,16 @@ nginx_sites:
         try_files /maintenance.html =503;
       }
 
+      location /cable {
+        {{ nginx_valid_methods }}
+        proxy_pass http://rails;
+        proxy_http_version 1.1;
+        proxy_set_header X-Forwarded-Proto https;
+        proxy_set_header X-Forwarded-Ssl on;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+      }
+
       client_max_body_size 4G;
       keepalive_timeout {{ puma_timeout }};
       proxy_read_timeout {{ puma_timeout }};


inventory/group_vars/all.yml
```diff
@@ -388,6 +388,18 @@ nginx_sites:
         try_files /maintenance.html =503;
       }
 
+      location /cable {
+        {{ nginx_valid_methods }}
```
dacook, line 392: It seems like a good idea that we limit for only the valid methods here. 
There's no comment for `nginx_valid_methods` so I assume it's just general good practice to only open up what's necessary. In that case, I wonder if it could be limited further, because websockets probably don't require all those verbs.

But.. I don't think it matters so would approve this.

inventory/group_vars/all.yml
```diff
@@ -388,6 +388,18 @@ nginx_sites:
...
+        proxy_pass http://rails;
+        proxy_http_version 1.1;
+        proxy_set_header X-Forwarded-Proto https;
+        proxy_set_header X-Forwarded-Ssl on;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+      }
+
+
+
```
dacook, line 402: Sorry to be picky, but I think we should avoid avoid multiple blank lines for consistent style.

dacook (changes requested):
Thanks for doing this.
It would be great to also update the code for style, then we are good for dev test!

inventory/group_vars/all.yml
jibees: I've used the same that others locations. I agree, it's probably not ideal. Maybe leave as it for now?

inventory/group_vars/all.yml
jibees: I agree.

jibees:
> It would be great to also update the code for style, then we are good for dev test!

Changed, so moving to `Dev Test Ready`

mkllnk:
I deployed https://github.com/openfoodfoundation/openfoodnetwork/pull/9804 and could reproduce the errors. Then I provisioned with this PR and everything seemed to work. :smile:

mkllnk:
Now I also provisioned all servers. This is shipped.

PR #855, opened 2023-01-04
Remove nakayoshi_fork gem from puma config

jibees:
This gem is 6 years old and makes puma 6 not start: remove it
Seems unnecessary when using ruby 2.7
https://github.com/ko1/nakayoshi_fork/issues/5

Unblocks:

* https://github.com/openfoodfoundation/openfoodnetwork/pull/10217

### DIFF
--- a/roles/webserver/templates/puma.rb.j2
+++ b/roles/webserver/templates/puma.rb.j2
@@ -20,7 +20,6 @@ if ["production", "staging"].include?(ENV["RAILS_ENV"])
   prune_bundler
   wait_for_less_busy_worker
   fork_worker
-  nakayoshi_fork
   preload_app! false
 
   # Ensure clean Redis connections when forking


mkllnk (approved):
Great! I'll test this one.

It would have been nice to reference the blocked upgrade issue to easier click through.

dacook:
Sounds good. From what I can see the nakayoshi_fork gem was created to deal with a memory issue created in Ruby 2.2.
Puma later included this in core, but then removed it in [v6](https://github.com/puma/puma/blob/master/6.0-Upgrade.md).

dacook (approved):

mkllnk:
I tested this on staging and the app works again after deploying https://github.com/openfoodfoundation/openfoodnetwork/pull/10217. :heavy_check_mark: 

I'm now provisioning all production servers.

PR #856, opened 2023-01-12
Adds link to the wiki on deployment with Ansible

filipefurtad0:
Adds link to the wiki on deployment with Ansiable

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,8 @@ Deploying Open Food Network
 
 These are [Ansible](http://docs.ansible.com/ansible/) playbooks (scripts) for managing an Open Food Network app server. This is **not for your local development environment**. Head to the [OFN getting started guide](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/GETTING_STARTED.md) to run the OFN locally.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,9 @@ Some playbooks require third-party roles, which are specified in `bin/requiremen
 $ bin/setup
 ```
 
+## Deployment with Ansiable
```
sigmundpetersen, line 68:
```suggestion
## Deployment with Ansible
```

dacook:
Thanks for working to make sure this is up to date @filipefurtad0 . What do you think of my suggested change in the new commit?

README.md
filipefurtad0: I did not spot that one :sweat_smile:  I've taken the name from [here](https://github.com/openfoodfoundation/ofn-install/wiki/Deployment-with-Ansiable) -> I've fixed it it now on the wiki, as well.

filipefurtad0:
> Thanks for working to make sure this is up to date @filipefurtad0 . What do you think of my suggested change in the new commit?

Sure @dacook - I agree, let's rather go with that option :+1:

README.md
sigmundpetersen: I think it should be 'Ansible' (https://www.ansible.com/)

filipefurtad0: > I think it should be 'Ansible' (https://www.ansible.com/)

Of course... Sorry, for my mix up.

PR #857, opened 2023-01-16
Reduce memory-hungry Puma workers, add threads

mkllnk:
Puma is multi-threaded and one process can handle many connections without duplicating memory-use. We may need to increase the database connection pool size.

* Helps with #10016.


mkllnk:
Looking further at recommended process numbers, we could go as high as 8 to utilise all 8 CPU cores. We just don't have the memory for that at the moment. If we can solve the reports problem with a background process then the current 4 workers should be no problem. So I'm closing this again.

PR #858, opened 2023-01-17
Update readme

dacook:
A couple of improvements.


dacook:
I don't know where the ansible variable `fetch_ofn_secrets` comes from, so didn't document that. Didn't remove it incase it breaks somebody's scripts.

https://github.com/dacook/ofn-install/blob/update-readme/playbooks/provision.yml#L13

dacook:
(cancelled CI as it's only a readme change)

mkllnk (approved):
Thank you. I think that it's okay to mention private repos when it's useful for us. Openness has many benefits.

PR #860, opened 2023-01-20
Create stats extension on all servers

mkllnk:
It's only used for Datadog but the Rails application is tracking extensions in its schema.rb file. Rails tries to create the extension itself if it doesn't exist and the db user is not allowed to create exensions at the moment. This seemed to be the easiest way but we could also give the db user the CREATE privilege instead.

### Dev test

Provision fr-staging and then try to deploy OFN master. It should succeed now. The other two staging servers have the extension already.

- [x] au-staging
- [x] fr-staging
- [x] uk-staging

Follow up from:

- https://github.com/openfoodfoundation/openfoodnetwork/pull/10278

### DIFF
--- a/roles/dbserver/tasks/main.yml
+++ b/roles/dbserver/tasks/main.yml
@@ -35,6 +35,7 @@
 - import_tasks: fix_template_encoding.yml
   tags: template_encoding
 
+# As required in db/schema.rb (for Datadog)
 - name: create pg_stat_statements extension
   become: yes
   become_user: postgres
--- a/roles/dbserver/tasks/main.yml
+++ b/roles/dbserver/tasks/main.yml
@@ -34,3 +34,10 @@
 
 - import_tasks: fix_template_encoding.yml
   tags: template_encoding
+
+- name: create pg_stat_statements extension
+  become: yes
+  become_user: postgres
+  postgresql_ext:
+    name: pg_stat_statements
+    db: "{{ db }}"
--- a/roles/datadog/tasks/pg_stats.yml
+++ b/roles/datadog/tasks/pg_stats.yml
@@ -1,11 +1,4 @@
 ---
-- name: create extensions
-  become: yes
-  become_user: postgres
-  postgresql_ext:
-    name: pg_stat_statements
-    db: "{{ db }}"
-
 - name: add postgres stats configuration
   template:
     src: stats.conf.j2


roles/dbserver/tasks/main.yml
```diff
@@ -34,3 +34,10 @@
 
 - import_tasks: fix_template_encoding.yml
   tags: template_encoding
+
```
dacook, line 37: Might as well mention why:

```suggestion

# As required in db/schema.rb (for Datadog)
```

dacook (approved):
Looks good, but I suggest adding a comment.

dacook:
## Prod deploy results

Katuma failed to update certificate. I'm guessing that domain is not important.
```
IMPORTANT NOTES:
 - The following errors were reported by the server:
   Domain: alpha.katuma.org
   Type:   unauthorized
   Detail: 185.199.110.153: Invalid response from
   http://alpha.katuma.org/.well-known/acme-challenge/A9Q8HhYAsv1lUY4GxDo3G_zEsjQtBrWQpdYToyO80cc:
   404

   To fix these errors, please make sure that your domain name was
   entered correctly and the DNS A/AAAA record(s) for that domain
   contain(s) the right IP address.
```

IE, BE and DE failed to update apt cache. It's probably fine I guess.
```
TASK [oefenweb.swapfile : install dependencies] *************************************************************************************************************
failed: [openfoodnetwork.ie] (item=['util-linux']) => {"ansible_loop_var": "item", "changed": false, "item": ["util-linux"], "msg": "Failed to update apt cache: W:The repository 'http://apt.postgresql.org/pub/repos/apt xenial-pgdg Release' does not have a Release file., W:Data from such a repository can't be authenticated and is therefore potentially dangerous to use., W:See apt-secure(8) manpage for repository creation and user configuration details., E:Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/xenial-pgdg/main/binary-amd64/Packages  404  Not Found [IP: 217.196.149.55 80], E:Some index files failed to download. They have been ignored, or old ones used instead."}
```

FR became unreachable. seems to happen from my machine sometimes. Retrying...
```
TASK [ssh_keys : add ssh keys users_sysadmin] ***************************************************************************************************************
ok: [coopcircuits.fr] => (item=paco)
fatal: [coopcircuits.fr]: UNREACHABLE! =>
```

dacook:
All good.
```
coopcircuits.fr            : ok=217  changed=14   unreachable=0    failed=0    skipped=166  rescued=0    ignored=0
```

dacook:
Ok, I should have noticed that the provisioning aborted on IE, BE and DE, so this change didn't get provisioned. Running again...
BE succeeds,
```
openfoodnetwork.be         : ok=147  changed=12   unreachable=0    failed=0    skipped=130  rescued=0    ignored=0
```
 but IE and DE same error.
```
Failed to update apt cache: W:The repository 'http://apt.postgresql.org/pub/repos/apt xenial-pgdg Release' does not have a Release file., W:Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
```

But why did BE succeed, after failing last week? Hmm this doesn't look like an intermittent issue, but I tried again, and they worked this time.

```
openfoodnetwork.de         : ok=141  changed=11   unreachable=0    failed=0    skipped=119  rescued=0    ignored=0
openfoodnetwork.ie         : ok=138  changed=11   unreachable=0    failed=0    skipped=122  rescued=0    ignored=0
```

Weird! Definitely all done now.

PR #862, opened 2023-01-24
Allow Metabase to connect to managed databases

mkllnk:
Preparation for:

* https://github.com/openfoodfoundation/openfoodnetwork/issues/9161

Our Metabase server has been working with replicated databases. But our replication setup breaks regularly and we lost the knowledge how it works.

To simplify the setup, we are going to connect Metabase directly to the live database (with a statement timeout).

This pull request allows connecting to postgresql. We will still need to add the user with a secret password managed in ofn-secrets.


jibees (approved):
👌

dacook (approved):
Sorry I missed this earlier. All good 👍

PR #864, opened 2023-01-31
Remove transition code from DelayedJob

mkllnk:
And a bit more cleaning.

We disabled DelayedJob a long time ago. We can consider all servers updated or an older version needs to be used.

### DIFF
--- a/roles/deploy/handlers/main.yml
+++ b/roles/deploy/handlers/main.yml
@@ -1,64 +1,11 @@
 ---
 
-- name: check services
-  service_facts:
-  listen: update jobs queue
-
-- name: check sidekiq presence in gemfile
-  lineinfile:
-    dest: "{{ build_path }}/Gemfile"
-    regex: "^gem .*sidekiq"
-    state: absent
-  check_mode: yes # Don't actually modify the file
-  become: yes
-  register: sidekiq_requirement
-  ignore_errors: true
-  listen: update jobs queue
-
 - name: restart sidekiq
   service:
     name: sidekiq
     state: restarted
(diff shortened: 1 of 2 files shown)


jibees (approved):
Great news this kind of cleaning 🧹 

Thanks for that @mkllnk !

jibees:
Is there any way to test this one? I guess not (and we should merge directly) but i'm wondering. 

Maybe it needs a second review as well.

mkllnk:
We could provision a staging server and see that it still works. I think I did that already. I'm requesting a second review but then I think that we can merge it.

roles/deploy/handlers/main.yml
```diff
@@ -1,64 +1,11 @@
...
-  listen: update jobs queue
-
-- name: pause to allow final jobs to complete
-  wait_for:
-    delay: 30
-    timeout: 0
-  when:
-    - "'delayed_job_openfoodnetwork.service' in services and services['delayed_job_openfoodnetwork.service']['state'] == 'running'"
-    - "'sidekiq.service' in services"
-    - sidekiq_requirement.changed
```
dacook, line 48: TIL about having a list of `when` conditions.

dacook (approved):
👍

dacook:
**Test deployment**: ✅

```
...
RUNNING HANDLER [deploy : restart sidekiq] *****************************************************************************
changed: [staging.openfoodnetwork.org.au]
Tuesday 07 February 2023  09:56:46 +1100 (0:00:13.021)       0:04:35.971 ******
```

PR #868, opened 2023-03-16
Remove auto fetch of ofn secrets

### LINKED ISSUE #861: Provisioning uses old secrets when using fetch_secrets
## Description
I've discovered and activated the `fetch_secrets` playbook to run before provisioning. But when I provision, it's still applying an old secret, specifically the smtp login settings in `env.j2`.

If I specify the secrets directly with the `-e` parameter, it uses the correct values.

## Expected Behavior
This playbook downloads secrets from a separate repository and saves them locally in the `inventory/host_vars` folder. I can see that it's loading the latest file there.

So when provisioning, I expect the latest secret value to be applied in a template, specifically the smtp login settings in `env.j2`.

## Actual Behaviour
The old value got used. I don't know where it came from..

Am I using it correctly?? Or should we remove this option to avoid confusion, in favour of the manual method instead.

## Steps to Reproduce
1. Update the smtp login values in ofn-secrets (requires Ansible vault, see ofn-secrents/readme)
2. Enable FETCH_OFN_SECRETS=TRUE
3. Run provision.yml playbook
4. Check for a changed file: `changed: [openfoodnetwork.ca] => (item={'src': 'env.j2', 'dest': '<hidden>/.env.production'})`
5. Check the contents of the above file

## Severity

bug-s5: we can live with it, only a few users impacted

## Your Environment
* Version used: latest
* Operating System and version (desktop or mobile): macOS 10.15
* ansible 2.9.20
* python version = 3.8.2

dacook:
* Fixes #861

This was faulty: it would copy the latest files, but not load the values into memory, until next time you run ansible. Therefore you're always one step behind, and could copy old config onto a server (this has happened).

I tried to fix it with ansible's `refresh_inventory` command, but frustratingly can't get it to work:
* closes https://github.com/openfoodfoundation/ofn-install/pull/867

So I suggest you have to manually execute `fetch_secrets.yml` before provisioning.

## Documentation
After merging I'll also update the wiki:

1. https://github.com/openfoodfoundation/ofn-install/wiki/Provisioning#using-ofn-secrets
2. https://github.com/openfoodfoundation/ofn-install/wiki/Tips-for-core-sysadmins#setup-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,18 +66,19 @@ $ bin/setup
 
 ## Secrets
 
-Some tasks require host-specific secrets. These can be provided with a parameter like so:
+Some tasks require host-specific secrets, and will show an error if they haven't been provided. These can change from time to time, so **always ensure you have the latest before provisioning**.
+
+Secrets can be provided with a parameter like so:
 
 ```sh
-ansible-playbook playbooks/provision.yml --limit=au-prod -e "@../path/to/au-prod/secrets.yml"
+ansible-playbook playbooks/provision.yml --limit=au-staging -e "@../ofn-secrets/au-staging/secrets.yml"
 ```
 
-If you have access to the `ofn-secrets` repository, you can automatically fetch them with the `fetch_secrets` role, activated by an environment variable:
+If you have access to the `ofn-secrets` repository, you can fetch them with the `fetch_secrets.yml` playbook. The secrets for each host will be loaded into the relevant directory in `inventory/host_vars/`.
 
 ```sh
-    echo "export FETCH_OFN_SECRETS=TRUE" >> ~/.bash_profile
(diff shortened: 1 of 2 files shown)


mkllnk (approved):
You found the bug! Congratulations. :star:

jibees (approved):
👌

jibees:
Should we merge ; dev test?

dacook:
I'm not sure what you think @mkllnk , but I think the only test needed is that the provision playbook still works.

The important thing will be to make sure we _always_ update new secrets before running it. 

I think ideally we would use a wrapper shell script to run the fetch playbook first (i didn't do this to avoid potentially spending more time than it was worth.. but it should be pretty easy..)

dacook:
I've updated the above documentation.

PR #869, opened 2023-04-07
Deployment post tasks: Force `split_checkout` feature to be toggle on

jibees:
Main question is: 

This will run after each deployment: should it be the case? Shouldn't we check for any version? 

Closes: https://github.com/openfoodfoundation/openfoodnetwork/issues/10658

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -72,3 +72,8 @@
         - debug:
             msg: Deployment failed
           failed_when: True # Ensure Ansible returns a non-zero error code after rescue
+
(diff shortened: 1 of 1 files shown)


RachL:
> we only want that on managed instances.

We can do that for everybody if it's easier. There is no reason to do it only for instances we managed.

dacook:
Thanks for confirming, I suggest we do a rails migration which happens just once when it is released (which we will communicate about the change). If we were to do it after every deploy, it creates more moving parts to manage, and we have to remove it again one day.

PR #873, opened 2023-05-01
Allow more time to clone big order cycles

### LINKED ISSUE #871: Increase timeout on USA server
## Description
We are seeing that cloning a complex order cycle is not possible because it seems there is a server timeout before it finished, resulting in a snail error and an order cycle that doesn't include all products. 

## Expected Behavior
The order cycle and other long complex tasks should be able to complete successfully without timing out.

## Actual Behaviour
The slug appears and the network tab of dev tools tools like this:
Screenshot 2023-04-13 9 06 11 AM

## Steps to Reproduce

1. Select "reference" order cycle for South Cumberland Farmers Market enterprise in openfoodnetwork.net
2. Click "clone" button
3. Wait for slug

## Animated Gif/Screenshot
[Screen recording 2023-04-13 9.01.18 AM.webm](https://user-images.githubusercontent.com/37803238/232074081-67db7fa0-f8c8-4bed-a4ef-2746ff4aefe3.webm)

## Your Environment
Multiple browsers, desktops, and mobile devices

## Possible Fix
Update server timeout (possibly like with https://github.com/openfoodfoundation/ofn-install/issues/841 but not sure how many timeouts there are and which would need to be adjusted)

mkllnk:
* Closes #871 

I ran this already.

### DIFF
--- a/inventory/host_vars/openfoodnetwork.net/config.yml
+++ b/inventory/host_vars/openfoodnetwork.net/config.yml
@@ -25,3 +25,6 @@ custom_hba_entries:
   - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 
 enable_rails_apm: "false"
+
+# Allow more time to clone big order cycles.
+puma_timeout: "60"


lauriewayne:
@mkllnk is it possible to bump it up a little more? The user reports that they are getting the snail 50% of the time. Would it be a problem to up the puma timeout to 120?
Screen Shot 2023-05-10 at 12 42 48 PM

mkllnk:
Looks like Matt has a better solution.

mkllnk:
Timeout increased to 120 seconds for now.

lauriewayne:
Wonderful, thank you @mkllnk! And @Matt-Yorkley thank you for the solution - it looks like it results in really significant performance improvements!! 🙏 🏅

drummer83:
https://github.com/openfoodfoundation/openfoodnetwork/pull/10832 has been merged now with huge performance improvement! :muscle: 
Shall we think about reducing some timeouts after deployment?

mkllnk:
> Shall we think about reducing some timeouts after deployment?

Let's do that after background reports is fully in production.

PR #874, opened 2023-05-02
Apply Postgres tuning to all servers

Matt-Yorkley:
The default configuration for Postgres isn't really adequate for production environments when you install it from the standard package and don't configure it. We previously applied this configuration to larger instances on a case by case basis but it's been removed on some servers during server upgrades. I've scaled it down a little bit to be more conservative and added it to the provision playbook.

The technical part here is that if Postges is performing a lot of joins or complex queries and it can't hold what it needs in buffers, it actually stops and writes things to *disk* half way through a query and has to read them from disk again before it can return the results. Potentially even multiple times per query! This can massively increase the time it takes to complete a request.

I did some testing by running some reports as superadmin with a large date range using production data to simulate worst case scenarios and got a response time up to 8 seconds. After applying these configs locally the same request dropped down to around 1 second. Basically any querying where the complexity is on the high end of the scale (like with permissions) it should make a real difference.

### DIFF
--- a/roles/postgres_tuning/tasks/main.yml
+++ b/roles/postgres_tuning/tasks/main.yml
@@ -1,0 +1,20 @@
+---
+- name: Add postgres tuning configuration
+  copy:
+    content: |
+      shared_buffers = {{ ( ansible_memtotal_mb / 5 ) | int }}MB
+      effective_cache_size = {{ ( ansible_memtotal_mb / 2 ) | int }}MB
+      maintenance_work_mem = {{ ( ansible_memtotal_mb / 25 ) | int }}MB
+      wal_buffers = 16MB
+      work_mem = {{ ( ansible_memtotal_mb / 400 ) | int }}MB
+    dest: "{{ postgresql_config_path }}/conf.d/tuning.conf"
+  become: yes
+  become_user: postgres
+  notify: restart postgres
+  when: remove_postgres_tuning is undefined
+
+- name: Optionally remove tuning configuration
+  file:
+    path: "{{ postgresql_config_path }}/conf.d/tuning.conf"
+    state: absent
+  when: remove_postgres_tuning is defined
--- a/playbooks/postgres_tuning.yml
+++ b/playbooks/postgres_tuning.yml
@@ -3,30 +3,9 @@
   hosts: ofn_servers
   remote_user: "{{ user }}"
 
-  tasks:
-    - name: Add tuning configuration
-      copy:
-        content: |
-          shared_buffers = {{ ( ansible_memtotal_mb / 4 ) | int }}MB
-          effective_cache_size = {{ ( ( ansible_memtotal_mb / 4 ) * 3 ) | int }}MB
-          maintenance_work_mem = {{ ( ansible_memtotal_mb / 16 ) | int }}MB
-          wal_buffers = 16MB
-          work_mem = 64MB
-        dest: "{{ postgresql_config_path }}/conf.d/tuning.conf"
-      become: yes
-      become_user: postgres
-      notify: restart postgres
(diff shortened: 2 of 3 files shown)


roles/postgres_tuning/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+---
+- name: Add postgres tuning configuration
+  copy:
+    content: |
+      shared_buffers = {{ ( ansible_memtotal_mb / 5 ) | int }}MB
+      effective_cache_size = {{ ( ansible_memtotal_mb / 2 ) | int }}MB
```
mkllnk, line 6: Does this mean that Postgres may use half of the total memory of the server?

mkllnk (approved):
:+1:

roles/postgres_tuning/tasks/main.yml
Matt-Yorkley: My understanding of `effective_cache_size` is that it doesn't change how much memory is actually used, it just gives an indication to Postgres as to how much memory the system has, which Postgres uses when making estimates about how to plan queries.

mkllnk (commented):
Provision all managed servers, at least with the tag `-t postgres_tuning`. That should install those settings.

jibees:
Ok, i will merge and provision all servers then.

jibees:
Run:
```
 ansible-playbook playbooks/provision.yml --limit=all-prod -t postgres_tuning --ask-vault-pass
```

(partial) Results:
```
PLAY RECAP ****************************************************************************************************************************************************************************************************************************************************
app.katuma.org             : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
coopcircuits.fr            : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.be         : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.ca         : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.de         : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.hu         : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.ie         : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.net        : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.org.au     : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.org.nz     : ok=3    changed=2    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
openfoodnetwork.org.uk     : ok=2    changed=0    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0

Vendredi 12 mai 2023  11:14:53 +0200 (0:00:08.446)       0:00:53.112 **********
===============================================================================
restart postgres -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 26.77s
postgres_tuning : Add postgres tuning configuration --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 18.01s
Gathering Facts ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 5.45s
postgres_tuning : Optionally remove tuning configuration ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.74s
```

PR #877, opened 2023-05-05
Update ubuntu server, use same than webapp CI

jibees:
Ref:
https://github.com/openfoodfoundation/openfoodnetwork/blob/3a1579104fc49caf49ad894b1922984259114930/.github/workflows/build.yml#L21

Context: https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   ansible-lint:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


sigmundpetersen:
I also think python 3.8.12 works for 22.04.
Maybe less work for now than bumping the minor version?

jibees:
> I also think python 3.8.12 works for 22.04. Maybe less work for now than bumping the minor version?

I saw that error:
```
 Error: The version '3.8.2' with architecture 'x64' was not found for Ubuntu 22.04.
```

sigmundpetersen:
Yes, and following that link in the error https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json , 3.8.12 is the first on the 3.8 branch to support 22.04.

But if 3.10.4 has a green build, that would be better of course 👍

jibees:
> I also think python 3.8.12 works for 22.04.

Sorry I read `3.8.2` ! Ok, maybe I should do step by step, not sure.

jibees:
I'm not sure how to progress here, maybe @Matt-Yorkley could have great ideas?

PR #881, opened 2023-05-12
Allow more time for cloning order cycles

mkllnk:
This should be reset after the cloning has been made more efficient:

* https://github.com/openfoodfoundation/openfoodnetwork/pull/10832


mkllnk (commented):
Provisioned and deployed.

Issue #876, opened 2023-05-05
Broken build


sigmundpetersen:
Github actions hasn't been running for a while.

![image.png](https://images.zenhubusercontent.com/557dc8f44f7a08fbb7f338ff/223344e3-e2db-4f90-8b8f-45b5c718add3)

We are running on ubuntu 18.04 and this seems to have been removed as a runner option https://github.com/actions/runner-images/issues/6002

PR #883, opened 2023-06-01
Install ruby on deploy

mkllnk:
Preparing for future Ruby upgrades.

* Unblocks https://github.com/openfoodfoundation/openfoodnetwork/pull/10888

There's more work to do. We currently install a specific Ruby version during provisioning which we don't need any more. But this pull request took me long enough to take a break here and continue another time.

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -35,16 +35,14 @@
 
 # Ensure the correct Ruby version is activated for this build
 
-- name: register ruby version in new build
-  slurp:
-    src: "{{ build_path }}/.ruby-version"
-  register: deploy_ruby_version
-  become: yes
-
-- name: use ruby {{ deploy_ruby_version['content'] | b64decode | trim }} # noqa 301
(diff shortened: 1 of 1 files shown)


roles/deploy/tasks/deploy.yml
```diff
@@ -35,16 +35,14 @@
...
-  become: yes
-  become_user: "{{ app_user }}"
+- name: Install required Ruby version if needed
+  command: bash -lc "./script/rbenv-install.sh"
+  args:
+    chdir: "{{ build_path }}"
+  environment:
+    RUBY_CONFIGURE_OPTS: "--with-jemalloc"
+  register: rbenv_install
+  changed_when: rbenv_install.stdout.find("Installing ruby") != -1
```
dacook, line 45: Thanks for improving this, complete with a clever `changed_when`.

dacook (approved):
😎 Very nice, great work.

Unfortunately, it looks like we can no longer run CI with our Ubuntu version. (https://github.com/openfoodfoundation/ofn-install/issues/876)

For this change I suggest we just test on staging. 
Can you confirm if you already have?

mkllnk:
Yes, I tested on staging.

Matt-Yorkley (approved):

Issue #885, opened 2023-06-21
Remove datadog agent


mkllnk:
We cancelled our Datadog subscription and need to stop using it this month. Otherwise we will get charged on-demand rates.

### WIP

* https://github.com/openfoodfoundation/ofn-install/compare/master...mkllnk:ofn-install:datadog

drummer83:
https://github.com/openfoodfoundation/openfoodnetwork/pull/11092 has been merged. Can we close here?

mkllnk:
> https://github.com/openfoodfoundation/openfoodnetwork/pull/11092 has been merged. Can we close here?

No. The datadog agent is a separate process running on the server to monitor resources and log files. We still need to do this. We are still getting charged by Datadog.

mkllnk:
All done.

PR #886, opened 2023-07-21
Disabled datadog agents

mkllnk:
* Prepares https://github.com/openfoodfoundation/ofn-install/issues/885

There's quite a bit to do to remove all our custom Datadog configurations but in the meantime it's very simple to just deactivate the Datadog agent services. fr-prod was the only host using it and I applied this change already. Hopefully Datadog will stop charging us now.


rioug (approved):

dacook (approved):
Looks good. Unfortunately we don't have CI to confirm but I'm assuming you tested this already.

dacook:
I just re-read the description; you already ran this on fr-prod. Merging!

PR #887, opened 2023-07-28
Uninstall Datadog agent

mkllnk:
In my last pull request I disabled the datadog-agent service. In this pull request I'll remove it. Once that has been applied to production and Datadog is gone, we can remove the role and config as well.

Testing out:

- [x] au-staging
- [x] uk-prod (the last server which had an active Datadog agent)

Provisioning all after merge:

- [ ] fr-staging
- [ ] uk-staging
- [ ] all-prod


dacook (approved):
Great cleanup.
I imagine we'll need things like nginx stub_status later, but better to have it installed intentionally when we're using it.

roles/app/templates/application.yml.j2
```diff
@@ -21,10 +21,6 @@ GOOGLE_MAPS_API_KEY: {{ google_maps_api_key }}
 ADMIN_EMAIL: {{ admin_email }}
 ADMIN_PASSWORD: {{ admin_password }}
 
-{% if enable_rails_apm is defined %}
-DATADOG_RAILS_APM: {{ enable_rails_apm }}
-{% endif %}
-
```
rioug, line 27: I am confused as why this duplicated in `roles/app/templates/env.j2` (or maybe it's the other way around ?)

rioug (approved):

rioug:
Test seems to be failing because of a lack of runners, is this something we need to look into ? If not this is ready for testing

roles/app/templates/application.yml.j2
mkllnk: We used to have our configuration in `config/application.yml` and switched to `.env` at some point. So in this transition it's duplicated. But I reckon that we can remove application.yml now. All servers should only be used .env now.

dacook:
> Test seems to be failing because of a lack of runners, is this something we need to look into ? 

Yeah, we need to look into that 😞 . We use a version of Ubuntu that's now only partially supported by GH Actions, so we need to upgrade. 

> If not this is ready for testing

I think the next step is to merge and deploy, which I'll leave for @mkllnk.

PR #890, opened 2023-08-06
test intsll

kefayatullahkhadem:
this is our test intall


mkllnk (commented):
Thank you for sharing your progress.

Your pull request looks like work-in-progress which is not meant to be merged yet. For example, you are replacing the configuration for our French server instead of adding your own.

May I ask which organisation you are working with? What is your goal with this new instance?

PR #895, opened 2023-09-25
Fix the linting offenses and run it on CI successfully

mkllnk:
We have two CI checks. The build is still broken but the linting should work now.


roles/deploy/tasks/deploy.yml
```diff
@@ -64,6 +64,7 @@
     "
   args:
     chdir: "{{ build_path }}"
+    creates: "{{ build_path }}/.bundle/config"
```
dacook, line 67: TIL about this interestingly named argument. I would have expected more generic "if" style code, which I thought ansible also supports.

All good, this is clearly the standard way to do it, it's just new to me.

dacook (approved):
Yay, thank you!

rioug (approved):
Looks good ! merging.

PR #896, opened 2023-09-28
Replace http with https and fake cert in local VMs

mkllnk:
Most local machines had their own nginx config to use only http. But the lxc host was missing that config which led me to these tasks:

* Create a common config for all local containers/VMs/hosts.
* Use the standard nginx config with the snake oil certificate in local containers.
* Get rid of all the duplicate config for the local containers.
* Get rid of outdated development scripts because ofn-install is not for a development environment.


inventory/group_vars/local.yml
```diff
@@ -18,3 +18,21 @@ developer_email: admin@example.com
...
   - "{{ core_devs }}"
+
+# Test host configuration
+# This is usually set in the host config but the same for all local containers.
+domain: localhost
+
+admin_email: ofn@example.com
+mail_domain: example.com
+
+# Add missing vars to emulate /local_vagrant/secrets.yml
```
dacook, line 29: I think the exact path of secrets.yml is unimportant here:
```suggestion
# Add missing vars to emulate secrets.yml
```

inventory/group_vars/local.yml
```diff
@@ -18,3 +18,21 @@ developer_email: admin@example.com
 
 users_sysadmin:
   - "{{ core_devs }}"
+
+# Test host configuration
```
dacook, line 22: Seeing as the whole section below is pertaining to test/local hosts, maybe we can make this look more like section comment:

```suggestion
# *** Test host configuration ***
```
(I'm not sure if we have another convention in this codebase already, or you have a different preference. If so pls update!)

dacook (approved):
Great cleanup! All the updates make sense to me, I love seeing less code, doing the same thing 🤩 

I have a couple of minor comment suggestions, but it's otherwise all good

rioug (approved):
Thanks for cleaning that up :pray: ! I remember having to fix some configuration when I used vagrant to test the new relic setup, but I never had the chance to go back apply my changes. Anyway that's miles better that what I would have done :smile:

rioug:
I cancelled the checks as it wants to use Ubuntu 18 runners which I believe github doesn't offer any more. 
@mkllnk you might want to address David's comment, otherwise it's ready to merge.

mkllnk:
Thank you! I updated the comments. Merging.

PR #900, opened 2023-10-05
Install New Relic infrastructure agent

mkllnk:
I applied this to the three staging servers and the biggest three production servers. Let's see how that goes.


playbooks/provision.yml
```diff
@@ -102,6 +102,11 @@
       become: yes
       tags: logrotate
 
+    - role: newrelic
+      become: yes
+      when: new_relic_api_key is defined
```
dacook, line 107: This role also depends on `new_relic_account_id` and `new_relic_region`, so I was wondering if we should check for those here too. 
But then if you forgot to set any of those, it would just silently fail. 
👍 So I think this way is best: if you intended to switch it on, but left out a detail, then it would be better to get an error message which shows what's missing.

roles/newrelic/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+- name: Download installer
+  ansible.builtin.get_url:
+    url: "https://download.newrelic.com/install/newrelic-cli/scripts/install.sh"
+    dest: "/tmp/new-relic-install.sh"
+    mode: "777"
+
+- name: Install New Relic command
+  ansible.builtin.command: "/tmp/new-relic-install.sh"
+  args:
+    creates: /usr/local/bin/newrelic
```
dacook, line 12: Nice to see some validation of the result. 
I can think of a couple of other ways to validate: 
1. check for a non-error return code, or 
2. grep the output for the success message. 

But I don't think it's worth spending more effort on figuring one of those out.

roles/newrelic/tasks/main.yml
```diff
@@ -19,3 +19,11 @@
...
   args:
     creates: /usr/bin/newrelic-infra
+
+- name: Use our standard names in dashboard
+  lineinfile:
+    name: "/etc/newrelic-infra.yml"
+    line: "display_name: {{ host_id }}"
+    regex: "^display_name:"
+    state: present
+  notify: restart newrelic_infra
```
dacook, line 29: Nice one!

dacook (approved):
Great!

dacook:
✅  Green build!

PR #901, opened 2023-10-06
Rename host group names for valid Ansible names

mkllnk:
Ansible was complaining:
```
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
```

Dashes are not allowed in group names like au-prod.


dacook:
@mkllnk unfortunately this broke the staging deployments (see mentioned issue above). I think because the script is relying on the fact that the host_id was the same as the hosts group ID.

PR #903, opened 2023-10-09
Fix CI build on Github Actions

mkllnk:
The build takes best part of an hour but it does run the whole suite and makes sure that the app is running on a fresh container.


dacook (approved):
💐 Great job!!! 🎉 

It's worth noting that any other devs will need to run `bin/setup` to download the latest requirements. 
But I think I'm the only other dev using this at the moment, so I'll just do that myself now ;)

PR #904, opened 2023-10-11
Rename host group names for valid Ansible names

mkllnk:
Ansible was complaining:
```
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
```

Dashes are not allowed in group names like au-prod.

This requires an update of the staging servers. Last time we tried it, we ran into:

* https://github.com/openfoodfoundation/openfoodnetwork/issues/11644

We need to update the deploy scripts for staging after merging this:

```
ansible-playbook playbooks/setup_semaphore_deployment.yml -l au_staging,fr_staging,uk_staging -t provision_deploy_script
```

### DIFF
--- a/inventory/hosts
+++ b/inventory/hosts
@@ -22,215 +22,215 @@ test
 #------------------------------------------------------------------------------
 # Australia
 
-[au-prod]
+[au_prod]
 openfoodnetwork.org.au ansible_host=43.239.97.238
 
-[au-staging]
+[au_staging]
 staging.openfoodnetwork.org.au
 
 [au:children]
-au-prod
-au-staging
+au_prod
+au_staging
 
 #------------------------------------------------------------------------------
 # Brazil
 
-[br-prod]
+[br_prod]
 openfoodbrasil.com.br ansible_host=52.20.5.42
 
-[br-staging]
+[br_staging]
 staging.openfoodbrasil.com.br ansible_host=54.242.172.26
 
 [br:children]
-br-prod
-br-staging
+br_prod
+br_staging
 
 #------------------------------------------------------------------------------
 # Colombia
 
-[co-prod]
+[co_prod]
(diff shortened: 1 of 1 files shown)


dacook (approved):

dacook:
That's interesting, it seems like a clever hack that the `limit` param is copied into the script:

https://github.com/openfoodfoundation/ofn-install/blob/master/roles/semaphore_deployment/templates/deploy.j2#L36

I wonder if it's even necessary; I would think that `--connection local` ensures it only runs locally.

mkllnk:
My command was actually not right. It inserted all three hosts into the command on all servers. But that will test if the local connection is restricting correctly. In that case we could get rid of that parameter.

mkllnk:
So, I ran a deploy on au-staging and got a "failure":

```
PLAY RECAP *********************************************************************
staging.coopcircuits.fr    : ok=31   changed=16   unreachable=0    failed=0    skipped=6    rescued=0    ignored=0   
staging.openfoodnetwork.org.au : ok=9    changed=2    unreachable=0    failed=1    skipped=1    rescued=1    ignored=0   
staging.openfoodnetwork.org.uk : ok=9    changed=5    unreachable=0    failed=1    skipped=1    rescued=1    ignored=0   
```

It tried to deploy the local machine (au-staging) three times in parallel. Removing old releases failed on a conflict. One host reported success and the others failed. It didn't connect to fr or uk though. If we didn't have the limit in, it would try to execute the deploy script as many times as we have hosts.

dacook:
Thanks for checking that out, now we know for sure! We should document how to run this playbook somewhere.

dacook:
I just had an idea: rather than `ansible_limit`, we should be able to reference the current hostname. 

But hopefully we won't need to use it again for a long time, so I won't bother trying it now.

PR #906, opened 2023-10-20
Retire JB's ssh access

mkllnk:
JB left the organisation and I removed his SSH key from all servers already:

```
ansible-playbook playbooks/remove_ssh_keys.yml --limit all_prod,all_staging -e "{'remove_users_sysadmin':[jb]}"
```

This change makes sure that we don't add it again.

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -240,7 +240,6 @@ core_devs:
   - maikel
   - matt
   - filipe
-  - jb
   - david
   - gaetan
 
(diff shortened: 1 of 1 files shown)


rioug (approved):
:+1:

dacook (approved):
👍  Thanks!

I note that we don't actually remove the key from this repo. I can't think of a good reason to remove it, or keep it, so will do nothing ;)

mkllnk:
We need to keep the key to remove it from files. Otherwise we don't know which line to remove. But we could prune the retired keys at some point. And I could imagine that there are many other unused keys in the folder.

PR #907, opened 2023-11-02
Add India configuration

duleorlovic:
Adding configuration for India.
Please let me know if we should split the changes in different PR (one for key, one for templates/env.j2 ...)


inventory/group_vars/in.yml
```diff
@@ -0,0 +1,18 @@
...
+
+# See https://github.com/openfoodfoundation/ofn-install/wiki/Setup for more info
+checkout_zone:
+country_code: IN
+currency: INR
+locale: en
+language: en
+language_packages:
+  - language-pack-en-base
+l10n_repo: https://github.com/duleorlovic/l10n_in.git
```
dacook, line 11: I notice this is your fork. Technically it's ok to use your own fork, but I can see it's simply to fix a small error in the data, so I think it would be better to submit your update to the `openfoodnetwork` repo. 
I've merged your update and granted you access to the repo, so would you like to point to that repo?

roles/app/templates/env.j2
```diff
@@ -14,6 +14,7 @@ CURRENCY="{{ currency }}"
 CHECKOUT_ZONE="{{ checkout_zone }}"
 
 GOOGLE_MAPS_API_KEY="{{ google_maps_api_key }}"
+GOOGLE_MAPS_REGION="{{ google_maps_region }}"
```
dacook, line 17: Will this be provided in the secrets file? 
If it's not a secret, you could consider including it in `inventory/group_vars/in.yml`

dacook: Actually, this is a new variable which has not been set for other countries, so it would create an error. Can you please update it to include a condition (like `
{% if s3_access_key is defined %}` below)?

dacook (changes requested):
Great job, it looks like we are nearly ready to support deployments to the server!

I noticed a couple of things, can you please take a look and update accordingly?

inventory/hosts
```diff
@@ -93,7 +93,7 @@ ie_prod
 # India
 
 [in_prod]
-openfoodnetwork.in
+openfoodnetwork.in ansible_user=ubuntu
```
mkllnk, line 96: I think that this is only required to start with, or it was required before your key was added. Once your key is in the ofn-admin account, this user isn't needed any more. This setting will prevent other developers to access the server unless they override it with ofn-admin, I think. I'm not 100% sure about precedence.

roles/app/templates/env.j2
mkllnk: Unfortunately, the openfoodnetwork code doesn't guard against an empty string either. Otherwise that would have been valid to define an empty string here.

mkllnk (changes requested):

inventory/group_vars/in.yml
duleorlovic: Great, I have updated the link so now it is OFN repo

inventory/hosts
duleorlovic: Yes, it is not required for `playbooks/deploy.yml` so I removed and I have successfully performed
```
ansible-playbook playbooks/deploy.yml --limit in_prod -e "git_version=v4.4.17"
```

Do you need to run setup also ?
When I try setup there is an error
```
ansible-playbook playbooks/setup.yml --limit in_prod -e "git_version=v4.4.17"

PLAY [python check] ****************************************************************************************************************

TASK [Install python] **************************************************************************************************************
Friday 03 November 2023  09:48:35 +0100 (0:00:00.017)       0:00:00.017 ******* 
fatal: [openfoodnetwork.in]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: root@openfoodnetwork.in: Permission denied (publickey).", "unreachable": true}

PLAY RECAP *************************************************************************************************************************
openfoodnetwork.in         : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0   
```
but probably you need only to deploy... Not sure if we need to add some system dependency or upgrade some package, probably I need to do that, right ?

roles/app/templates/env.j2
duleorlovic: `GOOGLE_MAPS_REGION` is used [on the same line](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/views/shared/_google_maps_js.html.haml) as `GOOGLE_MAPS_API_KEY` so I was thinking that we should put close to each other.

Why do you think it will cause an error ?
If it is blank it will not add region parameter.
```
ENV['GOOGLE_MAPS_REGION'] ? '&region=' + ENV['GOOGLE_MAPS_REGION'] : ''
```

duleorlovic (commented):
I have force pushed the updates to my branch.

roles/app/templates/env.j2
dacook: > `GOOGLE_MAPS_REGION` is used [on the same line](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/views/shared/_google_maps_js.html.haml) as `GOOGLE_MAPS_API_KEY` so I was thinking that we should put close to each other.

Good point, I agree with that reasoning.

dacook (changes requested):
Almost there, but I have one change to request as commented.

inventory/hosts
dacook: Hmm, it looks like the first-time setup playbook needs to log in as `root`, in order to create other users.

But I guess you've already executed the `setup.yml` playbook? In that case you don't need to run it again. 
All other playbooks (for example `provision.yml`) should still work, using the default user `ofn-admin`.

roles/app/templates/env.j2
duleorlovic: Created an [issue](https://github.com/openfoodfoundation/openfoodnetwork/issues/11783) and PR for .present?
Added `{% if google_maps_region is defined %}`

inventory/hosts
duleorlovic: @dacook  OK, we can keep without `ansible_user=ubuntu` since setup is already performed. Do you connect to other instances using `root` or only using those generated users `ofn-admin`, `openfoodnetwork` ?

dacook (approved):
Great! This passes CI and looks good to me.

inventory/hosts
dacook: I think we only connect using those generated users. I'm not sure, but I think we usually disable SSH access for the root user.

mkllnk (approved):
All looking good now. Thank you!

PR #911, opened 2023-11-09
Restart after config changes

dacook:
Ensure relevant services are restarted, after updating Rails or Postgres environment variables. Normally they would be restarted when the full provision playbook is executed anyway, but my vision is to be able to perform individual actions quicker. 
My vision: to update app config in a minute with`--tags=app`, instead of waiting ages for every single action to run. 

Probably this could have been separated into two separate tasks (one for Rails and one for Postgres). But it didn't seem worth spending more time on. Especially because additional tasks make ansible even slower...

I'm not sure why it failed on CI.
Update: I thought it was time to act on this one, so found a way to avoid the sidekiq error. Hmm, perhaps it was never started on initial setup? Then we should probably fix that.. but is this ok for now?

### DIFF
--- a/roles/shared_handlers/handlers/main.yml
+++ b/roles/shared_handlers/handlers/main.yml
@@ -1,8 +1,8 @@
 # This role holds reusable handlers that can be included in multiple playbooks to keep things DRY
 
-- name: check service status
+- name: fetch services status # into the global variable ansible_facts.services
   service_facts:
-  listen:
+  listen: # run if the following tasks have been notified:
     - restart sidekiq
 
 - name: restart puma
--- a/roles/shared_handlers/handlers/main.yml
+++ b/roles/shared_handlers/handlers/main.yml
@@ -13,3 +13,10 @@
     state: restarted
   become: yes
   become_user: root
+
+- name: restart sidekiq
+  service:
+    name: sidekiq
+    state: restarted
+  become: yes
+  become_user: root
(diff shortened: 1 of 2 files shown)


dacook:
```
Unable to start service sidekiq: Job for sidekiq.service failed because the control process exited with error code.\nSee \"systemctl status sidekiq.service\" and \"journalctl -xe\" for details
```

dacook:
I accidentally just "tested" this on us_prod.

Thankfully, it works 😌 ✅

roles/shared_handlers/handlers/main.yml
```diff
@@ -1,5 +1,10 @@
 # This role holds reusable handlers that can be included in multiple playbooks to keep things DRY
 
+- name: check service status
+  service_facts:
+  listen:
+    - restart sidekiq
+
```
rioug, line 7: I am not what this is used for ? Isn't that check done by this line below
```
 when: ansible_facts.services['sidekiq.service']['state'] == 'running'
```
I am not really familiar with ansible so I might be missing context.

dacook: It took me a while to understand it too, probably should add comments.

```suggestion
- name: fetch service status # into the global variable ansible_facts.services
  service_facts:
  listen: # run if the following tasks have been notified:
    - restart sidekiq

```

rioug: Yeah that makes it easier to understand :+1:

rioug (approved):

mkllnk (approved):
Great!

PR #914, opened 2023-11-26
Use hi locale

duleorlovic:
Depends on https://github.com/openfoodfoundation/openfoodnetwork/pull/11856


mkllnk (approved):
Great, this needs provisioning with `-t app`.

mkllnk:
Oh, I see that this has been provisioned already but the latest release doesn't include your other PR yet. The deployment fails because it can't find the locales. I think you have to wait a week for the next release.

PR #916, opened 2024-01-03
Reduce puma timeout to 2min on production

### LINKED ISSUE #859: Reduce Puma worker timeouts
## Description

Timeouts have been increased on several instances to allow for long-running reports. This is now been moved into a separate process and we can reduce the timeout again. I would suggest the default timeout of 30 seconds. So we can remove host-specific timeouts from the inventory.

Follow up from:

- https://github.com/openfoodfoundation/openfoodnetwork/issues/10282

mkllnk:
Since we run reports in the Sidekiq background process, response times are a lot more reasonable.

* Closes #859.

The original intention was to go back to 30 seconds timeout but some shop pages are still slow to load from time to time. We may be able to reduce the timeouts further when improving the performance of the app.


rioug (approved):
Nice :+1:

dacook (approved):
Great cleanup and thanks for fixing the missing role.

I'm guessing this was manually added to override the gitignore rule. Which is fine, I can't think of any downsides to that except being potentially confusing.

dacook:
This changes some staging config too, so we could test at least one of those before merging.

mkllnk:
I'll do all of them:

- [ ] au-staging
- [ ] fr-staging
- [ ] uk-staging

PR #920, opened 2024-03-11
Pass in configured domain to ActionCable - related to #12020

abdulazizali77:
**What? why?**

- Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/12020
- ActionCable is unable to modify the query string due to the difference in domains. In the `products_reflex.rb#current_url` the `request.original_url` gets set to `https://rails` instead of the configured domain because no `Host` header is sent
- Send relevant headers in the action cable location block

**What should we test**
- Currently no tests written yet
- Staging was manually tested by @mkllnk and worked https://github.com/openfoodfoundation/ofn-install/pull/920#discussion_r1520582714
- We should test prod to confirm fix addresses issue, hence [#12020](https://github.com/openfoodfoundation/openfoodnetwork/issues/12020) should remain open until that fact

**References**
https://stackoverflow.com/questions/67696655/what-do-i-need-to-do-to-hook-up-actioncable-on-nginx-and-puma

https://github.com/rails/rails/issues/29893

**Followup**
@dacook was able to reproduce issue in dev https://github.com/openfoodfoundation/openfoodnetwork/issues/12020#issuecomment-1958175366 however without the same error messages hence there needs to be a followup to investigate that behaviour

### DIFF
--- a/inventory/group_vars/all.yml
+++ b/inventory/group_vars/all.yml
@@ -397,6 +397,7 @@ nginx_sites:
         proxy_set_header X-Forwarded-Ssl on;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
+        proxy_set_header Host $host;
       }
 
       client_max_body_size 4G;


inventory/group_vars/all.yml
```diff
@@ -397,6 +397,9 @@ nginx_sites:
         proxy_set_header X-Forwarded-Ssl on;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $host;
```
mkllnk, line 402: I tested these three settings on a staging server and the `Host` header made a difference. We don't need the other two to solve the issue. Could you modify your PR to only add the one host header line?

I'm cautious about the forwarded-for header because of the security issue you referenced. We would need to add more to our configuration there. And maybe we won't need it.

mkllnk (changes requested):

inventory/group_vars/all.yml
abdulazizali77: thanks for testing @mkllnk 
have removed the other two headers

mkllnk (approved):

abdulazizali77:
@mkllnk what else should be done for this PR?

mkllnk:
I'm waiting for a second review. Then we can merge.

rioug (approved):

dacook (approved):
Looks like a logical config to me 👍 

@mkllnk it appears that you've tested this before and after on staging to confirm. I guess you could see that the error was no longer thrown?

It's worth pointing out that I was able to [replicate this issue in dev](https://github.com/openfoodfoundation/openfoodnetwork/issues/12020#issuecomment-1958175366), without other error messages, so I'm not sure if this solves the issue conclusively. 

This is good to merge, so I suggest we merge and ~~release~~ deploy it, but keep the issue open until it can be confirmed fixed in prod. If you're ok with that can you please edit the description before merging to prevent auto closing of the issue?

abdulazizali77:
@dacook have modified the title and description though i do not know what string/phrase the github pipeline picks up to autoclose an Issue linked to a PR. Would appreciate if you can modify the title and description where you see appropriate, thanks!

mkllnk:
Provisioning:

- [ ] au_staging                                                                      
- [ ] uk_staging                                                                      
- [ ] fr_staging 
- [ ] be_prod                                                                         
- [ ] es_prod                                                                         
- [ ] fr_prod                                                                         
- [ ] uk_prod                                                                         
- [ ] de_prod                                                                         
- [ ] ie_prod                                                                         
- [ ] hu_prod 
- [ ] ca_prod                                                                         
- [ ] us_prod                                                                         
- [ ] au_prod                                                                         
- [ ] nz_prod                                                                         
- [ ] in_prod

PR #921, opened 2024-03-18
Update locale settings

dacook:
This was to avoid multiple Englishes appearing

* Reported on https://github.com/openfoodfoundation/openfoodnetwork/issues/11513

There still seems to be some other issue, but I think this is still necessary.

This has already been provisioned on au_staging and uk_staging.

### DIFF
--- a/inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
@@ -13,7 +13,7 @@ users_sysadmin:
   - konrad
 
 # activate the language switcher
-available_locales: en,fr,es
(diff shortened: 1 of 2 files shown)


inventory/host_vars/staging.openfoodnetwork.org.au/config.yml
```diff
@@ -13,7 +13,7 @@ users_sysadmin:
   - konrad
 
 # activate the language switcher
-available_locales: en,fr,es
+available_locales: en_AU,fr,es
```
dacook, line 16: Because in `au.yml`:
```yaml
locale: en_AU
language: en_AU.UTF-8
```

mkllnk (approved):
Great!

PR #932, opened 2024-07-30
Add RACK_TIMEOUT_TERM_ON_TIMEOUT for uk-prod

rioug:
UK prod is getting a rather high number of Rack Timeout errors : https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/664378ec33e1080008ba685a?filters[error.status]=open&filters[event.since]=30d . Most likely due to long running queries.

As explained in the docs : https://github.com/zombocom/rack-timeout/blob/master/doc/settings.md#term-on-timeout it can cause the application to enter a in a corrupt state, and can cause database connection to not be released back in the db connection pool. This is most likely what has been happening on uk prod, we are seeing a high number of `ActiveRecord::ConnectionTimeoutError` see https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/664b458a96db3b0007ce304a?filters[error.status]=open&filters[event.since]=30d

To mitigate the issue we can RACK_TIMEOUT_TERM_ON_TIMEOUT configuration which will reboot the puma worker after the number of timeout set by the config.  

It doesn't solve the underlying issue of long running request, but it should help.

PR #934, opened 2024-08-19
Add RACK_TIMEOUT_TERM_ON_TIMEOUT for fr-prod

rioug:
Part of https://github.com/openfoodfoundation/openfoodnetwork/issues/12761

same as `uk-prod` #932, adding RACK_TIMEOUT_TERM_ON_TIMEOUT to mitigate issue with the db connection pool.


dacook (approved):
So we're pretty confident that this resolves the issue. 
In that case, shall we make this the default for all instances?

mkllnk (approved):

PR #942, opened 2024-10-07
[Citi OFN Voucher] Add Activerecord encryption configuration and VINE api config

rioug:
**⚠️ When working on this use the Clockify code: Citi Food Vouchers: #1A. https://github.com/openfoodfoundation/openfoodnetwork/issues/11922 Citi OFN Voucher - VINE Integration ⚠️**

Related to : https://github.com/openfoodfoundation/openfoodnetwork/pull/12886

Update env template and secret example to include activerecord encryption configuration and VINE_API_URL config.

### DIFF
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -94,6 +94,3 @@ smtp_password:
 #active_record_encryption_primary_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_deterministic_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_key_derivation_salt: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-
-# VINE API settings
-#vine_api_url: "https://vine-url.com/api/v1"
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -94,3 +94,6 @@ smtp_password:
 #active_record_encryption_primary_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_deterministic_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_key_derivation_salt: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+
+# VINE API settings
+#vine_api_url: "https://vine-url.com/api/v1"
--- a/inventory/host_vars/_example.com/secrets.example.yml
+++ b/inventory/host_vars/_example.com/secrets.example.yml
@@ -88,3 +88,9 @@ smtp_password:
 #new_relic_agent_enabled: true
 #new_relic_app_name: "Open Food Network"
 #new_relic_license_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+
+# database encryption configuration, required for vine connected app
+# generate with bin/rails db:encryption:init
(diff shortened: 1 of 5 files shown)


dacook (changes requested):
`AnsibleUndefinedVariable: 'active_record_encryption_primary_key' is undefined`
```
TASK [app : template files] ****************************************************
failed: [local_test] (item={'src': 'env.j2', 'dest': '/home/openfoodnetwork/apps/openfoodnetwork/shared/config/.env.staging'}) => {"ansible_loop_var": "item", "changed": false, "item": {"dest": "/home/openfoodnetwork/apps/openfoodnetwork/shared/config/.env.staging", "src": "env.j2"}, "msg": "AnsibleUndefinedVariable: 'active_record_encryption_primary_key' is undefined"}
changed: [local_test] => (item={'src': 'postgresql.yml.j2', 'dest': '/home/openfoodnetwork/apps/openfoodnetwork/shared/config/database.yml'})
```

We probably need to add conditions for these variables, like others in this file. 

I'd love to see a cleaner way to do it than what we currently have though. I can think of a couple of options. If we're happy to have these ENV vars defined but blank, we can:
* Set a default blank value for these variables in `all.yml`. Or
* Use Jinja's default function, eg `{{ active_record_encryption_primary_key | default('') }}` (https://stackoverflow.com/a/35107741/421243)

rioug:
Thanks for checking @dacook.

I ended up adding VINE_API_URL config as well, which I had forgotten.

inventory/host_vars/_example.com/secrets.example.yml
```diff
@@ -94,3 +94,6 @@ smtp_password:
 #active_record_encryption_primary_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_deterministic_key: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 #active_record_encryption_key_derivation_salt: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+
+# VINE API settings
+#vine_api_url: "https://vine-url.com/api/v1"
```
dacook, line 99: Do you think this would be ok to be a public var, defined in inventory (eg in host_vars/staging.*/config.yml)?

It should be easier to manage in the long run.

dacook (commented):
Sorry one more question

inventory/host_vars/_example.com/secrets.example.yml
rioug: Fixed now

dacook (approved):

roles/app/templates/env.j2
```diff
@@ -102,4 +102,18 @@ NEW_RELIC_APP_NAME="{{ new_relic_app_name }}"
...
+{% if active_record_encryption_deterministic_key is defined %}
+ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY="{{ active_record_encryption_deterministic_key }}"
+{% endif %}
+{% if active_record_encryption_key_derivation_salt is defined %}
+ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT="{{ active_record_encryption_key_derivation_salt }}"
+{% endif %}
+
+{% if vine_api_url is defined %}
+VINE_API_URL="{{ vine_api_url }}"
+{% endif %}
```
mkllnk, line 117: Do you think that a default would work instead?
```suggestion
VINE_API_URL="{{ vine_api_url | default('') }}"
```

Just a question, not a real suggestion.

mkllnk (approved):
:+1:

roles/app/templates/env.j2
rioug: The code handle a missing `VINE_API_URL` with an error saying the VINE api isn't configured. With an empty string default it would result in a API connection error which could be confusing when the issue is actually a missing configuration.

mkllnk: Ah yes, the code would need to check for the presence of the variable instead.

PR #945, opened 2024-10-16
Add swapfile to fr_staging

dacook:
- https://openfoodnetwork.slack.com/archives/CEBMTRCNS/p1726451615941249

Last month the memory usage hit 100% and interrupted testing. It  hasn't happened again, but I think it's worth having as a backup. Then the server will just slow down instead of crashing, which should still be enough of a canary to warn that something's going wrong. There's plenty of free disk space.

This hasn't been executed yet. 

## Action required after merge
Once approved I'll provision, using tag `swapfile`.


mkllnk (approved):

dacook:
```
ansible-playbook playbooks/provision.yml --limit=fr_staging --tags=swapfile
...
staging.coopcircuits.fr    : ok=8    changed=5    unreachable=0    failed=0    skipped=2    rescued=0    ignored=0
```

PR #951, opened 2024-10-27
Use IP of new Ubuntu 20 server for IE

cillian:
openfoodnetwork.ie has been upgraded to Ubuntu 20 now by following the instructions at https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server which were a great help. 

I noticed that the `pull_assets.yml` and `push_assets.yml` playbooks don't include the `storage` directory [here](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/migrate_assets/defaults/main.yml). If an instance is configured to use `local` ActiveStorage this can be an issue, because I had to manually copy over the `storage` directory to the new server,  although most instances probably don't use `local`.

Let me know if you need any more details.

### DIFF
--- a/inventory/hosts
+++ b/inventory/hosts
@@ -88,7 +88,7 @@ hu_prod
 ## Ireland
 
 [ie_prod]
-openfoodnetwork.ie ansible_host=167.172.44.121
+openfoodnetwork.ie ansible_host=64.227.64.162
 
 [ie:children]
 ie_prod


rioug (approved):
:+1:

dacook (approved):
Great work! FYI I'm also working on upgrading some servers to Ubuntu 20.

> I noticed that the `pull_assets.yml` and `push_assets.yml` playbooks don't include the `storage` directory [here](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/migrate_assets/defaults/main.yml).

I came across this too, because there's at least one other server using local storage. I aim to update the playbook to handle it better in future.

Just wondering if you have any additional backup strategy for the local files?

cillian:
@dacook Thanks. I backup up the local files periodically but I should probably automate it at some point.

dacook:
Fair enough. I think that is already more comprehensive than another server I was looking at recently..

mkllnk:
> although most instances probably don't use local.

Some files like reports are always stored locally because their lifetime is relatively short and we want a very fast response once the file is generated.

cillian:
:+1: Good to know.

PR #955, opened 2024-10-31
Add checklist for server migrations

dacook:


.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -0,0 +1,97 @@
...
+- [ ] Review `host_vars/x/config.yml`, clean up if needed
+  - Make a copy for the temp hostname, add temp domain to bottom of `certbot_domains`
+- [ ] Review `ofn-secrets:x_prod/secrets.yml`, clean up if needed
+   - Change to shared bugsnag projects
+   - Don't bother making a copy of this one
+
+### setup
+Ensure you have the correct secrets.
+`ansible-playbook -l x_prod2 -e "@../ofn-secrets/x_prod/secrets.yml playbooks/`
+- [ ] `letsencrypt_proxy.yml`
```
mkllnk, line 28: Isn't this one run on the old server?

```
ansible_playbook playbooks/letsencrypt_proxy --limit current-prod -e "proxy_target=new-prod.org"
```

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -0,0 +1,97 @@
...
+- [ ] `setup.yml`
+- [ ] `provision.yml`
+- [ ] `deploy.yml`
+- [ ] `db_integrations` (Permit DB access for n8n, Metabase)
+- [ ] Ensure sidekiq is disabled, to avoid creating subscription orders: `systemctl disable sidekiq`
```
mkllnk, line 33: Disable and stop in this case.

But I wonder if we should avoid copying production data before the switchover. :thinking: 

I guess it's part of the switchover and if problems are identified then we can fix it, go back and start again.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -0,0 +1,97 @@
...
+ - [ ] shop catalogue display correctly, with images, add to cart, begin checkout, login
+  - note: check cookies if login won't work
+ - Check integrations 
+   - [ ] Payments (check Stripe connect status `/admin/stripe_connect_settings/edit`)
+   - [ ] New Relic
+   - [ ] Bugsnag
+
+## 3. Migration
+### preparation
+- [ ] `deploy.yml -l x_prod2 -e "git_version=vX.Y.Z"` matching version with current prod
```
mkllnk, line 54: I would make sure to have an empty database before deploying anything on the new server. A deployment is starting Sidekiq and may have other side effects in migrations we don't know yet.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -0,0 +1,97 @@
...
+- [ ] old server: make a tiny data change to verify later (eg add `.` in meta description `/admin/general_settings/edit`)
+
+### switchover: old server
+- [ ] 🚧 `maintenance_mode.yml`
+- [ ] `sudo systemctl stop sidekiq redis-jobs`
+- [ ] Transfer `/var/lib/redis-jobs/dump.rdb` to new server (see guide)
+- [ ] `db_transfer.yml` ~3min
+- [ ] `transfer_assets.yml` just in case
+- [ ] `sudo systemctl stop puma postgres` (ensure other integrations no longer touch it)
```
mkllnk, line 63: This can go one step up right after db transfer. And we could stop puma above with sidekiq, right?

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -0,0 +1,97 @@
...
+- [ ] shut down the old server, cancel old VPS
+- [ ] remove DNS for temporary subdomain
+- [ ] make sure the entries in ofn-install are up to date: remove the temporary entry made for the migration, and set the new IP address. 
+- [ ] validate that `provision.yml` still works. This will rename x-prod2 to x-prod
+- [ ] check metabase sync if required: https://data.openfoodnetwork.org.uk/admin/databases/
+- [ ] check n8n
+- [ ] check backups are functioning
+- Update documentation: 
+  * [ ] https://github.com/openfoodfoundation/ofn-install/wiki/Current-servers
+  * migration guide if necessary
```
mkllnk, line 90:
```
 * [ ] migration guide if necessary
```

mkllnk (changes requested):
Good idea!

.github/ISSUE_TEMPLATE/server-migration.md
dacook: Hmm I guess it's a question of if you want to separate out the testing from the switchover. Ideally, we wouldn't need to: everything just works and we setup, copy db and proceed to migrate and it's all good.

But we need to test and validate that things are working. And often, there will be something not working.
I'd rather identify that before co-ordinating and communicating a downtime, and putting the old site into maintenance mode. 
Maybe for some smaller instances, we don't need that added layer of testing, and can accept more downtime while we frantically fix things quickly in-place. That might be a way more efficient use of time. (we ended up doing that for CA anyway!)
But I don't think that would be appropriate in all cases. Particularly for CA, I think it would have been too much to deal with all at once and we may have had to abort.

.github/ISSUE_TEMPLATE/server-migration.md
dacook: Yes, I think you're right!

.github/ISSUE_TEMPLATE/server-migration.md
dacook: Good pickup thanks, I must have mixed it up when formatting the list.

.github/ISSUE_TEMPLATE/server-migration.md
dacook: But if sidekiq is `disabled`, it wouldn't get started would it? Hmm unless ansible automatically enables it for you.

The database will be dropped by the subsequent `db_transfer.yml` anyway. So I don't see the need to empty the database, except to perhaps save time if there are any migrations. But if you think we'll avoid unknown side-effects, that's a good enough reason to do it.

dacook:
Thanks for the feedback @mkllnk , I've updated as per comments.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -23,16 +23,20 @@ Checklist based on general guide https://github.com/openfoodfoundation/ofn-insta
    - Don't bother making a copy of this one
 
 ### setup
-Ensure you have the correct secrets.
+Enable passthrough on _current_ server to allow new server to generate a certificate:
+- [ ] `ansible-playbook playbooks/letsencrypt_proxy.yml -l x_prod -e "proxy_target=<ip>" `
```
dacook, line 27: Hmm better to clarify which IP

```suggestion
- [ ] `ansible-playbook playbooks/letsencrypt_proxy.yml -l x_prod -e "proxy_target=<new_ip>" `
```

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -23,16 +23,20 @@ Checklist based on general guide https://github.com/openfoodfoundation/ofn-insta
...
 - [ ] `deploy.yml`
 - [ ] `db_integrations` (Permit DB access for n8n, Metabase)
-- [ ] Ensure sidekiq is disabled, to avoid creating subscription orders: `systemctl disable sidekiq`
 
 ### initial migration
+- [ ] Ensure sidekiq is disabled, to avoid creating subscription orders when data is loaded:
+    `sudo systemctl stop sidekiq && sudo systemctl disable sidekiq`
+
 `ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/`
 - [ ] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per guide
```
dacook: Move list item to preceeding list.
```suggestion
- [ ] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per guide

`ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/`
```

rioug (approved):
Looks good !

dacook:
I think this can be merged now, but I will gladly take further feedback.

PR #956, opened 2024-11-04
Fix remote deployment script

dacook:
This issue was caused because I took the shortcut to provision all three servers at once:
```
ansible-playbook playbooks/setup_remote_deployment.yml -l all_staging
```

The playbook assumed that the `limit` param only refers to the current server, and pastes it into the deploy script. So I've updated it to specify only a reference to itself.

### DIFF
--- a/roles/remote_deployment/templates/deploy.j2
+++ b/roles/remote_deployment/templates/deploy.j2
@@ -33,4 +33,4 @@ cd "/home/{{ deployment_user }}/ofn-install"
 git pull
 
 # Deploy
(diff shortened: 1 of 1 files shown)


dacook:
I've tested on uk_staging. All looks good 
* https://github.com/openfoodfoundation/openfoodnetwork/actions/runs/11656700860/job/32453178110 ✅ 


```
ofn-admin@uk-staging:~$ sudo less /home/ofn-deploy/deploy
...
# Deploy
ansible-playbook playbooks/deploy.yml --limit staging.openfoodnetwork.org.uk --connection local -e "git_repo=https://github.com/openfoodfoundation/openfoodnetwork.git git_version=$deploy"
(END)
```

rioug (approved):
:+1:

PR #964, opened 2024-11-20
Upgrade updates

dacook:
Is it ok not to have `netdev` group?

### DIFF
--- a/roles/default_user/tasks/main.yml
+++ b/roles/default_user/tasks/main.yml
@@ -2,7 +2,7 @@
 - name: create deploy user
   user:
     name="{{ user }}"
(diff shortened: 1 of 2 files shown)


roles/default_user/tasks/main.yml
```diff
@@ -2,7 +2,7 @@
 - name: create deploy user
   user:
     name="{{ user }}"
-    groups=adm,dialout,cdrom,floppy,audio,dip,video,plugdev,netdev,sudo
+    groups=adm,dialout,cdrom,floppy,audio,dip,video,plugdev,sudo
```
mkllnk, line 5: Totally fine. We have sudo for everything we want to do.

mkllnk (approved):
Nice!

PR #967, opened 2024-11-28
Upgrade checklist updates

dacook:
Some more updates to the upgrade checklist and readme.
I'd like to update the process further and simplify it.. but haven't got time for that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/server-migration.md
+++ b/.github/ISSUE_TEMPLATE/server-migration.md
@@ -98,7 +98,8 @@ Make sure to clear cache so that instance settings are applied:
 
 ## 4. Cleanup (after 48hrs)
 - [ ] check server access logs to verify no traffic
-- [ ] shut down the old server, cancel old VPS
+- [ ] shut down the old VPS
+- [ ] delete old VPS (or rename for future deletion)
 - [ ] remove DNS for temporary subdomain
 - [ ] check metabase sync if required: https://data.openfoodnetwork.org/admin/databases/
 - [ ] check n8n
--- a/.github/ISSUE_TEMPLATE/server-migration.md
+++ b/.github/ISSUE_TEMPLATE/server-migration.md
@@ -9,6 +9,8 @@ assignees: ''
 
 Checklist based on general guide https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server
 
+Tip: find/replace to set up most commands ready to go, eg: `x_prod` -> `ca_prod`
+
 ## 1. Setting up the new server
 - [ ] Check old server config for any additional services to be aware of. Document any necessary steps for migration. Eg:
   - `ls /etc/nginx/sites-enabled`
@@ -20,6 +22,7 @@ Checklist based on general guide https://github.com/openfoodfoundation/ofn-insta
 - [ ] Add temporary name to `inventory/hosts`
 - [ ] Review `host_vars/x/config.yml`, clean up if needed
   - Make a copy for the temp hostname, add temp domain to bottom of `certbot_domains`
+- [ ] Review `group_vars/x.yml`, clean up if needed
 - [ ] Review `ofn-secrets:x_prod/secrets.yml`, clean up if needed
    - Change to shared bugsnag projects
    - Don't bother making a copy of this one
@@ -41,7 +44,7 @@ Then setup new server. Ensure you have the correct secrets (current secrets are 
 - [ ] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per guide
 
 `ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/`
-- [ ] `db_transfer.yml`
+- [ ] `db_transfer.yml` &&
 - [ ] `transfer_assets.yml`
 
 Make sure to clear cache so that instance settings are applied:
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -17,9 +19,10 @@ Checklist based on general guide https://github.com/openfoodfoundation/ofn-insta
 - [ ] DNS: add temporary domain (eg `prod2.openfoodnetwork.org`)
 
 ### config
-- [ ] Add temporary name to `inventory/hosts`
+- [ ] Add temporary name to `inventory/hosts` (suggest doing this on separate branch)
```
mkllnk, line 22: I'm wondering about this _temporary_ host name. I don't think it's temporary. Don't we want to be able to identify a unique host like prod4.openfoodnetwork.org.au which is providing the site openfoodnetwork.org.au?

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -41,7 +44,7 @@ Then setup new server. Ensure you have the correct secrets (current secrets are
 - [ ] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per guide
```
mkllnk, line 44: Can you link to the guide?

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -9,6 +9,8 @@ assignees: ''
 
 Checklist based on general guide https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server
 
+Tip: find/replace to set up most commands ready to go, eg: `x_prod` -> `ca_prod`
```
mkllnk, line 12: I wonder if we should use easy to replace example names like `old_server` and `new_server` which could be prod or staging. Some of the guide just refers to `x` instead of `x_prod` as well. It would be good to make this consistent.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -60,44 +63,46 @@ Make sure to clear cache so that instance settings are applied:
...
+  `bin/rake db:reset` (You will need to confirm. Make sure you're on the new server!)
+- [ ] Update ansible_host IP in `inventory/hosts` and ensure provision works (this will update host in `.env.production`).
+    `ansible-playbook playbooks/provision.yml -l x_prod`
+- [ ] `ansible-playbook playbooks/deploy.yml -l x_prod -e "git_version=vX.Y.Z"` matching version with current prod
 - [ ] old server: make a tiny data change to verify later (eg add `.` in meta description `/admin/general_settings/edit`)
 
 ### switchover: old server
-- [ ] 🚧 `maintenance_mode.yml`
+- [ ] 🚧 `ansible-playbook playbooks/maintenance_mode.yml -l x_prod`
 - [ ] `sudo systemctl stop sidekiq redis-jobs puma`
```
mkllnk, line 75: Should we add `disable` as well? Just in case the server gets restarted accidentally. It means that rolling back is also one step more (`enable`) but it makes the intent of retiring the server at this stage clearer. And it could prevent us from accidentally running a start or restart command on the old server.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -60,44 +63,46 @@ Make sure to clear cache so that instance settings are applied:
...
+- [ ] `ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/db_transfer.yml &&`
+- [ ] `ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/transfer_assets.yml`
 - [ ] Transfer `/var/lib/redis-jobs/dump.rdb` to new server (see guide)
-- [ ] `db_transfer.yml` ~3min
-- [ ] `sudo systemctl stop postgres` (ensure other integrations no longer touch it)
-- [ ] `transfer_assets.yml` just in case
+- [ ] `sudo systemctl stop postgresql` (ensure other integrations no longer touch it)
```
mkllnk, line 79: Same here. I think that `disable` would be good as well.

.github/ISSUE_TEMPLATE/server-migration.md
```diff
@@ -60,44 +63,46 @@ Make sure to clear cache so that instance settings are applied:
...
 
 ### switchover: new server
 - [ ] `sudo systemctl restart puma; sudo systemctl start sidekiq redis-jobs`
```
mkllnk, line 82: Sidekiq is disabled at this point. We have to enable it first before we can start it. But I personally would probably do it staggered:

* Start puma and check with hosts file.
* Start sidekiq and check log file.
* Then install proxy forwarding.

mkllnk: How much time did you need for testing? If it's short, maybe we could do it during maintenance mode which would enable us to then switch over straight away. It would save us all the work of transferring data twice and resetting in between.

mkllnk (approved):
Nice improvements.

.github/ISSUE_TEMPLATE/server-migration.md
dacook: Agreed, I think the unique hostname is very helpful. I'm not sure, but we should be able to use that in the `hosts` definition, rather than the primary domain.

This is distinct from the app domain which is set in `host_vars` (`domain` and `certbot_domains`). The guide suggested also using the temporary domain there but I found it caused more problems than it was worth.

I feel like this list needs rewriting again but I won't do that while it's not fresh in my mind.

.github/ISSUE_TEMPLATE/server-migration.md
dacook:
```suggestion
- [ ] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per [guide](https://github.com/openfoodfoundation/ofn-install/wiki/Migrating-a-Production-Server)
```

.github/ISSUE_TEMPLATE/server-migration.md
dacook: That makes sense. Although in some places it's actually referring to the domain name (eg `host_vars/x/config.yml`) so that could have its own keyword.

.github/ISSUE_TEMPLATE/server-migration.md
dacook: I agree that would be much quicker, I just found that it was never straightforward enough to be confident that everything would work first go, so I preferred a more conservative approach.

.github/ISSUE_TEMPLATE/server-migration.md
dacook: Good idea. I just found that there's a convenient way to disable and stop a service in once command:

```suggestion
- [ ] `sudo systemctl disable --now sidekiq redis-jobs puma`
```

.github/ISSUE_TEMPLATE/server-migration.md
dacook:
```suggestion
- [ ] `sudo systemctl disable --now postgresql` (ensure other integrations no longer touch it)
```

.github/ISSUE_TEMPLATE/server-migration.md
dacook: > But I personally would probably do it staggered:

That's a good idea. I did those checks in the 'testing' stage, so skipped it here. But you could totally do it that way too.

.github/ISSUE_TEMPLATE/server-migration.md
mkllnk: Nice, TIL `disable --now`.

mkllnk (approved):

rioug:
@dacook can you have a look at the conflict ?

dacook:
Rebased and added bonus note.

rioug (approved):

Issue #966, opened 2024-11-21
[Upgrade Ubuntu] au_prod


dacook:
Clone existing image and upgrade it. Don't use temporary subdomain to save time. 
Switchover for minimal downtime.

## 1. Setting up the new server
- [x] clone current server in Rimu hosting
- [x] do-release-upgrade 18 -> 20

### config
- [ ] ~~Add temporary name to `inventory/hosts`~~
- [x] Review `host_vars/x/config.yml`, clean up if needed
- [x] Review `group_vars/x.yml`, clean up if needed
- [x] Review `ofn-secrets:x_prod/secrets.yml`, clean up if needed
- [x] enable Gmail IP authentication: Google Admin > Apps > Gmail > Routing > SMTP Relay Service
- [x] DNS: add temporary domain (eg `prod4.openfoodnetwork.org.au`)

### setup
Enable passthrough on _current_ server to allow new server to generate a certificate:
- [x] `ansible-playbook playbooks/letsencrypt_proxy.yml -l au_prod -e "proxy_target=prod4.openfoodnetwork.org.au"`

Then setup new server. Ensure you have the correct secrets (current secrets are usually fine).
`ansible-playbook -l x_prod2 -e "@../ofn-secrets/x_prod/secrets.yml" playbooks/`
- [x] `setup.yml`
- [x] `provision.yml`
- [x] `deploy.yml`
- [x] `db_integrations` (Permit DB access for n8n, Metabase)

### initial migration
- [x] Ensure sidekiq is disabled, to avoid creating subscription orders when data is loaded. Note that it will still get started by the deploy playbook each time it runs.
    `sudo systemctl stop sidekiq && sudo systemctl disable sidekiq`
- [x] Setup direct ssh access for `ofn-admin` and `openfoodnetwork` as per guide

`ansible-playbook -l x_prod -e rsync_to=x_prod2 playbooks/`
- [x] `db_transfer.yml` &&
- [x] `transfer_assets.yml`

Make sure to clear cache so that instance settings are applied:
`cd ~/apps/openfoodnetwork/current; bin/rails runner -e production "Rails.cache.clear"`

## 2. Testing
 - [ ] test `reboot`
 - [x] send test mail (`/admin/mail_methods/edit`). 
 - [x] terms of service file: `/admin/terms_of_service_files`
 - [x] shop catalogue display correctly, with images, add to cart, begin checkout, login
  - note: check cookies if login won't work
 - Check integrations 
[...]

PR #969, opened 2024-12-11
au-prod4

dacook:
Ready to go live tomorrow at 7am
- closes https://github.com/openfoodfoundation/ofn-install/issues/966

I'm skipping the temporary subdomain this time, it's too much hassle! I was able to test everything with a local hosts file entry.
I added a number to the hostname so that I could be extra sure which server I was working on.

@mkllnk is it ok to change the email address?

### DIFF
--- a/inventory/host_vars/openfoodnetwork.org.au/config.yml
+++ b/inventory/host_vars/openfoodnetwork.org.au/config.yml
@@ -4,7 +4,7 @@ domain: openfoodnetwork.org.au
 host_id: au-prod
 rails_env: production
 
-admin_email: maikel@openfoodnetwork.org.au
+admin_email: core-devs@openfoodnetwork.org.au
 
 mail_domain: openfoodnetwork.org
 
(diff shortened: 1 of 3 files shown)


inventory/host_vars/openfoodnetwork.org.au/config.yml
```diff
@@ -1,7 +1,7 @@
 ---
 
 domain: openfoodnetwork.org.au
-host_id: au-prod
+host_id: au-prod4
```
rioug, line 4: Why do we need to update this ?

rioug (commented):
I think it's all good, I am just not sure why we need to change the `host_id`, I believe it will affect New Relic, as in aus monitoring will now be under `au-prod4`

inventory/host_vars/openfoodnetwork.org.au/config.yml
dacook: I added the number to during switchover, which was really helpful to be sure which server was which.

It's still slightly relevant now, because we still have two vps in the hosting console:
Screenshot 2024-12-16 at 10 38 13 AM

But prod3 should be shutdown soon.. so yeah let's not have a superfluous number here. I will remove.

rioug (approved):

mkllnk (approved):
Nice. That's as simple as it should be.

mkllnk:
> is it ok to change the email address?

Yes, great.

Issue #922, opened 2024-03-26
Update New Relic redis monitoring configuration 


rioug:
## Description
This came up as part of investigating infinite loading bug : https://github.com/openfoodfoundation/openfoodnetwork/issues/11752#issuecomment-2017002049

New Relic provides monitoring for redis, however it's configured with the wrong port. We need to be monitoring redis on port 6380. Redis monitoring is accessible in New Relic via Infrastructure -> Third party services -> Redis

- production config : https://github.com/openfoodfoundation/openfoodnetwork/blob/6240f37adfc3993b1ce18decb2024a00968e77a7/config/environments/production.rb#L46
- staging config : https://github.com/openfoodfoundation/openfoodnetwork/blob/6240f37adfc3993b1ce18decb2024a00968e77a7/config/environments/staging.rb#L46

Server to fix :
- [ ] au_prod, we need to fix all third party monitoring see : #923
- [x] au_staging
- [ ] ca_prod 
- [x] hu_prod
- [ ] ie_prod
- ~~[ ] in_prod New Relic not setup~~
- [x] nz_prod 
- [x] uk_prod
- [ ] uk_staging
- [x] us_prod
- [x] es_prod
- [x] fr_prod
- [x] fr_staging
- [x] de_prod
- [x] be_prod

## TODO
Fix ansible script so the correct configuration is used next time we provision a server

## Fix 
As an ofn-admin, ssh to the server:
- `cd /etc/newrelic-infra/integrations.d`
- update the PORT entries to 6380 in `redis-config.yml`
- restart new relic service : `systemctl restart newrelic-infra`

## Expected Behavior
Relic monitoring reports redis activity :
![redis_good](https://github.com/openfoodfoundation/ofn-install/assets/40413322/8fc14f17-2738-4207-be15-3f5bfe42623e)

## Actual Behaviour
Relic monitoring shows a flat graph : 
![redis_bad](https://github.com/openfoodfoundation/ofn-install/assets/40413322/f8955545-99e8-4a95-8cea-7863ee33ae2b)

dacook:
I wondered if we can set this up with Ansible?

rioug:
All instance be `au_prod` (new relic not setup for third party monitoring ) and `in_prod` (new relic not setup) have been manually updated and are now reporting in new relic.

Leaving the ticket open till we fix the ansible script

rioug:
Server to fix :
- [ ] au_prod, we need to fix all third party monitoring see : #923
[...]

PR #970, opened 2024-12-11
Fix new relic redis configuration

rioug:
-closes https://github.com/openfoodfoundation/ofn-install/issues/922

~~By default New Relic monitor the redis instance running on port 6379. We are interested in monitoring the cache instance of redis, running on port 6380. This add a task to update the port to 6380.~~

~~Once merged, we'll need to update all the server (Will track that on #922 ) . You can use the following command :~~

By default New Relic monitor the redis instance running on port 6379. We are running three instances on port 6379,6380 and 6381. This change will add monitoring for all instances, and they get tagged with their role. 

Once merged, we'll need to update all the server (Will track that on #922 ) . You can use the following command :
```
ansible-playbook playbooks/provision.yml --limit=au_staging -e "@../ofn-secrets/au_staging/secrets.yml" --tags newrelic

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -119,6 +119,16 @@
       become_user: "{{ app_user }}"
       when: ansible_distribution_major_version >= '20'
 
+    - name: update new relic redis configuration # update redis port to monitor cache instance
+      replace:
+        dest: "/etc/newrelic-infra/integrations.d/redis-config.yml"
+        regexp: '^(\s*PORT:)\s6379$'
+        replace: '\1 6380'
+      become: yes
+      notify: restart newrelic
+      tags:
+       - newrelic_redis
+
     - meta: flush_handlers # Ensure handlers run successfully before reporting success
 
     - name: notify slack
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -119,16 +119,6 @@
       become_user: "{{ app_user }}"
       when: ansible_distribution_major_version >= '20'
 
-    - name: update new relic redis configuration # update redis port to monitor cache instance
-      replace:
-        dest: "/etc/newrelic-infra/integrations.d/redis-config.yml"
-        regexp: '^(\s*PORT:)\s6379$'
-        replace: '\1 6380'
-      become: yes
-      notify: restart newrelic
-      tags:
-       - newrelic_redis
-
     - meta: flush_handlers # Ensure handlers run successfully before reporting success
 
     - name: notify slack
--- a/roles/newrelic/tasks/main.yml
+++ b/roles/newrelic/tasks/main.yml
@@ -27,3 +27,95 @@
     regex: "^display_name:"
     state: present
   notify: restart newrelic_infra
+
+- name: Update new relic redis configuration # update redis port to monitor cache instance
+  copy:
+    content: |
+      integrations:
+        - name: nri-redis
+          env:
+            METRICS: true
+            HOSTNAME: localhost
+            PORT: 6379
+            KEYS: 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: playbooks/provision.yml [hunk 1]
```diff
@@ -119,6 +119,16 @@
       become_user: "{{ app_user }}"
       when: ansible_distribution_major_version >= '20'
 
+    - name: update new relic redis configuration # update redis port to monitor cache instance
+      replace:
+        dest: "/etc/newrelic-infra/integrations.d/redis-config.yml"
+        regexp: '^(\s*PORT:)\s6379$'
+        replace: '\1 6380'
```
dacook, line 126: This should work, but it assumes although it's worth noting that if the port was set to anything else, it would silently skip over and do nothing. I'm not sure how to improve it though.

dacook (commented):
Nice. Is it possible to monitor multiple instances? We also use it for sidekiq (https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/group_vars/all.yml#L75-L76), so it would be good to monitor that also.

dacook (changes requested):
Unfortunately it doesn't work in CI:

https://github.com/openfoodfoundation/ofn-install/actions/runs/12270155766/job/34234930102
```
TASK [update new relic redis configuration] ************************************
fatal: [local_test]: FAILED! => {
    "changed": false,
    "rc": 257
}

MSG:

Path /etc/newrelic-infra/integrations.d/redis-config.yml does not exist !
```

roles/newrelic/tasks/main.yml
```diff
@@ -27,3 +27,95 @@
     regex: "^display_name:"
     state: present
   notify: restart newrelic_infra
+
```
dacook, line 30: It took a bit of reading to work out what this all does before I could summarise in my head. 
Now that I have, I figured it worth adding for future reference.

```suggestion

# Monitor all three Redis instances:
#  - Metrics every 15s
#  - Inventory every 60s
```

dacook: I [looked up](https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/redis/redis-integration/#metrics) what Metrics and Inventory are, and wondered if we need it to be so frequent.

I thought once a minute for metrics would be fine, and inventory info would change very infrequently, so I'd suggest hourly or daily). What do you think?

dacook: Oh and one more thought: if we're replacing the contents of a file with this content, then it would be more manageable to store the content in a separate template file. Here's an example where we do that: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app_user/tasks/main.yml#L38-L44 (the file is in a folder called templates). I'm pretty sure it can be a plain .yml file.

dacook (changes requested):
Thanks for following up. I have a few suggestions, what do you think?

roles/newrelic/tasks/main.yml
rioug: > I thought once a minute for metrics would be fine, and inventory info would change very infrequently, so I'd suggest hourly or daily). What do you think?

Sounds good, I had just reused the config value provided by new relic.

roles/newrelic/templates/redis-config.j2
```diff
@@ -0,0 +1,87 @@
+{{ template_notice }}
```
dacook, line 1: Good idea to include this!! 🏅

dacook (approved):
Awesome 😎

mkllnk (approved):

Issue #968, opened 2024-12-08
Error when updating nodenv


dacook:
Playbook: `provision.yml`
OS: Ubuntu 20.04

When provision has been executed a second time, nodenv update fails with below message.
I've seen it on a few servers, including [uk_prod](https://github.com/openfoodfoundation/ofn-install/issues/946#issuecomment-2487038388).

```
TASK [node : fetch nodenv] *********************************************************************************************
ok: [staging.openfoodnetwork.org.au]
Monday 09 December 2024  10:33:25 +1100 (0:00:04.843)       0:07:27.864 *******

TASK [node : install nodenv plugins] ***********************************************************************************
failed: [staging.openfoodnetwork.org.au] (item={'name': 'node-build', 'repo': 'https://github.com/nodenv/node-build.git', 'version': 'master'}) => {
    "ansible_loop_var": "item",
    "changed": false,
    "cmd": [
        "/usr/bin/git",
        "fetch",
        "--tags",
        "origin"
    ],
    "item": {
        "name": "node-build",
        "repo": "https://github.com/nodenv/node-build.git",
        "version": "master"
    }
}

MSG:

Failed to download remote objects and refs:  From https://github.com/nodenv/node-build
   f6fab6d6..a502fc42  master     -> origin/master
   f6fab6d6..a502fc42  main       -> origin/main
 ! [rejected]          v5         -> v5  (would clobber existing tag)
 * [new tag]           v5.3.20    -> v5.3.20
```

## Workaround
Quick fix is to ssh and manually reset it, ie:

```
openfoodnetwork@uk-prod2:~/$ cd ~/.nodenv/plugins/node-build
openfoodnetwork@uk-prod2:~/.nodenv/plugins/node-build$ git fetch --tags -f
```

mkllnk:
I'm assuming that this is fixed now. Re-open if you come across this again.

dacook:
So I finally bothered to investigate, and it looks like they are maintaining a general `v5` tag, updating it each time there's a new release. Which creates an error when performing a standard git operation, bad idea!

![Screenshot 2025-01-10 at 11 56 09 am](https://github.com/user-attachments/assets/722b2383-61e6-4ade-9d93-205c738346bc)

I don't get why [this is not a problem for everyone else](https://github.com/nodenv/node-build/issues?q=is%3Aissue+is%3Aopen+v5)

PR #975, opened 2025-01-03
Update node-build branch from master to main

mkllnk:
The HEAD branch was renamed. Even after updating the name in Ansible, it still failed on not wanting to override a tag. The workaround is to delete the plugin remotely:

```
rm ~/.nodenv/plugins/node-build/ -rf
```

I don't think it's worth integrating into Ansible. It would be better to switch to a bash script installing node.

### DIFF
--- a/roles/node/defaults/main.yaml
+++ b/roles/node/defaults/main.yaml
@@ -3,7 +3,7 @@ nodenv_plugins:
   # Allow installation of node versions
   - name: "node-build"
     repo: "https://github.com/nodenv/node-build.git"
-    version: master
+    version: main
   # Automatically install yarn after installing node
   - name: "nodenv-yarn-install"
     repo: "https://github.com/pine/nodenv-yarn-install"


dacook (approved):
Thanks 👍 

So I'm guessing this is to
* fix https://github.com/openfoodfoundation/ofn-install/issues/968 ?

And that we would also need to manually delete/reset the plugin on each server after it is merged?

rioug:
I am a bit confused, are you saying this change doesn't fix the problem ? is it actually worth doing it ? I guess it makes sense for consistency, but if we are going to stop using it why not leaving it as it is till we move to a bash script ?

mkllnk:
> are you saying this change doesn't fix the problem ?

It does fix the problem but it also needs a once-off manual fix to sort out git.

mkllnk:
> we would also need to manually delete/reset the plugin on each server after it is merged?

Yes, I think so.

rioug (approved):

mkllnk:
I removed the outdated repositories:

```
ansible all_staging -u openfoodnetwork -a "rm ~/.nodenv/plugins/node-build/ -rf"
ansible all_prod -u openfoodnetwork -a "rm ~/.nodenv/plugins/node-build/ -rf"
```

Provisioning should restore that again. And we don't need it unless our node version changes.

dacook:
Great, I love it when ansible helps solve little problems like this!

PR #977, opened 2025-01-07
Remove db access for the old metabase server

rioug:
Now that data.openfoodnetwork.org.uk is retired, remove database access. 

I also clean up `au_prod` to be in line with the current production configuration.


dacook (approved):
Great, so the au_prod config in metabase has been updated to use the right user?

rioug:
> Great, so the au_prod config in metabase has been updated to use the right user?

Yes it has, thanks for checking.

PR #979, opened 2025-01-07
Skip installing ruby, the app does it

mkllnk:
Now we don't have to manage the ruby version in two places. The bash script in the app is also faster than Ansible. Follow up from:

* https://github.com/openfoodfoundation/openfoodnetwork/pull/13034


dacook (approved):
🏎️

rioug (approved):
:broom: :+1:

PR #982, opened 2025-01-08
Adjust Vagrant IP address to allowed Debian range

mkllnk:
I got this error before:

```
The IP address configured for the host-only network is not within the
allowed ranges. Please update the address used to be within the allowed
ranges and run the command again.

  Address: 192.168.50.4
  Ranges: 192.168.56.0/21

Valid ranges can be modified in the /etc/vbox/networks.conf file. For
more information including valid format see:

  https://www.virtualbox.org/manual/ch06.html#network_hostonly
```

Any objections?


dacook (approved):
Sounds good 👍

rioug (approved):
:+1:

PR #985, opened 2025-01-10
Re-introduce python installation

mkllnk:
This reverts commit a872e875d4367adba0bb9dc627a10db303f1d67c.

I misread the conditional chain and thought that the original code was only installing python for old Ubuntu versions. But it was just choosing the right python version depending on the Ubuntu version. We can now assume Python 3 and simplify the check.

### DIFF
--- a/playbooks/setup.yml
+++ b/playbooks/setup.yml
@@ -5,6 +5,23 @@
 # Use: `ansible-playbook playbooks/setup.yml -e "ansible_user=ubuntu" --limit <server_tag>` if
 # provisioning an EC2 instance for the first time (this only needs to be run once).
 
+# Ensure python is set up and accessible on the server before starting.
+- name: python check
+  hosts: ofn_servers
+  gather_facts: no
+  remote_user: root
+  tasks:
+    - name: Install python
+      become: yes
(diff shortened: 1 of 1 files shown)


playbooks/setup.yml
```diff
@@ -5,6 +5,23 @@
...
+# Ensure python is set up and accessible on the server before starting.
+- name: python check
+  hosts: ofn_servers
+  gather_facts: no
+  remote_user: root
+  tasks:
+    - name: Install python
+      become: yes
+      raw: |
+        test -e /usr/bin/python || (
```
rioug, line 17: Should we also test for `/usr/bin/python3` ? we could skip the install if already there ?

mkllnk: Currently, we depend on that symlink:

https://github.com/openfoodfoundation/ofn-install/blob/e1c518166463c1199ceb6c53b544783fd0bde0ac/inventory/group_vars/all.yml#L410-L411

But we could experiment with removing that config. I would expect Ansible to be clever enough to find any python on the machine.

rioug (approved):

dacook (approved):

PR #988, opened 2025-01-14
Enable default redis instance

rioug:
While looking at setting monitoring for all redis instances, I noticed a some of our prod instances don't have the default redis instances. That means we can't update the ActionCable configuration to use said instance as proposed here https://github.com/openfoodfoundation/openfoodnetwork/pull/13072

According to the [documentation](https://redis.io/docs/latest/operate/oss_and_stack/install/install-redis/install-redis-on-linux/) redis should be enable/start automatically but it appears to not always be the case.
This PR add a new role to make sure the default instance of redis is enable and running when provisioning a server.

### DIFF
--- a/playbooks/provision.yml
+++ b/playbooks/provision.yml
@@ -66,6 +66,9 @@
       become_user: root
       tags: redis
 
+    - role: enable_redis
+      tags: enable_redis
+
     - role: sidekiq
       tags: sidekiq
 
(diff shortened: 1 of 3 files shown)


dacook (approved):
Great, although it seems like something that should be handled by the galaxy role.

It might be easy enough to contribute: https://github.com/libre-ops/multi_redis/blob/main/tasks/main.yml

But for now, this does the job 👍

mkllnk (approved):

PR #989, opened 2025-01-15
Run tests on CI with LXC container 

mkllnk:
We were running provisioning and deploy through a local connection on the Github runner itself. But when we upgrade to Ubuntu 22 and higher, that fails because of the permissions within the Github Actions container. So I decided to use a lightweight container which I've been using in development as well.

Now we are independent of the Github images and we could even run the tests against multiple Ubuntu versions in parallel.

The LXC setup took a little bit longer but not much:

* Old local setup: 17m
* LXC setup: 19m

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,41 @@ jobs:
       - name: Ansible Lint
         run: bin/lint
 
+  playbook-tests-lxc:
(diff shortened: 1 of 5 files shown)


rioug (approved):
Nice one !

dacook (approved):
Great!

Issue #157, opened 2018-04-28
Upgrade Ubuntu


mkllnk:
New servers should have a supported OS for as long as possible. 
~~The just released Ubuntu 18.04 will probably be supported till 2023.~~ 
*~~Update: It's 2023 now and we should go straight to Ubuntu 22.~~*
*Update: Our Ansible stack doesn't support past Ubuntu 20. So for now, we will upgrade the oldest servers from v16 to 20.*

- [x] #938
- [x] #940
- [x] #946
- [x] #957
- [x] #958
- [x] #959
- [x] #939
- [x] #966
- [x] #965

elf-pavlik:
I will try to run current playbook in container running bionic, i think it may just work.

enricostano:
We need to check the priority of this and the current situation of all the instances that are gradually adopting `ofn-install`. I propose to do that before we dedicate any effort to this.

myriamboure:
Ping @pacodelaluna

sauloperez:
In Katuma we will soon migrate our old and small production server and will do it with 18.04. I'll keep you posted.

sigmundpetersen:
AFAIK every managed instance is running on Ubuntu 18, closing

RachL:
@sigmundpetersen UK and FR at least are on 16.

sigmundpetersen:
Yes I was confusing 18 compatibility with actual implementation on the servers. Sorry

Matt-Yorkley:
Current status:
```
openfoodnetwork.be:	16.04
openfoodnetwork.org.uk:	16.04
openfoodnetwork.ie:	16.04
openfoodnetwork.de:	16.04
coopcircuits.fr:	16.04
openfoodnetwork.ca:	16.04
-----------------------------
app.katuma.org:         18.04
openfoodnetwork.net:	18.04
openfoodnetwork.org.au:	18.04
```

mkllnk:
Servers still running on Ubuntu 16:

- staging.openfoodnetwork.org.au
- staging.openfoodnetwork.org.uk
- openfoodnetwork.ca
- openfoodnetwork.org.uk
- openfoodnetwork.ie
- openfoodnetwork.de
- coopcircuits.fr

mkllnk:
### Estimate

I actually did the last upgrade of the French server. I logged 6h40m for that task.

* #825

We probably need a similar time for the first server upgrade because there are always little issues coming up. But then we are in a position to just execute scripts for each server. 

[...]

PR #990, opened 2025-01-16
Add support for Ubuntu 22 and Ubuntu 24

mkllnk:
Relates to:

* #157

Includes and depends on:

* #989

The brotli compilation failed on newer Ubuntu versions. I chose the easy path for now to deactivate brotli for newer Ubuntu since it's only relevant for a few servers and only when we actually replace servers again.

From Ubuntu 24 onwards, brotli can be installed via apt. That may require replacing the used nginx role which isn't maintained any more anyway. That's a different task though.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,7 +15,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # options are documented and commented below. For a complete reference,
   # please see the online documentation at vagrantup.com.
 
-  config.vm.box = "generic/ubuntu2004"
+  config.vm.box = "generic/ubuntu2204"
 
   # VM network config.
   config.vm.network "forwarded_port", guest: 22, host: 2222
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,7 +15,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # options are documented and commented below. For a complete reference,
   # please see the online documentation at vagrantup.com.
 
-  config.vm.box = "generic/ubuntu2204"
+  # Not available in the generic repository yet.
+  #config.vm.box = "generic/ubuntu2404"
+  config.vm.box = "alvistack/ubuntu-24.04"
 
   # VM network config.
   config.vm.network "forwarded_port", guest: 22, host: 2222
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,41 @@ jobs:
       - name: Ansible Lint
         run: bin/lint
 
+  playbook-tests-lxc:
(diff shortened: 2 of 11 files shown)


rioug (approved):
Good one :+1: 
I am wondering what impact turning off `brotli` would have, do we have any way track that ? I am not sure if New Relic would be useful here.

mkllnk:
> I am wondering what impact turning off brotli would have, do we have any way track that ?

We are not tracking that, no. But the brotli filter is currently only applied to JS and CSS assets and JSON responses. The biggest instances use S3 as CDN for assets. Canada or US may be affected by asset delivery. All servers are affected by compression of JSON responses. Small payloads don't benefit from brotli though. So the biggest impact would be on big APIv0 responses. But popular responses like loading products for a shop front are paginated and loaded lazily. Each chunk is small.

With the perspective of us moving away from APIv0 and more towards rendering server-side, trying to operate in small quick actions instead of big bulk actions, the benefit of botli may not be there.

Anyway, I was thinking that we would get brotli back at some point. Since we can easily install it via Apt, there won't be much needed to support it again. And currently we have all servers running Ubuntu 20 with brotli for a while. It would only affect new instances which probably don't have big payloads anyway (except assets and we should encourage a CDN then).

Vagrantfile
```diff
@@ -15,7 +15,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # options are documented and commented below. For a complete reference,
   # please see the online documentation at vagrantup.com.
 
-  config.vm.box = "generic/ubuntu2004"
+  # Not available in the generic repository yet.
+  #config.vm.box = "generic/ubuntu2404"
+  config.vm.box = "alvistack/ubuntu-24.04"
```
dacook, line 20: Hrm, seems like there might not be any official boxes anymore: https://askubuntu.com/a/1521304/164829

Another reason to use LXC I suppose! (for Linux users).

dacook (approved):

dacook:
Fantastic work, congrats and well done on getting us there!

Vagrantfile
mkllnk: > Another reason to use LXC I suppose! (for Linux users).

So what will you do on Mac?

I tend to ignore Windows because they can run Linux in any virtual environment and go from there. But Mac is a unix and popular amongst a lot of devs. I wouldn't mind you switching to Linux though. :wink:

dacook: I did actually consider staying on Mac so that we can maintain support for all environments that contributors may have (yes, even the proprietary closed Apple ecosystem). I

I started looking yesterday and came across [Macpine](https://beringresearch.github.io/macpine/) which might be a good option, using Macpine

PR #1001, opened 2025-05-28
🇫🇮 Add Finland (FI) server configuration

jarihu:
This PR adds configuration files for the Finnish Open Food Network instance (openfoodnetwork.fi) to the ofn-install project. These changes enable automated setup and maintenance for the Finland server, consistent with existing OFN instances.

Changes:
Added inventory hosts/host_vars/group_vars with specific configuration for the Finnish instance and SSH pubkey.


inventory/group_vars/fi.yml
```diff
@@ -0,0 +1,21 @@
...
+l10n_repo: https://github.com/jarihu/l10n_fi.git
+timezone: Helsinki
+
+
+# Used for the LetsEncrypt email setting
+developer_email: hello@openfoodnetwork.fi
+
+# Users listed here will have their key from /files/keys added to the app_user
+users_sysadmin:
+  - jari-fi
```
dacook, line 23: Note that all other instances on in this repo are configured to allow the core dev team access to the server, in order to provide support, and deploy weekly releases.
```yaml
      - "{{ core_devs }}"
```

But we can also only provide this where an agreement has been made, including a funding commitment. I'm not sure if you've seen this yet, but the agreement is described here: https://docs.google.com/spreadsheets/d/1uHc3Pn-XSi5vJ95YIvWRtRJCbz3nXL5apdBWipmHUYc/edit?gid=2012212782#gid=2012212782

Of course, this is your choice, and you can choose later. I can see that you are quite capable of supporting the server yourself already.

dacook (approved):
Great job 👍 
Just awaiting a second review.

mkllnk (approved):

inventory/hosts
```diff
@@ -207,6 +207,7 @@ uk_prod
 de_prod
 ie_prod
 hu_prod
+fi_prod
```
mkllnk, line 210: Sorry, I'm just noticing that the `europe` group is used in our release deployments. But we don't have access to your server. So this will fail when we try to deploy. I'll probably remove this entry then.

PR #1004, opened 2025-06-10
Run CI on ubuntu-latest and set up containers with Incus

### LINKED ISSUE #1000: Upgrade Ubuntu version on CI
Ubuntu 20 is no longer supported and we need to run a newer version for our tests.

mkllnk:
Current Ubuntu 20 is no longer supported and we need to run a newer version for our tests.
Incus works better for us than the new LXD maintained by Canonical.

* Fixes #1000

### DIFF
--- a/bin/setup-incus
+++ b/bin/setup-incus
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+# Set up Linux Containers with Incus
+#
+# If you don't run Linux, you better look at Vagrant to set up virtual machines.
+#
+# But if you run Linux anyway then you can use containers to run virtual
+# machines on the same kernel instead of creating a full virtual machine.
+#
+# Installation in Debian:
+#
+#   sudo apt install incus
+#   sudo incus init
+#   sudo usermod -aG incus "$USER"
+#   # Restart your user's session to gain the new group privilege.
+
+set -e
+
+: ${LXC_IMAGE='images:ubuntu/24.04'}
+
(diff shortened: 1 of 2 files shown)


bin/setup-incus
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+
+: ${LXC_IMAGE='images:ubuntu/24.04'}
+
+echo "Launching $LXC_IMAGE container:"
+incus launch "$LXC_IMAGE" ofn-dev
+incus exec ofn-dev -- apt-get update
+incus exec ofn-dev -- apt purge openssh-client --yes
+incus exec ofn-dev -- apt install openssh-server --yes
```
dacook, line 25: I thought it was weird that ubuntu 24 wouldn't have an ssh server ready to go, but figured you worked everything out and it's not worth questioning if it works.

But I just randomly saw this topic pop up on StackExchange which seems related, so I thought it worth sharing in case it illuminates things further: https://askubuntu.com/questions/1550366/ssh-sshd-breaking-changes-in-24-04

mkllnk: Ah, that's interesting. I didn't know that and I didn't try to run ssh.socket instead of sshd.service. That would require two different scripts though, one for Ubuntu 20 and one for Ubuntu 24. So maybe let's look into that when we ditch Ubuntu 20.

rioug (approved):

PR #1011, opened 2025-08-01
Install libyaml-dev to build psych gem for Rails 7.1

mkllnk:
* Unblocks https://github.com/openfoodfoundation/openfoodnetwork/pull/13232

All servers have been updated already.


mkllnk:
It looks like we  have a problem with our test setup again. :-( 

I applied this to all servers already, so the failing tests are not indicating anything bad.

PR #1018, opened 2025-10-16
Update to Ansible 12

### LINKED ISSUE #1017: Update to Ansible 12
The idea is to use the latest version of Ansible.
I tried it locally for FR_staging and it was working with few tweaks, so I will share them and let's decide what to do next.

pacodelaluna:
Closes #1017.

The objective is fix the breaking changes related to Ansible 12.

### DIFF
--- a/roles/nodenv_install/tasks/main.yaml
+++ b/roles/nodenv_install/tasks/main.yaml
@@ -7,4 +7,4 @@
 - name: node installation messages
   debug:
     msg: "{{nodenv_install.stderr}}"
-  when: nodenv_install.stderr
+  when: nodenv_install.stderr | length > 0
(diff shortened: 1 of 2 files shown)


roles/nodenv_install/tasks/main.yaml
```diff
@@ -7,4 +7,4 @@
 - name: node installation messages
   debug:
     msg: "{{nodenv_install.stderr}}"
-  when: nodenv_install.stderr
+  when: nodenv_install.stderr | length > 0
```
pacodelaluna: Related to the [breaking changes on Ansible 12](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_12.html#broken-conditionals). Need to check if we have other cases in `ofn-install`.

PR #1019, opened 2025-10-17
Add role jdauphant.nginx, not avaible in galaxy

mkllnk:
The repository has been archived:

* https://github.com/jdauphant/ansible-role-nginx

We should replace this role with our own version though. Storing large configuration files in variables isn't easy to work with. And it probably does a lot more than we need it to do.


rioug (approved):
:ok_hand:  should we open a ticket to track replacing this  ?

mkllnk:
> should we open a ticket to track replacing this ?

We can. Not sure it solves a problem at the moment. I would probably do it when we have to change large amounts of nginx config. :shrug:

dacook (approved):
It still appears available in galaxy: https://galaxy.ansible.com/ui/standalone/roles/jdauphant/nginx/documentation/

But as it's archived, good idea to store our own copy 👍

PR #1022, opened 2025-11-13
Use the same version of Python everywhere, and let it be specified in a single place

deivid-rodriguez:
By keeping a more standard `.python-version` file, `actions/setup-python` will respect it and does not need a specific input with the version of Python. That should ensure the same version is used locally and in CI.

### DIFF
--- a/bin/pyenv-install
+++ b/bin/pyenv-install
@@ -7,9 +7,5 @@ python_version="$(cat .python-version)"
 # Install python version
 pyenv install --skip-existing
 
-# Create python virtual environment
-if pyenv virtualenvs 2>&1 | grep -q "not installed"; then
-  pyenv virtualenv "$python_version" "ofn-install-$python_version"
-fi
 pip install -r requirements.txt
 
--- a/bin/virtualise-pyenv
+++ b/bin/virtualise-pyenv
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Test different Ansible version in virtual python environments.
+# Create and activate a new env with:
+#
+#   source bin/virtualise-pyenv
+#
+# The env is only active for you current shell.
+# Closing your shell will revert back to the default python environment.
+
+set -e
+
+python_version="$(cat .python-version)"
+ansible_version="$(grep "^ansible==" requirements.txt)"
+venv="$python_version-$ansible_version"
+
+# Install python version
+pyenv install --skip-existing
+
+# Create python virtual environment
+if ! pyenv shell "$venv"; then
+  pyenv virtualenv "$venv"
+  pyenv shell "$venv"
+fi
+
+pip install -r requirements.txt


bin/pyenv-install
```diff
@@ -2,15 +2,14 @@
...
 
 # Install python version
-pyenv install "$python_version" --skip-existing
+pyenv install --skip-existing
 
 # Create python virtual environment
 if pyenv virtualenvs 2>&1 | grep -q "not installed"; then
-  pyenv virtualenv "$python_version" "$env_version"
+  pyenv virtualenv "$python_version" "ofn-install-$python_version"
 fi
```
mkllnk: Hm, this is now broken. The virtualenv plugin was creating a "new python version" based on another version. It used the `.python-version` file to track which python version is currently used. So since that's just the original version now, the if statement is always false and we never create a new virtualenv. But even if we created a new environment, we would need to use it.

The suffix in the version was indicating the Ansible version. That allowed us to switch between Ansible versions by choosing a different virtualenv. Now I can still set `pyenv shell ofn-install-3.10.16` to use that version without overriding the .python-version file. But the pyenv-install script isn't really supporting me in this.

Maybe we could write another script that looks at the Ansible version in requirements.txt and set the right virtualenv. You could use it like this:

```
source bin/virtualise-pyenv

cat bin/virtualise-pyenv
python_version="$(cat .python-version)"
ansible_version="$(grep "^ansible==" requirements.txt)"
venv="$python_version-$ansible_version"
if !pyenv shell "$venv"; then
  pyenv virtualenv "$venv"
  pyenv shell "$venv"
fi
```

That's only a temporary state but that's okay, I guess. I don't know. Maybe something to try when we upgrade Ansible next.

I think, for now, you can just delete that section because it's not doing anything. And then the virtualenv thing is optional for devs that want to use it.

mkllnk (changes requested):

bin/pyenv-install
mkllnk: Yes, correct, but your PR still has some value. The virtualenv setup was added by one developer to switch environments more easily but it's not essential.

Given that virtualenv is messing with the .python-version file and we need it only for development, I would go with your PR and manage virtualenv manually in dev.

Hm, it does mean that we can't upgrade Ansible without updating the Python version as well or wipe Ansible and re-install it. Anyway, I don't want to update the python version and many places either, so let's go ahead with your PR, I think.

@rioug and @dacook, what do you think?

rioug: I am not really across virtualenv to be honest, but I am happy to leave that to devs to figure out when they need it. 
Do we need to update the README if we go ahead ?

dacook (approved):

bin/pyenv-install
dacook: Oh, i see that this might still need some work. Of course, it should be tested in a dev environment. I'm happy to give that a go once it's ready.

dacook: Hm I'm not sure, but I think what we want is dependency management for Python, just like we have Bundler for Ruby. Ansible is a dependency and we would like to lock the version in the same way. 

I guess we were using a script with virtualenv to achieve that, but maybe there is a better more standard tool that allows us to keep `.python-version` clean. A quick search suggests `pipenv`, or maybe [`uv`](https://github.com/astral-sh/uv). 

Or, now we're on a newish version of Ansible, maybe we could just try to keep up to date and not worry about locking it?

mkllnk (approved):
I just added a commit to accept your change of not using virtualenv by default but have a new script to use it optionally. That means that our development is as flexible as before and we stay compatible with all the standard tools. We are now also open to other managers.

deivid-rodriguez:
Thanks @mkllnk!

PR #1025, opened 2025-11-17
Add log rotate for redis-cache and redis-jobs

rioug:
While looking at the NZ server I realised that log for the non default instance of redis were not rotated. 
This add log rotation with the same config as the one for the default instance of redis.

Deploy command : 
```
ansible-playbook playbooks/provision.yml --tag logrotate --limit all_prod
```


dacook (approved):
Great!

PR #1028, opened 2025-11-30
Give access to Belgium server to Paco

pacodelaluna:
Coopcircuits team is supposed to manage now the Belgium instance at business side, so I was thinking to get an access to the server the same way I got to Coopcircuits one.


rioug:
@pacodelaluna I just deployed the change, you should be good to go

Issue #1029, opened 2025-12-01
Update allowed Zappier ip range


rioug:
## Description
From Karla at OFN Canada :

 

> Hello there, We are using Zapier for some our OFN Canada integrations, and it looks like Zapier is making some updates that require we update our firewall or network configuration for the MySQL, PostgreSQL, and SQL Server to allow the following IP ranges:
>  44.214.195.64/28
>  18.246.81.208/28

The updates are needed by Feb. 15 2026

PR #1030, opened 2025-12-05
Add new zapier ip to postgres hba config

rioug:
Closes: #1029 

Update allowed zapier addresses, they are due by Feb. 15 2026


rioug:
@mkllnk did you deploy the change ?

Issue #1003, opened 2025-05-28
Ubuntu 24.04 requires updated Ansible due to Python 3.12


jarihu:
If target host is running Ubuntu 24.04, Ansible tasks like apt_repository fail with older Ansible versions due to Python 3.12 removing deprecated SSL args. Python 3.12 is installed on Ubuntu 24.04 by default. 
Following error is shown when running playbooks:
`Failed to fetch PPA information, error was: An unknown error occurred: HTTPSConnection.__init__() got an unexpected keyword argument 'cert_file'
`

Cause:
Ansible-core versions older than 2.16.0 are incompatible with Python 3.12’s updated ssl module.
In Python 3.12, the parameters cert_file, key_file, and check_hostname were fully removed from ssl.create_default_context() and related HTTPS connection code. These were deprecated for years, but now their removal breaks tools or modules (including add-apt-repository, software-properties, or Ansible modules) that internally depended on Python’s old HTTPSConnection behavior.

Solution:
```
sudo apt purge ansible
sudo apt install python3.10-venv
source ~/ansible-venv/bin/activate
pip install ansible
```

Add ~/ansible-venv/bin to your PATH in .bashrc or .zshrc:
`export PATH="$HOME/ansible-venv/bin:$PATH"`

Verify:
`ansible --version`
ansible [core 2.17.12]

mkllnk:
@jarihu was this solved by your PR?

* #1002

mkllnk:
> Ansible-core versions older than 2.16.0 are incompatible with Python 3.12’s updated ssl module.

We updated from Ansible 2.10 to 3 earlier this year and are now on Ansible 8.5.0. Happy to update further. We use pyenv.

mkllnk:
Ah, I'm seeing now that you are referring to the ansible-core version. 2.16 which is in community Ansible version 9. I'll try to update.

PR #1034, opened 2026-01-07
Bump to Python 3.11 and Ansible 9.13 (core 2.16.15)

mkllnk:
* Closes https://github.com/openfoodfoundation/ofn-install/issues/1003

We could update further.

### DIFF
--- a/bin/pyenv-install
+++ b/bin/pyenv-install
@@ -5,12 +5,19 @@ set -e
 python_version="$(cat .python-version)"
 
 # Install python version
+# On failure, update pyenv to find new version and try again
 # List available versions:
 #   pyenv install --list | grep '^  3\.1.\.[0-9]*$'
 if ! pyenv install --skip-existing; then
-  (cd "$(pyenv root)" && git pull)
+  # If installed with homebrew, use that to update
+  if (command -v brew && brew list --versions pyenv); then
+    brew update && brew upgrade pyenv
+  else
+    # Update with git
+    (cd "$(pyenv root)" && git pull)
+  fi
+
   pyenv install --skip-existing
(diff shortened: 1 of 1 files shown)


bin/pyenv-install
```diff
@@ -5,7 +5,12 @@ set -e
 python_version="$(cat .python-version)"
 
 # Install python version
-pyenv install --skip-existing
+# List available versions:
+#   pyenv install --list | grep '^  3\.1.\.[0-9]*$'
+if ! pyenv install --skip-existing; then
+  (cd "$(pyenv root)/plugins/python-build/../.." && git pull)
```
dacook: This `cd` command doesn't work on my computer. If the `/../..` means up two levels, then it would simply land at pyenv root.

dacook (changes requested):
The `cd` command doesn't work for me, is it a mistake?

bin/pyenv-install
mkllnk: I'm sorry. I'm not sure what happened there. It should have been just the pyenv root. That's what pyenv tells me:

```
$ pyenv install 3.11.15
python-build: definition not found: 3.11.15

See all available versions with `pyenv install --list'.

If the version you need is missing, try upgrading pyenv:

  cd /home/maikel/.pyenv && git pull && cd -
```

While `plugins/python-build/../..` is an unnecessary indirection, it should have worked still. Why does it fail? Does the simpler version work?

```
(cd "$(pyenv root)" && git pull)
```

dacook (approved):

mkllnk:
Tests pass now that the database seeding is fixed. Merging.

PR #1039, opened 2026-01-20
Move bash_profile content to a managed block

dacook:
Before, this role was replacing the whole file. The node role also adds a managed block to this file, which was getting overwritten.

Now, the roles can function independently of each other.

Context: https://openfoodnetwork.slack.com/archives/GK2T38QPJ/p1768866468549609

This includes an extra task necessary for migration, which can be removed once it has been executed on all servers.

## After merge
This needs to be deployed to all managed staging and production servers. Some staging servers have been done already. Eg the bare minimum:
```
ansible-playbook playbooks/provision.yml -l fr_staging -t app_user,node
ansible-playbook playbooks/provision.yml -l all_prod -t app_user
```

## Follow-up 
After it has been provisioned, we can remove the temp code.

I realised we dont' have a process for announcing these changes to the wider community that may or may not be using these playbooks.
If other people _are_ using these scripts and miss out on the migration step, all that will happen is there will be few lines of duplicated code, which I think is no big deal.

### DIFF
--- a/roles/app_user/tasks/main.yml
+++ b/roles/app_user/tasks/main.yml
@@ -49,7 +49,13 @@
     group: "{{ app_user }}"
     mode: 0640
 
-# Temporary task to migrate to new managed block (see next)
+# Temporary tasks to migrate to new managed block (see next)
+- name: ensure file exist
+  become: true
+  become_user: "{{ app_user }}"
+  file:
+    path: "~/.bash_profile"
+    state: touch
 - name: Remove known un-managed bash_profile content
   become: true
   become_user: "{{ app_user }}"


dacook:
I'll go ahead with the provision

dacook:
## Provisioned
To fr_staging and all_prod

FYI @mkllnk  I noticed that at least one server didn't have the "template_message" at the top of the file, so the cleanup task didn't work, and there's some duplication 😞 

```bash
openfoodnetwork@ca-prod2:~$ cat ~/.bash_profile
set -o allexport
. /etc/default/openfoodnetwork
set +o allexport
# BEGIN ANSIBLE MANAGED BLOCK: nodenv
export PATH="$HOME/.nodenv/bin:$PATH"
eval "$(nodenv init -)"
# END ANSIBLE MANAGED BLOCK: nodenv
# BEGIN ANSIBLE MANAGED BLOCK: app_user
set -o allexport
. /etc/default/openfoodnetwork
set +o allexport
# END ANSIBLE MANAGED BLOCK: app_user
```

I don't think it's much of a problem, and it's probably not worth spending more time on this, so I will leave it.

PR #1048, opened 2026-03-17
Enhance asset precompilation reliability with async execution

### LINKED ISSUE #1047: Improve Reliability of assets:precompile During Deployment
## Summary
During production deployments, the `assets:precompile` step can get stuck or misreported as running by Ansible, even when the underlying process has already completed or terminated. This leads to inconsistent and unreliable deployment behavior across multiple regions.

---

## Problem

While performing a production deployment (first-time run), the following issues were observed:

- Consistent failures on:
  - 🇨🇦 Canada
  - 🇩🇪 Germany
  - 🇺🇸 USA instances
- Deployment gets stuck at the `assets:precompile` step
- Ansible reports the task as **running indefinitely**

### Observed Behavior

- When SSH-ing into the server:
  - The precompile process was **already completed OR terminated**
- However:
  - Ansible continued to show the task as **running**
  - No failure or error was reported

---

## Root Cause (Hypothesis)

This is likely due to how long-running tasks behave over SSH:

- If the deployer location is very far away from servers, this may cause timeout issues due to high latency
- SSH connection drops or becomes unstable
- Silent timeouts occur
- Ansible loses track of the actual process
- Task state becomes **out of sync** with reality

---

## Proposed Solution

Run the precompile step **asynchronously** using Ansible’s async mechanism instead of relying on a single long-lived SSH connection.

### Change

```yaml
- name: precompile assets # noqa 301
  command: bash -lc "bundle exec rake assets:precompile RAILS_ENV={{ rails_env }}"
  args:
    chdir: "{{ build_path }}"
  become: yes
  become_user: "{{ app_user }}"
  async: 1800
  poll: 10

chahmedejaz:
**Closes #1047** 

## Summary
Improve reliability of asset precompilation during deployment by running the task asynchronously and monitoring it with controlled polling.

## Changes
- Added `async: 1800` to allow the asset precompilation task to run in the background with a maximum timeout of 30 minutes.
- Added `poll: 10` to check task status every 10 seconds while waiting for completion.

## Reason
Asset precompilation can take significant time and may cause deployment failures if executed synchronously. Running it asynchronously prevents blocking the playbook, while polling ensures the task is still tracked and completed reliably without overwhelming the server.

## Impact
- Reduces deployment failures due to long-running asset builds
- Improves stability of production deployments
- Maintains visibility of task progress

## Testing
- Verified deployment completes successfully with async execution
- Confirmed assets are precompiled correctly

### DIFF
--- a/roles/deploy/tasks/deploy.yml
+++ b/roles/deploy/tasks/deploy.yml
@@ -123,6 +123,8 @@
     chdir: "{{ build_path }}"
   become: yes
   become_user: "{{ app_user }}"
+  async: 1800
+  poll: 10
 
 - name: migrate database
   command: bash -lc "bundle exec rake db:migrate RAILS_ENV={{ rails_env }}"


dacook (approved):

mkllnk (approved):
Nice. I didn't know about this feature.

PR #1053, opened 2026-04-08
Install node with our script in the app

mkllnk:
I removed and replaced the node installation role with the script we already have in our main repository. The main benefit is that our script upgrades the node_install plugin automatically when it's out of date. Merging this will make upgrading node easier, for example:

- https://github.com/openfoodfoundation/openfoodnetwork/pull/14138

Tested on all staging servers.


rioug (approved):
Great ! thanks for taking care of this.

dacook (approved):
Perfect 👍

PR #1056, opened 2026-04-21
Remove old Zapier IP address

### LINKED ISSUE #1031: Remove old zapier ip address
## Description

See #1029 , zapier are updating their ip range. We 'll need to remove the old ones after Feb 15 2025. 

!! Make sure integration still work with only the new ips enabled

mkllnk:
- Closes https://github.com/openfoodfoundation/ofn-install/issues/1031

Zapier updated their IP addresses. The official entries are listed in this article:

- https://help.zapier.com/hc/en-us/articles/15406083674509-Use-a-static-IP-address-to-connect-to-Zapier

The old IP address should not be in use anymore. We should test that though.

### DIFF
--- a/inventory/host_vars/coopcircuits.fr/config.yml
+++ b/inventory/host_vars/coopcircuits.fr/config.yml
@@ -28,7 +28,6 @@ custom_hba_entries:
   - "{{ custom_hba_n8n_IPv6 }}"
   - "{{ custom_hba_zapier_1 }}"
   - "{{ custom_hba_zapier_2 }}"
-  - { type: hostssl, database: "{{ db }}", user: zapier, address: '54.86.9.50/32', auth_method: md5 }
 
 attachment_path: "home/openfoodnetwork/apps/openfoodnetwork/current/public/spree/products/:id/:style/:basename.:extension"
 attachment_url: ofn-prod.s3.us-east-1.amazonaws.com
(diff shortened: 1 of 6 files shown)


dacook (approved):
I'm sure that we're not using it in Australia. It might be worth checking with Bethan also?

rioug (approved):

rioug:
> I'm sure that we're not using it in Australia. It might be worth checking with Bethan also?

It was flagged by Karla from OFN Canada, so we should check with them as well.

mkllnk:
> I'm sure that we're not using it in Australia.

I just logged into our account ~and it looks like there are a few enterprises set up to add new customers to Mailchimp via Zapier. I haven't checked if those enterprises are still trading though~ and half of the zaps seem to be broken.

_Update:_ I think that we are checking on new enterprise creation. It looks like we have two active zaps accessing Postgres. We could probably replace them with n8n but I'm not sure that there's a need for that at the moment.

dacook:
As this is still being used, I'm guessing we will need to update the IP address before they close down the old ones?

rioug:
It's done already : https://github.com/openfoodfoundation/ofn-install/issues/1029 , I think we just need to make sure that automation are still working with just the new ip/

dacook:
Oh ok. Your PR says they are due by Feb 2026, so I would consider the test complete ;)
In which case, we can merge and provision, no?

mkllnk:
> we can merge and provision, no?

Yes, I think so. Production test should be enough.

mkllnk:
Provisioning:
```
ansible-playbook --limit au_staging playbooks/provision.yml -t dbserver
```

- [x] au-staging
- [x] fr
- [x] ca
- [x] us
- [x] au
- [x] uk

PR #1060, opened 2026-05-10
Install FerrumPDF browser dependencies

### LINKED ISSUE #1059: Install FerrumPDF browser dependencies
We are migrating from `wicked_pdf/wkhtmltopdf` to FerrumPDF/Ferrum.

Unlike wkhtmltopdf, Ferrum requires a Chrome browser binary to be installed and accessible on the server. CI currently works because GitHub Actions runners already include Chrome, but production/staging environments don't have one right now.

This is required by: https://github.com/openfoodfoundation/openfoodnetwork/issues/13547

chahmedejaz:
## What changed and why

- Closes #1059 

Ferrum/FerrumPDF requires a Chrome browser binary for headless PDF rendering, unlike wkhtmltopdf which bundled its own rendering engine.

This PR adds a dedicated Ansible `chrome` role to install Google Chrome and required runtime libraries on provisioned environments. This ensures Ferrum-based PDF rendering works consistently outside CI, including staging and production servers.

## Changes

- Added new `chrome` Ansible role
- Installed `google-chrome-stable` from the official Google repository
- Added required headless Chrome runtime dependencies:
  - `libgbm1`
  - `libxshmfence1`
- Wired the role into `playbooks/provision.yml`

## Required By

- https://github.com/openfoodfoundation/openfoodnetwork/pull/14221


rioug (approved):
Looks good 👍

roles/chrome/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+# Runtime libraries required by headless Chrome that may not be pulled
+# in automatically on minimal Ubuntu server installations.
+- name: install headless chrome runtime dependencies
+  apt:
+    name:
+      - libgbm1          # GPU buffer management, required even in headless mode
+      - libxshmfence1    # Shared memory fences for X11 shm rendering
+    state: present
+  become: yes
```
dacook: I would have thought that the chrome package would have published these dependencies so that apt could automatically install them.
Oh well, thanks for making this work 👍

dacook (approved):
Good work, thanks for sorting this out. 
I wondered if Chromium would be a better option because it is open-source, but I don't know that it would make much difference.

I also wonder if this will consume more resources for PDF generation than before, but have no idea. We will see!

I would just like to also get @mkllnk's review so he is aware of it.

roles/chrome/tasks/main.yml
mkllnk: I guess that these dependencies are only needed when you try to run it without a desktop.

roles/chrome/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    repo: "deb http://dl.google.com/linux/chrome/deb/ stable main"
+    filename: google-chrome
+  become: yes
+
+- name: install google chrome stable
+  apt:
+    name: google-chrome-stable
+    state: present
+    update_cache: yes
+  become: yes
```
mkllnk: Can't we use Chromium? Then we don't need an additional repository.

roles/chrome/tasks/main.yml
chahmedejaz: Exactly 👍

chahmedejaz:
@mkllnk @dacook - I think we should be able to use Chromium as well. I added Chrome here because it was already present on my machine, so I tested with it, and FerrumPDF worked as expected with it. Moreover, I thought it's well maintained and would be more stable for production. 😅 

Please let me know if you'd like me to try Chromium instead and update the PR accordingly. Thanks.

mkllnk:
I've got only Chromium installed on my machine and I can generate PDFs with FerrumPdf. Chromium and Chrome share the same code and therefore I would expect similar quality. I think that we should go for completely open source when it's easy and in this case it seems even easier than installing Chrome.

chahmedejaz:
Sure, makes sense @mkllnk. I'll update the PR and get back. Thanks 👍🏻

mkllnk (approved):

mkllnk:
@chahmedejaz, I wanted to apply this to all production servers now because the FerrumPDF PR is tested and ready for merging. First I tried to run this against the staging servers but it failed on some, e.g. uk_staging. It looks like there's another PPS for ffmpeg4 on au_staging that is needed but not installed by this role. And on fr_staging that PPA wasn't needed as libpipewire-0.3 is supplied by the Ubuntu sources.

It looks like uk_staging has Ubuntu's extended security support enabled and those servers don't serve the newer version of libpipewire. I thought you installed it on all staging servers? Did I get that wrong?

Now looking at au_prod, it doesn't have ESM but it also doesn't have the new libpipewire available.

By the way, I didn't need secrets to run only this role because it doesn't use any secrets:

```
ansible-playbook -l uk_staging -t chromium playbooks/provision.yml
```

In order to resolve this quickly, I added the ppa that was on au-staging and I'm applying it to production servers. Several failed at first:

- [x] fr
- [x] be
- [x] in
- [x] nz

Now working through them. ... The common problem seems to be outdated apt sources. I deactivated the ones that didn't work. We need to replace our servers soon anyway.

chahmedejaz:
Thanks @mkllnk for picking this up and going through this.

I think we had a discussion on it in recent delivery circle that it's not installed on UK. Ref meeting notes: https://community.openfoodnetwork.org/t/delivery-circle-20th-may-2026/3958

I'm about to work on it hopefully today. I'll add the missing ppa in the role as well for consistency.

Anyway, thank you so much for going through this. ♥️

Edit: Just saw, you have already addressed this in the role. Thanks 👍🏻

PR #1063, opened 2026-05-21
Add ffmpeg4 ppa for chromium

mkllnk:
Most Ubuntu 20 servers didn't have access to the required version of libpipewire-0.3. This PPA adds that package and we can install Chromium.

This has already been applied to all managed servers.


chahmedejaz (approved):

chahmedejaz (approved):
Thanks for addressing this @mkllnk

rioug (approved):

